PR #14, opened 2025-01-17
feat(API): expose invocation wait and fallback on it on blocking requests

seriousben:
## Context

Running ` graph.run(block_until_done=True, text=text)`, if there were any blips with the server or the client, the function would throw without a way to resume waiting.

## What

This PR leverages a new wait invocation endpoint (https://github.com/tensorlakeai/indexify/pull/1179) to add a new http_client method that allows waiting on a invocation using its invocation id.

We also leverage this new endpoint to do exponential backoff retries when `graph.run(block_until_done=True` is disconnected.

## Testing

Start a long running graph, stop the server and restart it multiple times.

image

## Contribution Checklist

- [x] please run `make fmt`
- [x] Make sure all PR Checks are passing.

### DIFF
--- a/src/tensorlake/http_client.py
+++ b/src/tensorlake/http_client.py
@@ -1,20 +1,28 @@
 import json
 import os
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, Optional, Union
 
 import cloudpickle
 import httpx
-from httpx_sse import connect_sse
+from httpx_sse import ServerSentEvent, connect_sse
 from pydantic import BaseModel, Json
 from rich import print  # TODO: Migrate to use click.echo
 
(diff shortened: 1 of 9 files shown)


src/tensorlake/http_client.py
```diff
@@ -161,7 +174,7 @@ def __enter__(self):
         return self
 
     def __exit__(self, exc_type, exc_value, traceback):
-        self.close()
+        self._close()
```
eabatalov, line 177: I'm wondering if this was the source of leaking socket warning messages when we are running SDK tests

eabatalov (approved):

src/tensorlake/http_client.py
seriousben: I still see tons of them unfortunately.

PR #17, opened 2025-01-22
Test fixes

eabatalov:
run_tests.sh:
Add --exclude arg to exclude some tests temporarily.

test_broken_graphs.py:
removed unused import, fix flaky errors

test_file_descriptor_caching.py:
make it work with read only file system and not depend on host OS filesystem state.

test_function_process_crash.py:
Reduce number of iterations to speed up the test.

test_function_concurrency.py:
Use processes instead of threads to run invocations really concurrently.
This should reduce test flakiness and fix its current failures.

### DIFF
--- a/tests/compose.yaml
+++ b/tests/compose.yaml
@@ -28,6 +28,6 @@ services:
       server:
     deploy:
       mode: replicated
-      replicas: 2
+      replicas: 1 # Some tests require a single executor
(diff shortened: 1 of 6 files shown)


eabatalov:
The TensorlakeCompute constructor test will stop failing once we publish new Executor container image with updated SDK.

PR #19, opened 2025-01-23
Reformat code to pass lint checks and add TensorlakeCompute caching test

eabatalov:
Also bumped the version to remove debug prints in the new release.


eabatalov:
Only expected test failures are in place due to use of the currently published Indexify container image which doesn't have the fixes for the tests.

PR #50, opened 2025-02-24
Fix bug with wrong invocation state check and add tests

eabatalov:
Looks like in PR https://github.com/tensorlakeai/indexify/pull/1235 invocation status values were changed and SDK started to return empty outputs for all async invocations. This resulted in failures of graph update tests that use async invocations.

I also added tests for sync and async invocations outputs and a test for running invocation graph version update.

I'm not sure why tests that run async invocations didn't fail in PR https://github.com/tensorlakeai/indexify/pull/1235 but now they are failing because of this. It looks like some transient error with tensorlake submodule resulted in tensorlake SDK tests not running for it.

I ran SDK tests also in Indexify mainline and this PR fixes the failures there.

### DIFF
--- a/examples/datasets.py
+++ b/examples/datasets.py
@@ -127,4 +127,5 @@ async def main():
         for key_info, data in items.items():
             writer.writerow([key_info.job_id, key_info.file_name, data])
 
+
 asyncio.run(main())
--- a/src/tensorlake/http_client.py
+++ b/src/tensorlake/http_client.py
@@ -416,8 +416,10 @@ class TensorlakeClient:
         )
         response.raise_for_status()
         graph_outputs = GraphOutputs(**response.json())
-        if graph_outputs.status == "pending":
+        # "pending" is old server API behavior, left here for backward compatibility.
+        if graph_outputs.status in ["pending", "Pending", "Running"]:
             raise GraphStillProcessing()
+
         outputs = []
         for output in graph_outputs.outputs:
             if output.compute_fn == fn_name:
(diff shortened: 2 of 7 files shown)


examples/datasets.py
```diff
@@ -127,4 +127,5 @@ async def main():
         for key_info, data in items.items():
             writer.writerow([key_info.job_id, key_info.file_name, data])
 
+
```
eabatalov, line 130: This is autoformatting. Was missing looks like.

src/tensorlake/http_client.py
```diff
@@ -416,8 +416,10 @@ def graph_outputs(
         )
         response.raise_for_status()
         graph_outputs = GraphOutputs(**response.json())
-        if graph_outputs.status == "pending":
```
seriousben, line 419: Oh thanks. Sorry about that.

tests/tensorlake/testing.py
```diff
@@ -25,3 +27,11 @@ def test_graph_name(test_case: unittest.TestCase) -> str:
     ...         # test_graph_reduce_test_simple
     """
     return unittest.TestCase.id(test_case).replace(".", "_")
+
+
+def wait_function_output(graph: RemoteGraph, invocation_id: str, func_name: str) -> Any:
```
seriousben, line 32: There is also a new wait API endpoint exposed by the server.

eabatalov: This is nice

PR #52, opened 2025-02-27
Make a more strict GPU health check using pytorch

eabatalov:
It turns out that nvidia-smi doesn't detect the known issue https://github.com/NVIDIA/nvidia-container-toolkit/issues/857

So in addition to that I'm adding a simple pytorch CUDA computation as a health check. This is known to detect the known issue.

Also refactored the code to extra the health check handler into a separate class because it quite complex now.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "tensorlake"
-version = "0.1.27"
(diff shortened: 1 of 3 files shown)


eabatalov:
The health check actually works so we don't need this health check.

PR #55, opened 2025-02-28
Add additional_modules support when creating Graph.

j3m7:
This adds support for specifying additional modules to a graph while also allowing tensorlake CLI to handle deployment steps.

Specifically it adds an `additional_modules` parameter to the Graph class. These the additional_modules in the Graph object will be added to whatever `additional_modules` where passed during RemoteGraph.deploy(). This preserves the existing behavior.


tests/tensorlake/test_graph_update.py
```diff
@@ -93,7 +98,12 @@ def middle_node2(data: dict) -> dict:
             def end_node2(data: dict) -> int:
                 return data["num"] + 3
 
-            g = Graph(name=graph_name, start_node=start_node2, version="2.0")
+            g = Graph(
+                name=graph_name,
+                start_node=start_node2,
+                version="2.0",
+                additional_modules=[testing, parameterized],
```
diptanu, line 106: What is being tested here? graph behaviors are probably the better place to do this instead of graph update if you just want to do a sanity check

j3m7: No idea, this PR just ports the test to use the new additional_modules functionality.

PR #60, opened 2025-03-07
Avoid streaming logs if we don't have to.

j3m7:
Avoid streaming build logs during a build if they aren't going to be shown to the user.


stangirala:
+1

PR #62, opened 2025-03-10
feat: adding secrets management in CLI

seriousben:
New commands:

- `tensorlake secrets list`
- `tensorlake secrets set NAME1=val NAME2=val2`
- `tensorlake secrets unset NAME1 NAME2`


src/tensorlake/cli/prepare.py
```diff
@@ -0,0 +1,39 @@
+import asyncio
+from typing import Dict
+
+import click
+
+from src.tensorlake.cli.deploy import _import_workflow_file, _prepare_images
```
seriousben:
```suggestion
from tensorlake.cli.deploy import _import_workflow_file, _prepare_images
```

PR #65, opened 2025-03-11
Don't use `which -s` flag for GPU health check

eabatalov:
`-s` flag suppresses `which` output but it's not available in all Linux distributions, see example error:

```
Usage: /usr/bin/which [-a] args
Illegal option -s
```

Lack of -s flag results in GPU health check not being enabled when GPU is available. So this is a bug.

Instead of using `-s` flag run which with /dev/null (suppressed) stdout, stderr.


eabatalov:
The test workflow is stuck for some reason but all tests pass locally.

PR #74, opened 2025-03-18
fix: secrets list pagination

seriousben:
Fixing secrets pagination only returning 10 secrets.

### DIFF
--- a/src/tensorlake/cli/secrets.py
+++ b/src/tensorlake/cli/secrets.py
@@ -1,4 +1,3 @@
-import json
 from datetime import datetime, timezone
 from typing import Dict, List, Tuple
 
@@ -33,7 +32,7 @@ def list(auth: AuthContext):
 
     table = Table()
 
-    table.add_column("Name", justify="right", no_wrap=True)
+    table.add_column("Name", no_wrap=True)
     table.add_column("Created At", style="green")
 
     for secret in secrets:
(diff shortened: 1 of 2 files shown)


src/tensorlake/cli/secrets.py
```diff
@@ -134,7 +133,7 @@ def unset(auth: AuthContext, secret_names: str):
 
 def _get_all_existing_secrets(auth: AuthContext) -> List[dict]:
     resp = auth.client.get(
-        f"/platform/v1/organizations/{auth.organization_id}/projects/{auth.project_id}/secrets?limit=100"
+        f"/platform/v1/organizations/{auth.organization_id}/projects/{auth.project_id}/secrets?pageSize=100"
```
stangirala, line 136: do we expect a user to have fewer than 100 secrets?

src/tensorlake/cli/secrets.py
```diff
@@ -98,7 +97,7 @@ def set(auth: AuthContext, secrets: str):
     resp.raise_for_status()
 
     if len(upsert_secrets) == 1:
```
stangirala, line 99: is there a reason we check for len == 1?

seriousben: Just pluralization :)

stangirala: you could just do `set <1> secret(s)`

src/tensorlake/cli/secrets.py
seriousben: more than 100 is very unlikely, if it gets to a high number we'll need to rethink the CLI UX.

stangirala:
+1

PR #78, opened 2025-03-22
Supporting metrics

diptanu:
Added an API to support metrics in functions. This requires a user inject the context to access metrics APIs

```
@tensorlake_function(inject_ctx=True)
def my_fun(ctx: GraphInvocationCtx, foo: str) -> str:
   ctx.timer("bar", 1.0)
   ctx.counter("foo", 5)
```


src/tensorlake/functions_sdk/invocation_state/invocation_state.py
```diff
@@ -9,14 +9,24 @@ class InvocationState:
     Each new invocation starts with an empty state (empty set of key-value pairs).
     A value can be any CloudPickleSerializer serializable object."""
 
+    def __init__(self):
+        self.timers: Dict[str, float] = {}
+        self.counters: Dict[str, int] = {}
```
eabatalov, line 14: Nit. These metrics are part of a single task run state (not whole graph invocations state).

eabatalov:
LGTM, clean PR.

PR #79, opened 2025-03-23
Simplify routing 

diptanu:
We are not currently using Routers because they need to be inserted into the graph, which essentially adds latency to the whole graph run.

We change the behavior to let any function decide the next set of edges dynamically. 

```python

@tensorlake_function(inject_ctx=True):
def foo(ctx, val: int) -> int:
   if val % 2 == 0:
      ctx.next_edges("bar")
      
   return val**2
   
 @tensorlake_function():
 def bar(val: int) -> str:
    return str(val)
    
```

in this example the graph would go from foo -> bar if the val is even otherwise the graph would just terminate. 

One the server side next_edges would decide where data is routed from the current node. If there are edges defined in the graph, we will override them with the next_edges if they are returned by the function. Why? Because it's simpler and we have a use-case for this in production, and no open source user has asked for anything related to routers yet. In Document AI if a user suppresses layout detection we can go straight to OCR stage, and then we can also avoid doing table detection, if there is no layout for a page.


src/tensorlake/functions_sdk/functions.py
```diff
@@ -212,21 +155,17 @@ def construct(fn):
...
     ser_outputs: List[TensorlakeData]
     traceback_msg: Optional[str] = None
-
-
-class RouterCallResult(BaseModel):
-    edges: List[str]
-    traceback_msg: Optional[str] = None
+    metrics: Optional[Metrics] = None
+    output_encoding: str
+    next_nodes: Optional[List[str]] = None
```
eabatalov, line 160: Should the next node only be a single node instead of a list?
If the next_node is None and there are more the two edges coming out of the current node then this is a runtime error right? Or we take the first edge by default?

PR #85, opened 2025-03-24
Fix Job object to match Doc AI API

stangirala:


src/tensorlake/documentai/jobs.py
```diff
@@ -155,7 +155,10 @@ class Job(BaseModel):
     DocumentAI job class.
     """
 
-    job_id: str = Field(alias="jobId")
-    file_id: str = Field(alias="fileId")
+    job_id: str = Field(alias="id")
```
j3m7, line 158: For my personal and professional enrichment, where is the server side struct for this?

stangirala: it's in inkwell-server https://github.com/tensorlakeai/inkwell-server/tree/main/server/src/api/v1/handlers.

It might help to use openapi to autogenerate some of these clients.

PR #87, opened 2025-03-25
Handle async connection re-use: close connection after call

stangirala:
See https://github.com/encode/httpx/discussions/2959

For the SDK close the connection once we are done with it.


stangirala:
Bypass +1, tests are still broken.

PR #104, opened 2025-04-10
feat: integrate with Image Builder V2

miguelhrocha:
## Description

Now we have a new service to build images in Tensorlake Cloud. This PR adds a boolean flag to build images using that service.

The new image builder integration is gated under the `--builder-v2` flag in the deploy command. It is `false` by default.

## Screenshots

Screenshot 2025-04-16 at 17 09 28

### DIFF
--- a/src/tensorlake/builder/client_v2.py
+++ b/src/tensorlake/builder/client_v2.py
@@ -36,7 +36,7 @@ class BuildContext:
     and function name used for building the image.
 
     Attributes:
-        graph (str): The name of the graph to be built.
+        graph_name (str): The name of the graph to be built.
         graph_version (str): The version of the graph to be built.
         function_name (str): The name of the function used in the build.
 
@@ -48,7 +48,7 @@ class BuildContext:
         )
     """
 
-    graph: str
+    graph_name: str
     graph_version: str
     function_name: str
 
@@ -161,14 +161,14 @@ class ImageBuilderV2Client:
             dict: The response from the image builder service.
         """
         click.echo(
-            f"Building {context.graph} version {context.graph_version} for {context.function_name}"
+            f"Building {context.graph_name} version {context.graph_version} for {context.function_name}"
         )
 
         _fd, context_file = tempfile.mkstemp()
         image.build_context(context_file)
 
         click.echo(
-            f"{context.graph}: Posting {os.path.getsize(context_file)} bytes of context to build service...."
+            f"{context.graph_name}: Posting {os.path.getsize(context_file)} bytes of context to build service...."
         )
 
         files = {}
@@ -177,7 +177,7 @@ class ImageBuilderV2Client:
 
         os.remove(context_file)
         data = {
-            "graph_name": context.graph,
+            "graph_name": context.graph_name,
             "graph_version": context.graph_version,
             "graph_function_name": context.function_name,
             "image_hash": image.hash(),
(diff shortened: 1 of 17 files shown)


diptanu (approved):
Approving since I don't see any issue to unblock Miguel. @eabatalov and @j3m7 please review this before we merge this one.

src/tensorlake/builder/client_v2.py
```diff
@@ -0,0 +1,251 @@
...
+            context_collection (Dict[Image, BuildContext]): A dictionary mapping images to their build contexts.
+        Returns:
+            dict: A dictionary mapping image hashes to their corresponding build IDs.
+        """
+        click.echo("Building images...")
+
+        builds = {}
+        for image, context in context_collection.items():
+            click.echo(f"Building {image.image_name}")
+            build = await self.build(context, image)
```
eabatalov, line 146: we are intentionally not doing this in parallel right now, right?

src/tensorlake/builder/client_v2.py
```diff
@@ -0,0 +1,251 @@
...
+                                    graph version, and function name.
+            image (Image): The image to be built.
+        Returns:
+            dict: The response from the image builder service.
+        """
+        click.echo(
+            f"Building {context.graph_name} version {context.graph_version} for {context.function_name}"
+        )
+
+        _fd, context_file = tempfile.mkstemp()
```
eabatalov, line 167: We are not using a context manager to automatically remove the files in case of errors (exceptions).

src/tensorlake/builder/client_v2.py
```diff
@@ -0,0 +1,251 @@
...
+
+        os.remove(context_file)
+        data = {
+            "graph_name": context.graph_name,
+            "graph_version": context.graph_version,
+            "graph_function_name": context.function_name,
+            "image_hash": image.hash(),
+        }
+
+        res = await self._client.put(
```
eabatalov, line 186: Would adding retries help here for better ux in case e.g. if we deploy image builder service and it's temporarily down?

eabatalov (approved):
lgtm, see the small comments

src/tensorlake/builder/client_v2.py
miguelhrocha: Yes, it is intentional. We do have a limit of concurrent builds with our buildkit vendor, and we don't want a single deployment to take all that concurrency.

j3m7: Is there a ticket somewhere to build some sort of fairness and add concurrency. Not having concurrency will raise our inkwell image builds from 10 min to 50+ minutes?

j3m7 (approved):
I think we will need concurrency in order to use this for inkwell images but this looks good enough to run some tests and see if that's actually the case. Maybe namespace can build those images fast enough.

miguelhrocha:
@j3m7 @eabatalov, your concerns are understandable. I will merge it now because this is not enabled by default and will unblock the integration efforts. I expect to iterate over this.

src/tensorlake/builder/client_v2.py
diptanu: @miguelhrocha We need some concurrency man. Like hardcode this to 5 or something at least. We can make this into a queue later. When this goes to production now, we need to at least get the document ai builds done fast.

PR #108, opened 2025-04-17
Log Indexify requests

earhart:


earhart:
FWIW: I considered reformatting the dictionary as JSON (so that the entire logged string would be JSON, simplifying prettifying a little bit), but it didn't really seem necessary.

PR #110, opened 2025-04-17
Add cache keys for functions and start node parameters

earhart:
This change adds keys for cache access checks to graph functions and the start node's parameters.
For graph functions, these are attached to the graph's JSON definition; for the start node's parameters, the key is computed by hashing the serialized input.

This change does _not_ take steps to ensure a deterministic serialization for the start node's parameters; if/when that becomes useful, we can add it separately.

Fixes tensorlakeai/indexify#1339


eabatalov (commented):
Can you please explain how these new cache keys are going to be used by SDK users and by the server?

earhart:
Switching to a draft for now -- I've already had to change this a little locally.  :-)

diptanu:
@earhart As discussed offline, I think adding a graph behavior test would clarify what @eabatalov is asking. The tests would also explain the developer experience of the feature. Other than that, the changes look good to me.

src/tensorlake/functions_sdk/functions.py
```diff
@@ -243,6 +244,14 @@ def construct(fn):
             "memory": memory,
             "ephemeral_disk": ephemeral_disk,
             "gpu": gpu,
+            "cache_key": (
```
eabatalov: This approach slightly changes they way we were handling function attributes, the current approach for SDK function attributes is to either use basic data types like int, str, None or to define a struct like here https://github.com/tensorlakeai/tensorlake/blob/main/src/tensorlake/functions_sdk/functions.py#L84.
If we need to derive a value/struct from a function attribute then we do it at the place where this derived value is needed.

If you don't mind let's keep using it for consistency.

src/tensorlake/functions_sdk/functions.py
```diff
@@ -222,6 +222,7 @@ def tensorlake_function(
     memory: float = _DEFAULT_MEMORY,
     ephemeral_disk: float = _DEFAULT_EPHEMERAL_DISK,
     gpu: Optional[Union[str, List[str]]] = _DEFAULT_GPU,
+    cache_key: Optional[str] = None,
```
eabatalov: This new field needs to be added to TensorlakeCompute struct and ideally described in its constructor doc string.

The @tensorlake_function decorator calls TensorlakeCompute constructor.

src/tensorlake/functions_sdk/graph.py
```diff
@@ -234,6 +235,14 @@ def definition(self) -> ComputeGraphMetadata:
                     delay_multiplier=start_node.retries.delay_multiplier,
                 )
             ),
+            cache_key=(
+                CacheKeyMetadata(
+                    algorithm=start_node.cache_key["algorithm"],
+                    value=start_node.cache_key["value"],
+                )
+                if hasattr(start_node, "cache_key") and start_node.cache_key is not None
```
eabatalov: iirc this `hasattr(start_node, "cache_key")` is not needed, is there a bug somewhere?

src/tensorlake/functions_sdk/graph.py
```diff
@@ -234,6 +235,14 @@ def definition(self) -> ComputeGraphMetadata:
                     delay_multiplier=start_node.retries.delay_multiplier,
                 )
             ),
+            cache_key=(
+                CacheKeyMetadata(
+                    algorithm=start_node.cache_key["algorithm"],
```
eabatalov: Let's not use dicts/maps and use dataclasses or pydantic models. Python is dynamic enough already, let's not add more without a need.

src/tensorlake/http_client.py
```diff
@@ -1,3 +1,5 @@
+from base64 import b64encode
```
eabatalov: Looks like redundant changes

eabatalov (changes requested):
Please add a test to https://github.com/tensorlakeai/tensorlake/blob/main/src/tensorlake/functions_sdk/graph_definition.py - this will validate graph definition we're sending to Server. This functionality belongs to SDK.

Then when it's ready please add actual integ tests to Indexify package because this caching functionality won't be used in local graph execution mode available in Tensorlake afaik.

src/tensorlake/functions_sdk/functions.py
earhart: Done -- I simplified it to a single string, since that's really all the server cares about.

src/tensorlake/functions_sdk/functions.py
earhart: Done.

src/tensorlake/functions_sdk/graph.py
earhart: It was needed for nodes that weren't functions; that was fixed by adding the field to `TensorlakeCompute`.

src/tensorlake/functions_sdk/graph.py
earhart: Done

src/tensorlake/http_client.py
earhart: Old code :-). Done

earhart:
I created a new test file, since this one wound up being a little different, but: there's a test there now.
(With invocation event streaming, too :-)
PTAL?

diptanu (approved):

earhart:
Looks like this is just pending on the server update at this point.

PR #111, opened 2025-04-22
Remove SDK side resource validation

eabatalov:
The available resources depend on region, Indexify cluster configuration and etc. So SDK (client) can't make any assumptions about possible resorce constraints.


eabatalov:
I'm doing cleanups in Indexify to make these tests passing so this can be merged.

PR #112, opened 2025-04-22
fix: ADD operations in Image include directories/files mentioned

miguelhrocha:
The Image add operations are not adding the directory mentioned as the source in the tarball created as the Docker context.

### DIFF
--- a/src/tensorlake/functions_sdk/image.py
+++ b/src/tensorlake/functions_sdk/image.py
@@ -148,7 +148,7 @@ class Image:
 
     def _is_git_repo_url(self, path: str) -> bool:
         parsed = urlparse(path)
-        return parsed.scheme == "git" or "github.com" in parsed.netloc
+        return parsed.scheme == "git" or (parsed.hostname and (parsed.hostname == "github.com" or parsed.hostname.endswith(".github.com")))
 
     def _is_inside_git_dir(self, path: str) -> bool:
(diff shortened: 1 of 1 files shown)


src/tensorlake/functions_sdk/image.py
```diff
@@ -150,6 +162,23 @@ def build_context(self, filename: str):
...
+                elif op.op_type == "ADD":
+                    if self._is_url(src) or self._is_git_repo_url(src):
+                        logging.warning(
+                            "Skipping ADD: %s is a URL or Git repo reference", src
+                        )
+                        continue
+                    if not os.path.exists(src):
+                        logging.warning("Skipping ADD: %s does not exist", src)
+                        continue
+                    if self._is_inside_git_dir(src):
```
j3m7, line 174: Why are we hardcoding this instead of honoring .dockerignore?

miguelhrocha: The .dockerignore is another set of directives. This is to avoid adding files from a .git directory.

I can add honoring .dockerignore in another PR, but this is blocking my platform executor work  because existing tests require the usage of the ADD operation, and now that we are using the context for tests, these won't pass

j3m7: Ok, approved

adithyaakrishna (approved):

PR #119, opened 2025-04-28
chore: append workflows to default service url

ilbum:
## Description

- append `/workflows` to `DEFAULT_SERVICE_URL`

### DIFF
--- a/src/tensorlake/settings.py
+++ b/src/tensorlake/settings.py
@@ -1,3 +1,3 @@
 import os
 
-DEFAULT_SERVICE_URL = os.getenv("INDEXIFY_URL", "https://api.tensorlake.ai")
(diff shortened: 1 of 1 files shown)


src/tensorlake/settings.py
```diff
@@ -1,3 +1,3 @@
 import os
 
-DEFAULT_SERVICE_URL = os.getenv("INDEXIFY_URL", "https://api.tensorlake.ai")
+DEFAULT_SERVICE_URL = os.getenv("INDEXIFY_URL", "https://api.tensorlake.ai/workflows")
```
j3m7, line 3: This variable is already being set by a bunch of existing CI, what is the rationale for changing it?

PR #121, opened 2025-04-30
update defaults

diptanu:
Reduce default resource requirements


diptanu:
@eabatalov Merging this. For some reason it's stuck in CI. I can run this locally

PR #138, opened 2025-05-16
cache function wrappers

diptanu:
Cache functions so that they are not reloaded every time if run in a loop during local testing


cooleel (approved):
LGTM

PR #150, opened 2025-06-01
Langchain tool init

diptanu:
Started a tools package for Langchain

### DIFF
--- a/src/tensorlake/langchain/documents.py
+++ b/src/tensorlake/langchain/documents.py
@@ -1,26 +1,213 @@
 from langchain_core.tools import StructuredTool
 from pydantic import BaseModel, Field
+from tensorlake.documentai import DocumentAI, ParsingOptions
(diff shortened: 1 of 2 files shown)


src/tensorlake/langchain/documents.py
```diff
@@ -0,0 +1,216 @@
...
+from tensorlake.documentai import DocumentAI, ParsingOptions
+from tensorlake.documentai.parse import (
+    ChunkingStrategy,
+    TableParsingStrategy,
+    TableOutputMode,
+    ExtractionOptions,
+    ModelProvider,
+    FormDetectionMode
+)
+
```
drguthals:
```suggestion
load_dotenv()
```

src/tensorlake/langchain/documents.py
```diff
@@ -0,0 +1,216 @@
+import time
+import os
```
drguthals, line 2:
```suggestion
import os
from dotenv import load_dotenv
```

PR #152, opened 2025-06-02
Refactoring router implementation

earhart:
Fixes tensorlakeai/indexify#1344


src/tensorlake/functions_sdk/functions.py
```diff
@@ -339,10 +372,13 @@ def run_fn(
         ctx: GraphInvocationContext,
         input: Union[Dict, Type[BaseModel], List, Tuple],
         acc: Optional[Type[Any]] = None,
-    ) -> Tuple[List[Any], Optional[str]]:
+    ) -> Tuple[Union[RouteTo, Any], Optional[str]]:
```
diptanu: Not a direct comment here, but let's remove the run_router functions

diptanu (changes requested):
Left some minor comments, but this is headed in the right direction. Thanks!

diptanu (approved):
LGTM. Feels like it's at parity with what we had before

src/tensorlake/function_executor/proto/function_executor.proto
```diff
@@ -105,7 +102,7 @@ message Metrics {
 message RunTaskResponse {
     optional string task_id = 1;
     optional FunctionOutput function_output = 2;
-    optional RouterOutput router_output = 3;
+    reserved 3;  // Was: router_output
```
eabatalov: FE needs to keep returning responses in both modes so Executor mainline and Executor from server-sse-stream-stable branch can work with the FE responses.
Basically we need backward compatibility in this FE change or we need to break it and have a clear plan on how we manage it.

e.g. we can pin tensorlake pacakge version used by Document AI for now. However, this will block any new features in Workflows being accessible to Document AI. This might not happen though.

eabatalov: Also please keep in mind that we are already breaking backward compatibility in tests here so updting tensorlake git submodule will be restricted too.

tests/tensorlake/test_graph_api.py
```diff
@@ -0,0 +1,94 @@
...
+def fan_out(value: int) -> int:
+    return value
+
+
+@tensorlake_function(next=[a, b, c])
+def route_out(value: int) -> RouteTo[int, Union[a, b, c]]:
+    if value % 2 == 0:
+        return RouteTo(value + 1, [a])
+
+    return RouteTo(value + 2, [b, c])
```
eabatalov: What about the syntax we discussed previously:
`RouteTo(b=value + 2, c=value +2)`
Is it excessive for now?

tests/tensorlake/test_graph_api.py
```diff
@@ -0,0 +1,94 @@
...
+
+@tensorlake_function(next=square_values)
+def parallel_map(count: int) -> List[int]:
+    return list(range(count))
+
+
+_GRAPH_PARAM_SETS = [(False), (True)]
+
+
+class TestGraphAPI(unittest.TestCase):
```
eabatalov: I'd suggest to call it routing test and rename the file.
Graph api is very generic so it'd grow a lot while ppl add tests randomly here

src/tensorlake/functions_sdk/functions.py
```diff
@@ -272,22 +211,37 @@ class FunctionCallResult(BaseModel):
...
+    """Describes a routing of data values to downstream functions.
+
+    RouteTo is returned by compute functions that require non-trivial
+    output value routing.  It's constructed with a value (as
+    ordinarily returned by a compute function), together with a list
+    of the downsteam functions that should receive that value.
+
+    NB: Each downstream function supplied to a RouteTo must be listed
+    in the compute function's @tensorlake_function decorator's "next"
+    argument.
```
eabatalov, line 232: I think this docstring doesn't help a user who just looks at it to figure out how to use it. Is there any way to improve it? Also maybe making it more concise?

tests/tensorlake/test_graph_api.py
earhart: Yeah -- I liked the idea, but it added more complication to the system; it's a bigger change than we really need right now.

tests/tensorlake/test_graph_api.py
earhart: Done

src/tensorlake/functions_sdk/functions.py
earhart: I added an example; it's not more concise, though.  :-/

diptanu (approved):

src/tensorlake/functions_sdk/functions.py
eabatalov: Thank you. Given the size of the docstring I'd prefer to only add a concise and not detailed text here and let people understand how this is used from actual code examples in separate files that we're providing anyway. No need to duplicate code examples and detailed explanations from example code files here.

src/tensorlake/functions_sdk/functions.py
```diff
@@ -296,6 +270,8 @@ def get_output_model(self) -> Any:
         extract_method = self.indexify_function.run
         type_hints = get_type_hints(extract_method)
         return_type = type_hints.get("return", Any)
+        if get_origin(return_type) is RouteTo:
```
eabatalov, line 273: Interesting. This is not related to this PR, but do I understand correctly that due to the use of `get_origin` we're strongly relying on type hints provided by users for SDK to work correctly?

If yes then this looks not quite correct to me because Python is a dynamic language and it looks like we can use actual runtime types here without any cons. Does this sound right or not?

src/tensorlake/functions_sdk/functions.py
```diff
@@ -318,38 +294,12 @@ def get_input_types(self) -> Dict[str, Any]:
...
-            edges.append(fn.name)
-        return edges, None
-
     def run_fn(
         self,
         ctx: GraphInvocationContext,
         input: Union[Dict, Type[BaseModel], List, Tuple],
         acc: Optional[Type[Any]] = None,
-    ) -> Tuple[List[Any], Optional[str]]:
+    ) -> Tuple[List[Any], Optional[str], List[str]]:
```
eabatalov, line 302: Nit. I'd add a short docstring documenting the fields in the returned tuple and what this function is doing.

tests/tensorlake/test_graph_behaviours.py
```diff
@@ -327,14 +328,15 @@ def raise_if_called_with_multiple_values(x: int, y: int, z: int) -> int:
...
+def route_multiple_values(x: int, y: int, z: int) -> RouteTo[
+    Tuple[int, int, int],
+    Union[sum_multiple_values, raise_if_called_with_multiple_values],
+]:
     if x + y + z == 0:
-        return raise_if_called_with_multiple_values
+        return RouteTo((x, y, z), raise_if_called_with_multiple_values)
     else:
-        return sum_multiple_values
+        return RouteTo((x, y, z), sum_multiple_values)
```
eabatalov, line 339: What would be the error message/UX for the customer if we return a wrong function here (not listed in next=[])

tests/tensorlake/test_graph_behaviours.py
```diff
@@ -446,6 +448,16 @@ def return_field_from_pydantic_base_model_json(input: dict) -> int:
...
 
 
+@tensorlake_function()
+def should_not_run(x: str) -> str:
+    return x + "c"
+
+
+@tensorlake_function(next=should_not_run)
+def simple_success(x: str) -> RouteTo[str, should_not_run]:
+    return RouteTo(x + "b", [])
```
eabatalov, line 458: I think passing None instead of [] would be more Pythonic and readable here especially given that we allow a single function to be passed instead of a list.

tests/tensorlake/test_graph_validation.py
```diff
@@ -110,8 +111,7 @@ def router2(a: int) -> Union[node1, node3]:
             )
 
             g.add_edge(node0, router)
```
eabatalov, line 113: Should we make graph.add_edge private now? I thought users won't be calling it anymore. At the same time we need DocAI code to get updated to not use add_edge and use the new graph definition API (next=...).
Maybe we can add a TODO to make add_edge private (like _add_edge) once Doc AI migrated?

tests/tensorlake/test_routing.py
```diff
@@ -0,0 +1,94 @@
...
+def square_values(value: int) -> int:
+    return value * value
+
+
+@tensorlake_function(next=square_values)
+def parallel_map(count: int) -> List[int]:
+    return list(range(count))
+
+
+_GRAPH_PARAM_SETS = [(False), (True)]
```
eabatalov: fyi parameterized.expand supports naming of test cases (and possibly parameter values). I mean `_GRAPH_PARAM_SETS = [(False), (True)]` is very implicit when used a global variable away from the actual test definition that has `is_remote` parameter names.

eabatalov (approved):

src/tensorlake/functions_sdk/functions.py
earhart: This is being used by graph validation; I don't think we can use actual runtime types here.

tests/tensorlake/test_graph_behaviours.py
earhart: TBH I'd rather remove the exceptional case for a single function.  :-)
"None" here means "Route to every edge specified in the `next` attr"; "Route nowhere" is indicated by an empty list.
I think we can definitely change it if we want, but I think I'd rather do that separately after some discussion and not delay this PR for it.

tests/tensorlake/test_graph_behaviours.py
earhart: Currently, I think it'll just invoke the wrong function(s).  We might want to catch that and do something.

src/tensorlake/functions_sdk/functions.py
earhart: Done

src/tensorlake/functions_sdk/functions.py
eabatalov: Hmmmm if graph validation fails then it prevents the graph from getting deployed. Might not be what we want (if users don't want to do type annotations), just fyi.

tests/tensorlake/test_graph_validation.py
earhart: That's a much bigger change to our tests -- it prevents code reuse and it prevents listing functions in workflow order.  I'd much rather get the current change in and continue our discussion around the graph API separately.

tests/tensorlake/test_routing.py
earhart: Done :-)

PR #154, opened 2025-06-04
Set multiprocessing to use spawn by default

### LINKED ISSUE #153: Avoid fork() for function executor subprocesses
Using fork() to create subprocesses is fast, but interacts in interesting ways with concurrency and IPC, and it's possible this is causing some issues with our health checks; we should use spawn semantics instead.

earhart:
This change sets the multiprocessing library's default process creation mechanism to `spawn`, creating a new Python process when required instead of simply forking the current process.

Fixes #153

### DIFF
--- a/src/tensorlake/function_executor/main.py
+++ b/src/tensorlake/function_executor/main.py
@@ -8,6 +8,7 @@ configure_logging_early()
 
 import argparse
 from typing import Any
+import multiprocessing as mp
 
 import structlog
 
@@ -27,6 +28,7 @@ def validate_args(args, logger: Any):
 
 
 def main():
+    mp.set_start_method("spawn")
     parser = argparse.ArgumentParser(
         description="Runs Function Executor with the specified API server address"
     )


src/tensorlake/function_executor/main.py
```diff
@@ -27,6 +28,7 @@ def validate_args(args, logger: Any):
 
 
 def main():
+    mp.set_start_method("spawn")
```
diptanu, line 31: We probably don't need this here for testing whether the workflows work with mp. Let's do this in the document ai workflow function where we are using processpool executor.

diptanu: Actually let's try this approach. so that no other system tries to se context

diptanu (approved):

src/tensorlake/function_executor/main.py
earhart: It turns out this is how the multiprocessing docs suggest to do it, if you're doing it globally (which I think we should -- there's no time when we should be using fork).  NB in more recent Python, spawn is the default.

PR #160, opened 2025-06-10
Start adding graphs commands to CLI

earhart:


src/tensorlake/functions_sdk/graph_definition.py
```diff
@@ -77,3 +78,20 @@ def get_input_encoder(self) -> str:
...
         raise ValueError("start node is not set on the graph")
+
+
+class TaskInfoMetadata(BaseModel):
+    pending_tasks: int
+    successful_tasks: int
+    failed_tasks: int
+
+
+class InvocationMetadata(BaseModel):
```
eabatalov, line 89: I'd prefer if we only have actual workflow graph related data model types here, not graph invocation related. The name of the file also suggests that. Let's have a separate file for the "runtime" data models if you don't mind.

eabatalov (changes requested):
The CI tests timed out. This typically indicates a hanging test (bug), ptal.

src/tensorlake/cli/graphs.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+    List remote graphs
+    """
+    if verbose and use_json:
+        raise click.UsageError("--verbose and --json are incompatible")
+
+    graphs = auth.tensorlake_client.graphs()
+
+    if use_json:
+        all_graphs = json.dumps(graphs, default=pydantic_encoder)
```
eabatalov, line 39: Nit. Seem like this could be replaced more clearly with 3 function calls:

```
if use_json:
    return print_json_graphs(graphs)
elif verbose:
    return print_graph_objects(graphs)
else:
    return print_graphs_table(graphs)

File: src/tensorlake/cli/graphs.py [hunk 3]
```diff
@@ -0,0 +1,73 @@
...
+
+@click.group()
+def graph():
+    """
+    Serverless Graph Management
+    """
+    pass
+
+
+@graph.command()
```
eabatalov, line 19: Lets add subcommand help message like here: https://github.com/tensorlakeai/indexify/blob/main/indexify/src/indexify/cli/build_image.py#L10

src/tensorlake/cli/graphs.py
```diff
@@ -0,0 +1,73 @@
...
+    table.add_column("Name")
+    table.add_column("Description")
+
+    for graph in graphs:
+        table.add_row(graph.name, graph.description)
+
+    print(table)
+
+
+@graph.command()
```
eabatalov, line 57: Let's add subcommand help message

src/tensorlake/cli/graphs.py
```diff
@@ -0,0 +1,73 @@
...
+@click.option(
+    "--json", "-j", is_flag=True, help="Export graph information as JSON-encoded data"
+)
+@click.argument("graph")
```
eabatalov, line 61: Maybe graph_name? Cause can't add a help message for arguments in click. Let's make it more clear what this means maybe?

src/tensorlake/cli/invocations.py
```diff
@@ -0,0 +1,87 @@
...
+
+@click.group()
+def invocation():
+    """
+    Serverless Graph Management
+    """
+    pass
+
+
+@invocation.command()
```
eabatalov, line 20: let's address my comments from the other file here too

src/tensorlake/cli/_common.py
```diff
@@ -13,38 +13,53 @@
...
 
     api_key: Optional[str] = None
+    base_url: str = None
     version: str = VERSION
     _client: Optional[httpx.Client] = None
     _introspect_response: Optional[httpx.Response] = None
+    _tensorlake_client: Optional[TensorlakeClient] = None
 
     def __post_init__(self):
         self.api_key = os.getenv("TENSORLAKE_API_KEY")
```
eabatalov, line 31: The commands you add in this PR might be used by OSS Indexify cluster users. Might be a good idea to support a missing API key in case direct communication with Indexify cluster is needed for the user.

src/tensorlake/functions_sdk/graph_definition.py
earhart: Done in #164

src/tensorlake/cli/graphs.py
earhart: I'm not sure about separate functions (we need to pass around arguments, have function headers, &c), but I went ahead and refactored into a more single-statement approach in #164

src/tensorlake/cli/graphs.py
earhart: No need; for commands, [help can be specified in the docstring](https://click.palletsprojects.com/en/stable/documentation/) :-)

src/tensorlake/cli/graphs.py
earhart: Ditto, it's in the docstring

src/tensorlake/cli/graphs.py
earhart: Done in #164

src/tensorlake/cli/invocations.py
earhart: Done in #164

src/tensorlake/cli/_common.py
earhart: Done in #164

PR #166, opened 2025-06-10
Upgrade libs

diptanu:
Upgrade some libraries, and made the min version of python to be 3.10


eabatalov (approved):
We'll also need to update GH actions to use 3.10 in all the repos we're using. It's a bunch of them. We can do this gradually.

PR #172, opened 2025-06-11
fix: abort deployment when graph fails to build image

miguelhrocha:
### Description

When an image fails to build with the builder V2, the deploy command should be aborted.

### DIFF
--- a/src/tensorlake/cli/deploy.py
+++ b/src/tensorlake/cli/deploy.py
@@ -171,18 +171,7 @@ async def _prepare_images_v2(
 ):
     for image_info in images.values():
         for build_context in image_info.build_contexts:
-
-            build = await builder_v2.build(build_context, image_info.image)
-            if build.status == "failed":
-                click.secho(
-                    f"Building {image_info.image.name()} failed with error message: {build.error_message}",
-                    fg="red",
-                )
-                click.secho(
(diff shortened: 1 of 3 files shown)


src/tensorlake/cli/deploy.py
```diff
@@ -172,7 +172,17 @@ async def _prepare_images_v2(
     for image_info in images.values():
         for build_context in image_info.build_contexts:
 
-            await builder_v2.build(build_context, image_info.image)
+            build = await builder_v2.build(build_context, image_info.image)
```
eabatalov: Nit. Normally the errors needs to be logged inside the builder_v2.build and exception raised. We can't put all the code like this in the main cli file. Python uses exceptions for errors. This is expectation and then we'll automatiically exit with non 0 status if build() fails.

src/tensorlake/cli/deploy.py
```diff
@@ -172,7 +172,17 @@ async def _prepare_images_v2(
...
+            if build.status == "failed":
```
eabatalov: Shouldn't we check for success here instead to make this more robust?

miguelhrocha: We also have a canceled option, which we haven't implemented, but it shouldn't raise an error. And pending too. I feel that `failed` is a better condition.

PR #174, opened 2025-06-12
add quickstart code as an example

drguthals:


diptanu:
Making this a notebook might be better.

drguthals:
We will also make it a notebook, but devs expect to be able to copy and paste code snippets from docs also. I will add an issue for that though 👍

PR #175, opened 2025-06-13
Add support for graph and invocation errors

earhart:


src/tensorlake/error.py
```diff
@@ -10,3 +10,25 @@ def __str__(self):
 class GraphStillProcessing(Exception):
     def __init__(self) -> None:
         super().__init__("graph is still processing")
+
+
+class InvocationError(Exception):
```
eabatalov: The existing errors in this file are coming from Server API. Would be confusing for future readers (both us and workflows users) I think.

eabatalov: More like api_errors.py/http_error.py and function_errors.py would work better with a comment on what's in these files.

src/tensorlake/function_executor/handlers/run_function/handler.py
```diff
@@ -92,9 +93,9 @@ def _run_func_safe_and_captured(self, inputs: FunctionInputs) -> RunTaskResponse
                 finally:
                     # Ensure that whatever outputted by the function gets captured.
                     self._flush_logs()
-        except Exception:
+        except Exception as exc:
```
eabatalov: Nit. We typically catch "Exception as e" afaik.

src/tensorlake/function_executor/handlers/run_function/response_helper.py
```diff
@@ -1,9 +1,11 @@
 from typing import List, Optional
 
-from tensorlake.functions_sdk.data_objects import TensorlakeData
+from tensorlake.error import GraphError, InvocationError
+from tensorlake.functions_sdk.data_objects import Failure, FailureScope, TensorlakeData
 from tensorlake.functions_sdk.functions import FunctionCallResult
 from tensorlake.functions_sdk.object_serializer import get_serializer
 
+from ...proto.function_executor_pb2 import FailureScope as ProtoFailureScope
```
eabatalov: Nit. Please consider if FailureScopeProto is more readable and applicable to other imports. idk but it seems to me like proto suffix makes reading easier but I'm not a native speaker.
Asking because it's better to have a consistent approach and I usually used Proto suffix

src/tensorlake/function_executor/proto/function_executor.proto
```diff
@@ -102,6 +102,23 @@ message Metrics {
     map<string, uint64> counters = 2;
 }
 
+enum FailureScope {
+    FAILURE_SCOPE_UNKNOWN = 0;
+    FAILURE_SCOPE_TASK = 1;        // The task failed, but can be retried.
+    FAILURE_SCOPE_INVOCATION = 2;  // The invocation cannot continue (e.g. invalid input).
+    FAILURE_SCOPE_GRAPH = 3;       // The graph is broken, requiring repair; block invocations.
```
eabatalov: idk if graph scope would ever be useful tbh but okay if it's not more effort to implement this..

src/tensorlake/function_executor/proto/function_executor.proto
```diff
@@ -111,6 +128,7 @@ message RunTaskResponse {
     optional bool is_reducer = 6;
     optional bool success = 7;
```
eabatalov, line 132: Let's mark success field for deletion then and stop using it in Executor?

src/tensorlake/function_executor/proto/function_executor.proto
```diff
@@ -111,6 +128,7 @@ message RunTaskResponse {
...
     optional Metrics metrics = 8;
+    optional FailureInfo failure = 9;  // Implies !success
```
eabatalov, line 134: Let's stop using success here and say that if this is not None then the task failed.

src/tensorlake/function_executor/proto/function_executor.proto
```diff
@@ -102,6 +102,23 @@ message Metrics {
...
 }
 
+enum FailureScope {
+    FAILURE_SCOPE_UNKNOWN = 0;
+    FAILURE_SCOPE_TASK = 1;        // The task failed, but can be retried.
+    FAILURE_SCOPE_INVOCATION = 2;  // The invocation cannot continue (e.g. invalid input).
+    FAILURE_SCOPE_GRAPH = 3;       // The graph is broken, requiring repair; block invocations.
+}
+
+message FailureInfo {
```
eabatalov, line 115: Why do we use "failure" instead of "error" term? I think error is much wider used,
Do we need Info suffix here, what value it adds?

This message would be useful in all cases where we run customer code. We run it in Task and in FE initialization. So we can't call it TaskError. We can call it CustomerCodeError. And let's use it both in RunTaskResponse and in InitializeResponse?

Calling it CustomerCodeError will prevent scope creep for it and makes it more clear on how to use it in future code maintenance.

src/tensorlake/function_executor/proto/function_executor.proto
```diff
@@ -102,6 +102,23 @@ message Metrics {
...
+    // Scope determines the overall system's behavior in response to this failure.
+    optional FailureScope scope = 1;
+
+    // Cls, Msg, and Trace are human-comprehensible diagnostics.
+    optional string cls = 2;
+    optional string msg = 3;
```
eabatalov, line 118: All the human comprehensive diagnostics is currently going into stderr.
With this change we're going very inconsistent with how we report human readable data. I think we should delete it from here and just add it to stderr until we decided that we're not using stderr and I won't expect this to happen for now.

src/tensorlake/function_executor/proto/function_executor.proto
eabatalov: But still set success for backward compatibility.

src/tensorlake/function_executor/handlers/run_function/response_helper.py
```diff
@@ -64,6 +66,18 @@ def failure_response(
             success=False,
         )
 
+        response.failure.scope = ProtoFailureScope.FAILURE_SCOPE_TASK
+        if failure.scope == FailureScope.Invocation:
+            response.failure.scope = ProtoFailureScope.FAILURE_SCOPE_INVOCATION
+        elif failure.scope == FailureScope.Graph:
+            response.failure.scope = ProtoFailureScope.FAILURE_SCOPE_GRAPH
+
+        response.failure.cls = failure.cls
```
eabatalov, line 72: Why do we need to save all this human readable data in response object if we already saved it to stderr and customers will see it in stderr?

src/tensorlake/error.py
earhart: Done

src/tensorlake/function_executor/handlers/run_function/handler.py
earhart: Done

src/tensorlake/function_executor/handlers/run_function/response_helper.py
earhart: Done -- TBH as a native English speaker I find both fine, but if the suffix is easier for you I'm definitely good with that.  :-)

src/tensorlake/function_executor/proto/function_executor.proto
earhart: Removed

src/tensorlake/function_executor/proto/function_executor.proto
earhart: Yeah, once these PRs land and are deployed I think we can go back and remove this field.

src/tensorlake/function_executor/proto/function_executor.proto
earhart: +1 -- once these land and are deployed we should remove the success field.

src/tensorlake/function_executor/proto/function_executor.proto
earhart: It doesn't matter that much to me, but IMHO... an error is a very specific problem, like an exception being thrown; what the error causes is a failure of the overall task, and it's the task failure that we're reporting here.  <shrug>.  But like I said, I could go either way on this one if you think it's more clear some other way.

src/tensorlake/function_executor/handlers/run_function/response_helper.py
earhart: stderr is a bit much to go through at times, and keeping the failure cause separate makes it more straightforward to attach to the invocation failure and graph failure (I don't know that we want to be attaching all of stderr to those).

eabatalov:
This PR was replaced by https://github.com/tensorlakeai/tensorlake/pull/205

PR #193, opened 2025-06-19
Update event attributes to be consistent

j3m7:


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "tensorlake"
-version = "0.2.2"
```
eabatalov: Need to rebase the PR on main

src/tensorlake/function_executor/handlers/run_function/handler.py
```diff
@@ -36,7 +36,7 @@ def __init__(
         self._invocation_state: InvocationState = invocation_state
         self._logger = logger.bind(
             module=__name__,
-            graph_invocation_id=request.graph_invocation_id,
+            invocation_id=request.graph_invocation_id,
             task_id=request.task_id,
```
eabatalov, line 40: Let's add a new optional field allocation_id to the request and add it to the logger here.

eabatalov: We'll also need to set allocation_id on host Executor side in the request.

PR #202, opened 2025-06-24
Signature Detection Example Update

Itz-Antaripa:
- Updated example with langchain-tensorlake
- Removed helper functions


examples/signature-detection/langgraph-agent-signature-detection-cli/requirements.txt
```diff
@@ -3,4 +3,5 @@ langchain>=0.1.0
 langchain-openai>=0.1.0
 langgraph>=0.1.0
 tensorlake>=0.1.0
+langchain-tensorlake
```
drguthals: We only need `langchain-tensorlake` here because it will bring in the rest

examples/signature-detection/langgraph-agent-signature-detection-cli/signature_detection_langgraph_agent.py
```diff
@@ -215,164 +55,94 @@ class SignatureConversationAgent:
...
+    path = "path/to/your/document.pdf"
+
+    analysis_questions = [
+        "How many signatures were detected in this document and who are the parties involved?",
+        "What contextual information can you extract about any signatures?",
+        "Are there any missing signatures on any pages?",
+    ]
+
+    # run the example
+    asyncio.run(signature_detection_real_estate(path, analysis_questions))
```
drguthals: If we want to move the path and questions to this part instead of in the above function, let's also update the tutorial to reflect this.

src/tensorlake/function_executor/service.py
```diff
@@ -72,7 +72,7 @@ def initialize(
             graph_version=request.graph_version,
             fn=request.function_name,
         )
-        
+
```
drguthals: Despite this being a nothing change, lets remove this file change from this PR since it seems to have just been a mistake.

Itz-Antaripa: `make fmt` before the push made that change.

PR #207, opened 2025-06-27
feat: parse v2

drguthals:
- [x] Matching the parsing data models to the v2 api

## Contribution Checklist


src/tensorlake/documentai/parse.py
```diff
@@ -24,30 +24,42 @@ class ChunkingStrategy(str, Enum):
...
+    Supported MIME types for document parsing.
+
+    PDF: Portable Document Format files.
+    DOCX: Microsoft Word documents.
+    PPTX: Microsoft PowerPoint presentations.
+    KEYNOTE: Apple Keynote presentations.
+    JPEG: JPEG image files.
+    TEXT: Plain text files.
+    HTML: HTML files.
+    XLSX: Microsoft Excel spreadsheets.
```
diptanu: CSV, XLS

diptanu: PNG

src/tensorlake/documentai/parse.py
```diff
@@ -24,30 +24,42 @@ class ChunkingStrategy(str, Enum):
...
+class FormDetectionMode(str, Enum):
     """
-    Algorithm to use for parsing tables in a document.
+    Algorithm to use for detecting forms in a document.
 
-    TSR: Table Structure Recognition. Great for structured tables.
-    VLM: Visual Layout Model. Great for unstructured tables or semi-structured tables.
+    VLM: Uses a VLM to identify questions and answers in a form.
+         Does not provide bounding boxes and is prone to hallucinations.
+    OBJECT_DETECTION: Uses a layout detector to identify questions and answers.
```
diptanu: Lets make it TSR instead of object detection: Uses structure detection to understand the layout of the form to identify questions and answers.

src/tensorlake/documentai/parse.py
```diff
@@ -98,18 +132,70 @@ class ParsingOptions(BaseModel):
...
-    extraction_options: Optional[ExtractionOptions] = None
-    deliver_webhook: bool = False
-    detect_signature: Optional[bool] = False
-    table_summary: Optional[bool] = False
-    figure_summary: Optional[bool] = False
     disable_layout_detection: Optional[bool] = False
     form_detection_mode: Optional[FormDetectionMode] = (
         FormDetectionMode.OBJECT_DETECTION
     )
+    remove_strikethrough: bool = False
```
diptanu: remove_strikethrough_lines

src/tensorlake/documentai/parse.py
```diff
@@ -98,18 +132,70 @@ class ParsingOptions(BaseModel):
...
+class StructuredExtractionOptions(BaseModel):
+    """
+    Options for structured data extraction from a document.
+    """
+
+    chunking_strategy: Optional[ChunkingStrategy] = None
+    json_schema: Union[Type[BaseModel], Json] = Field(..., alias="schema")
+    model_provider: ModelProvider = ModelProvider.TENSORLAKE
+    page_class: Optional[str] = None
+    page_class_definition: Optional[str] = None
```
diptanu: Page Class defintion should be outside of structured extraction options. Its a separate request object.

src/tensorlake/documentai/parse.py
```diff
@@ -98,18 +132,70 @@ class ParsingOptions(BaseModel):
...
+        None,
+        description='Labels to attach to the parse operation. These labels can be used to store metadata about the parse operation. The format should be a JSON object, e.g. {"key1": "value1", "key2": "value2"}.',
+    )
+    mime_type: Optional[MimeType] = Field(
+        None,
+        description="The MIME type of the document being parsed. It is optional if `file_id` or `file_url` are provided, as the MIME type will be inferred from the file extension or content.",
+    )
+    page_range: Optional[str] = Field(
+        None,
+        description='The range of pages to parse in the document. This should be a comma-separated list of page numbers or ranges (e.g., "1,2,3-5"). If not provided, all pages will be parsed.',
```
diptanu: Write a note about inclusive vs exclusive page ranges.

src/tensorlake/documentai/parse.py
```diff
@@ -98,18 +132,70 @@ class ParsingOptions(BaseModel):
...
+    )
+    enrichment_options: Optional[EnrichmentOptions] = Field(
+        None,
+        description="Options for enriching a document with additional information.",
+    )
+    parsing_options: Optional[ParsingOptions] = Field(
+        None,
+        description="Additional options for tailoring the document parsing process.",
+    )
+    structured_extraction_options: Optional[StructuredExtractionOptions] = Field(
```
diptanu: We are missing page classifcation request, and we need to represent a list of structured extraction request. We should also verify that the schema names are different when people are sending multiple structured extraction requests.

src/tensorlake/documentai/parse.py
drguthals: Wait.. that doesn't make any sense. We need to talk about that if that is the case @diptanu 

Our understanding is that a structured data extraction object can have a page_class which is unique the this specific structured data extraction object (e.g. schema). That is how the API is also written. Are you saying that is different?

src/tensorlake/documentai/parse.py
drguthals: If we want this to be `remove_strikethrough_lines` we should update the API to match. Right now the API is `remove_strikethrough`: https://github.com/tensorlakeai/inkwell-server/blob/4a48f1c530e25cfc316a19fb01419e7b1e49d5d8/server/src/api/v2/routes/parse/request_http_objects.rs#L374

Why add the lines? Are you saying the entire line where a strikethrough exists will be removed? My understanding is only the words that have a strikethrough will be removed. If my understanding is correct, this should not be `remove_strikethrough_lines`

src/tensorlake/documentai/parse.py
drguthals: If this is TSR then this also needs to be changed in the API: https://github.com/tensorlakeai/inkwell-server/blob/4a48f1c530e25cfc316a19fb01419e7b1e49d5d8/server/src/api/v2/routes/parse/request_http_objects.rs#L460

I don't think this one is TSR, this is for Form Detection, not Table Parsing..

src/tensorlake/documentai/parse.py
drguthals: WRT "page classification request" - from the API and SDK perspective (e.g. the user perspective) I do not think this is a separate request.

src/tensorlake/documentai/parse.py
ilbum: Agreed, good callout.

src/tensorlake/documentai/parse.py
ilbum: Quick note, I also believe we should prevent double-negatives in our semantics if possible.

src/tensorlake/documentai/client.py
```diff
@@ -44,166 +55,283 @@ def __headers__(self):
...
+        page_range: Optional[str] = None,
+        labels: Optional[dict] = None,
+        mime_type: Optional[MimeType] = None,
+    ) -> dict:
+        payload = {}
+
+        # file check
+        if file.startswith("http://") or file.startswith("https://"):
+            payload["file_url"] = file
+        elif file.startswith("tensorlake-"):
```
drguthals, line 77: We were going to change this to be `file-` instead of `tensorlake-`. It's an internal change that can be a follow. I've opened an issue to change in the backend.

src/tensorlake/documentai/client.py
```diff
@@ -44,166 +55,283 @@ def __headers__(self):
...
-        """
-        response = self._client.get(
-            url=f"jobs/{job_id}",
-            headers=self.__headers__(),
-        )
-        response.raise_for_status()
-        return Job.model_validate(response.json())
+    def __create_parse_req__(
+        self,
+        file: str,
```
drguthals, line 61: Since we can also put raw text in here, I don't think naming this "file" makes sense. Maybe this should be "data"?

src/tensorlake/documentai/client.py
```diff
@@ -44,166 +55,283 @@ def __headers__(self):
...
+        if enrichment_options:
+            payload["enrichment_options"] = enrichment_options.model_dump(
+                exclude_none=True
+            )
 
-    async def get_job_async(self, job_id: str) -> Job:
+        if page_classifications:
+            payload["page_classifications"] = [
+                page_classification.model_dump(exclude_none=True)
+                for page_classification in page_classifications
```
drguthals, line 104: Woah...I didn't know you could do for loops like this in Python 🤯

src/tensorlake/documentai/client.py
drguthals: Fast follow -> Make `raw_text` a separate parameter and add logic to ensure at least a `file` or a `raw_text` is provided

src/tensorlake/documentai/client.py
```diff
@@ -44,166 +55,283 @@ def __headers__(self):
...
         """
         Wait for a job to complete.
         """
-        job = self.get_job(job_id)
-        finished_job = job
-        while finished_job.status in ["pending", "processing"]:
+        parse = self.get_parse(parse_id)
+        finished_parse = parse
+        print(f"Waiting for job {parse} to complete...")
+        while finished_parse["status"] in ["pending", "in_progress"]:
```
drguthals: Did we update `processing` to `in_progress`?

src/tensorlake/documentai/client.py
```diff
@@ -44,166 +55,283 @@ def __headers__(self):
...
+    async def wait_for_completion_async(self, parse_id: str) -> ParseResult:
         """
         Wait for a job to complete asynchronously.
         """
-        job = await self.get_job_async(job_id)
-        finished_job = job
-        while finished_job.status in ["pending", "processing"]:
+        parse = await self.get_parse_async(parse_id)
+        finished_parse = parse
+        while finished_parse["status"] in ["pending", "in_progress"]:
```
drguthals: Did we update `processing` to `in_progress`?

src/tensorlake/documentai/models/enums.py
```diff
@@ -0,0 +1,121 @@
...
+    GPT4OMINI = "gpt-4o-mini"
+
+
+class ParseStatus(str, Enum):
+    """
+    Status of a parse job in the v2 API.
+    """
+
+    PENDING = "pending"
+    IN_PROGRESS = "in_progress"
```
drguthals: Why is this one "in_progress" but the other is "processing", why did we change this?

src/tensorlake/documentai/models/enums.py
```diff
@@ -0,0 +1,121 @@
...
+class TableOutputMode(str, Enum):
+    """
+    Output mode for tables in a document.
+
+    JSON: The table is returned in JSON format.
+    MARKDOWN: The table is returned in Markdown format.
+    HTML: The table is returned in HTML format.
+    """
+
+    JSON = "json"
```
drguthals: I think this was meant to be removed.
```suggestion

```

drguthals: @diptanu we wanted tables to only be Markdown or HTML right?

src/tensorlake/documentai/models/enums.py
```diff
@@ -0,0 +1,121 @@
...
+    IN_PROGRESS = "in_progress"
+    COMPLETED = "completed"
+    FAILED = "failed"
+
+
+class TableOutputMode(str, Enum):
+    """
+    Output mode for tables in a document.
+
+    JSON: The table is returned in JSON format.
```
drguthals:
```suggestion

```

If my comment below is correct.

src/tensorlake/documentai/models/enums.py
drguthals: That's what matches our API: https://github.com/tensorlakeai/inkwell-server/blob/0431d13a52144de568aa7d011b199bcef906c747/server/src/api/v2/routes/parse/request_http_objects.rs#L526

src/tensorlake/documentai/models/enums.py
```diff
@@ -0,0 +1,121 @@
...
+    Algorithm to use for detecting forms in a document.
+
+    VLM: Uses a VLM to identify questions and answers in a form.
+         Does not provide bounding boxes and is prone to hallucinations.
+    TSR: Uses a layout detector to identify questions and answers.
+         Does not work well with very complex forms.
+    """
+
+    VLM = "vlm"
+    TSR = "tsr"
```
drguthals: Are we sure `FormDetectionMode` is `VLM` or `TSR`? Isn't `TSR` for tables?

```suggestion
    ObjectDetection = "option_detection"
```

https://github.com/tensorlakeai/inkwell-server/blob/0431d13a52144de568aa7d011b199bcef906c747/server/src/api/v2/routes/parse/request_http_objects.rs#L568

src/tensorlake/documentai/models/options.py
```diff
@@ -0,0 +1,146 @@
...
+    )
+    disable_layout_detection: bool = Field(
+        False,
+        description="Useful flag for documents with a lot of tables or images. If set to `true`, the API will skip the layout detection step, and directly extract text from the document.",
+    )
+    form_detection_mode: FormDetectionMode = Field(
+        FormDetectionMode.VLM,
+        description="Algorithm to use for detecting forms in a document.",
+    )
+    remove_strikethrough_lines: bool = Field(
```
drguthals, line 58: Didn't we decide to have this be `remove_strikethrough`?

```suggestion
    remove_strikethrough: bool = Field(
```

src/tensorlake/documentai/client.py
diptanu: This is an internal api, probably doesn’t matter. I hope the parse api resembles what we have on HTTP api?

src/tensorlake/documentai/models/enums.py
diptanu: Yes just html and md

src/tensorlake/documentai/models/enums.py
diptanu: It used to be object detection before. TSR is closer to what happens internally, we retain the structure of the form.

src/tensorlake/documentai/models/options.py
diptanu: I think remove strikethrough lines is fine since it clearly explains what is going on? I am fine with whatever is easier to understand.

src/tensorlake/documentai/models/enums.py
ilbum: Don't know why, but I matched it with the API.

src/tensorlake/documentai/models/enums.py
ilbum: Done

src/tensorlake/documentai/models/enums.py
ilbum: We changed it to TSR at the API level.

src/tensorlake/documentai/client.py
ilbum: Yes I built it around what we have on HTTP API. 👍

src/tensorlake/documentai/models/enums.py
drguthals: Hmmm ok - @miguelhrocha can you chime in as to why we changed that in the API?

PR #217, opened 2025-07-03
feat: parse results updates

ilbum:
## Description

- Refactor of Data Models for Document AI
- Updated get_parse and get_parse_async methods in client.py to return validated ParseResult Pydantic model instances instead of raw JSON.

## Contribution Checklist


drguthals (approved):
🚢

PR #225, opened 2025-07-07
feat: Datasets V2

miguelhrocha:
## Contribution Checklist


diptanu (approved):

src/tensorlake/documentai/client.py
```diff
@@ -394,256 +436,236 @@ def delete_file(self, file_id: str):
...
-            headers=self.__headers__(),
-            params={"cursor": cursor} if cursor else None,
-        )
-        response.raise_for_status()
-        result = PaginatedResult[Job].model_validate(response.json())
-        return result
+        # file check
+        if file.startswith("http://") or file.startswith("https://"):
+            payload["file_url"] = file
+        elif file.startswith("tensorlake-"):
```
drguthals, line 650: This makes sense for now, but I think this is something we need to update in the future. For example, what if I sent:
```
file="tensorlake-contract.pdf"
```
Or:
```
file=tensorlake-data/myfile.pdf
```

Then would this assume that this is a file_id?

src/tensorlake/documentai/models/enums.py
```diff
@@ -121,3 +121,8 @@ class TableParsingFormat(str, Enum):
 
     TSR = "tsr"
     VLM = "vlm"
+
+
+class DatasetStatus(str, Enum):
```
drguthals, line 126: So a dataset's status is `idle` if it's not currently parsing any documents and `processing` if it is?

src/tensorlake/documentai/client.py
drguthals: TODO:
1. @drguthals Ensure docs specify that Datasets is not uploading documents and so this needs to be either a URL, FileID, or Raw Test
2. @miguelhrocha Ensure that this is included in the test cases and that when someone does do this we will error out saying that `file_id 'tensorlake-contract.pdf' is not valid` (or whatever the error message says)

src/tensorlake/documentai/models/enums.py
miguelhrocha: Correct!

PR #240, opened 2025-07-10
feat: automatically upload file if user provided a file_path as file input

miguelhrocha:
### Description

To reduce friction when using the SDK, we will automatically identify if the user is trying to use a local filepath as input for the parse operation, be it from a dataset or directly.

The SDK will upload the file to the associated Project and use it as parse input.

Test run:
https://github.com/tensorlakeai/tensorlake/actions/runs/16199420480

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "tensorlake"
-version = "0.2.20"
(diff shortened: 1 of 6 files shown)


diptanu:
Closing this as this deviates from the REST API. I have merged a PR which throws exception if the path is local

PR #241, opened 2025-07-10
RFC DON'T MERGE: grpc stream based protocol for running task allocations on Function Executor

eabatalov:
I'm happy with naming in the protocol and messages.
Please give me any feedback before I progress further into the implementation.


src/tensorlake/function_executor/proto/function_executor.proto
```diff
@@ -2,6 +2,15 @@ syntax = "proto3";
...
 package function_executor_service;
 
+message Status {
+  // The status code, which should be an enum value of google.rpc.Code.
+  // See https://cloud.google.com/tasks/docs/reference/rpc/google.rpc
+  int32 code = 1;
+
+  // A human-readable error message explaining the error if there's any.
+  string message = 2;
+}
```
earhart, line 12: Maybe `import "google/protobuf/any.proto";`, and add a `repeated google.protobuf.Any details = 3;` here to carry additional structured info?  (Longer term we should probably break this out into a separate proto we can use in other APIs, but that can happen when we need it.)

Google's [status.proto](https://github.com/googleapis/api-common-protos/blob/main/google/rpc/status.proto) and [code.proto](https://github.com/googleapis/api-common-protos/blob/main/google/rpc/code.proto) are Apache-licensed; I think we could just copy them directly.

src/tensorlake/function_executor/proto/function_executor.proto
```diff
@@ -161,6 +173,128 @@ message InfoResponse {
...
 }
 
+// === run_task_allocations_session RPC === //
+
+message SerializedObjectID {
+    optional string id = 1;
+}
+
+message SerializedObjectManifest {
+    optional SerializedObjectID id = 1;
```
earhart, line 183: How does the FE choose an ID?  And do these need to be strings?

src/tensorlake/function_executor/proto/function_executor.proto
```diff
@@ -161,6 +173,128 @@ message InfoResponse {
...
+    optional SerializedObjectEncoding encoding = 2;
+    optional uint64 encoding_version = 3;
+    optional uint64 size = 4;
+    optional string sha256_hash = 5;
+    optional uint32 chunk_count = 6;
```
earhart, line 188: Do we need a chunk count?

src/tensorlake/function_executor/proto/function_executor.proto
```diff
@@ -161,6 +173,128 @@ message InfoResponse {
...
+}
+
+message SerializedObjectChunk {
+    // The ID of the serialized object this chunk belongs to.
+    optional SerializedObjectID id = 1;
+    // The chunk data.
+    optional bytes data = 2;
+    // The index of the chunk in the serialized object. 0 based.
```
earhart, line 196: It's probably better to use a data offset here, instead of a chunk index - requires less computation on everyone's part.

src/tensorlake/function_executor/proto/function_executor.proto
```diff
@@ -161,6 +173,128 @@ message InfoResponse {
...
+
+message SerializedObjectID {
+    optional string id = 1;
+}
+
+message SerializedObjectManifest {
+    optional SerializedObjectID id = 1;
+    optional SerializedObjectEncoding encoding = 2;
+    optional uint64 encoding_version = 3;
+    optional uint64 size = 4;
```
earhart, line 186: Is this optional for cases where we don't know the eventual size, like error logs?

src/tensorlake/function_executor/proto/function_executor.proto
```diff
@@ -161,6 +173,128 @@ message InfoResponse {
...
+message TaskAllocationOutput {
+    optional string graph_invocation_id = 1;
+    optional string task_id = 2;
+    optional string allocation_id = 3;
+    optional TaskOutcomeCode outcome_code = 4;
+    optional TaskFailureReason failure_reason = 5;
+    repeated SerializedObjectID function_outputs = 6;
+    // User payload for invocation error if task failed with invocation error.
+    optional SerializedObjectID invocation_error_output = 7;
+    optional SerializedObjectID stdout = 8;
```
earhart, line 273: This means we don't know the ID of the output / error streams until the allocation completes.

src/tensorlake/function_executor/proto/function_executor.proto
```diff
@@ -161,6 +173,128 @@ message InfoResponse {
...
+message SerializedObjectChunk {
+    // The ID of the serialized object this chunk belongs to.
+    optional SerializedObjectID id = 1;
+    // The chunk data.
+    optional bytes data = 2;
+    // The index of the chunk in the serialized object. 0 based.
+    optional uint32 index = 3;
+}
+
+// Used for uploading a serialized object from initiating side to the accepting side.
```
earhart, line 200: It seems a little odd to have two completely different forms of serialized objects; can we combine them?

src/tensorlake/function_executor/proto/function_executor.proto
eabatalov: Both sides actually choose IDs themself. They don't have to be strings but strings give flexibility and readability for humans for free so I'd use them.

src/tensorlake/function_executor/proto/function_executor.proto
eabatalov: Yep, this allows to avoid sending "file transmission finished" message. Chunk size is not fixed in the protocol.

src/tensorlake/function_executor/proto/function_executor.proto
eabatalov: I'd actually better remove the chunk ix as chunks are sent one by one in order. No index makes this more clear.

src/tensorlake/function_executor/proto/function_executor.proto
eabatalov: I'll remove chunk count too, both sides can understand if they got all the data by comparing the size they have recieved.

src/tensorlake/function_executor/proto/function_executor.proto
eabatalov: When we do live log streaming I think we'd be using a different manifest message or explicitly document some fields like size here as optional. This is easy to do.

src/tensorlake/function_executor/proto/function_executor.proto
eabatalov: When doing live log streaming it'd be initiated by a different message that wraps a serialized object manifest. The wrapper message will clarify which obj id is stdout and which is stderr, then the chunks for them will follow.
I think this even would be fully backward compatible.

src/tensorlake/function_executor/proto/function_executor.proto
eabatalov: You mean 2 types is chunked serialized object and full SerializedObject? The full SerializedObject is kept without any changes for backward compatibility. We have to keep it this way for now.

src/tensorlake/function_executor/proto/function_executor.proto
eabatalov: Thank you. I'll copy them. Dealing with int code is harder indeed.

PR #244, opened 2025-07-11
RFC DON'T MERGE: stream based protocol for running tasks allocations in Function Executor

eabatalov:


src/tensorlake/function_executor/proto/function_executor.proto
```diff
@@ -170,10 +299,34 @@ service FunctionExecutor {
...
+    // Deprecated.
     rpc initialize_invocation_state_server(stream InvocationStateResponse) returns (stream InvocationStateRequest);
     // Executes the task defined in the request.
     // Multiple tasks can be running in parallel.
+    // Deprecated.
     rpc run_task(RunTaskRequest) returns (RunTaskResponse);
+
+    // Creates session streams for running task allocations on Function Executor.
+    // A typical message flow is:
+    // 1. Client sends OpenSessionRequest to join a session by ID. The session is created if it does not exist.
```
diptanu, line 311: Will there be multiple executor threads talking to the FE to run tasks concurrently?

src/tensorlake/function_executor/proto/function_executor.proto
```diff
@@ -170,10 +299,34 @@ service FunctionExecutor {
...
+    // 2. Function Executor responds with OpenSessionResponse.
+    // 3. Client uploads task allocation inputs using UploadSerializedObjectRequest. Chunks are sent in original order they appear in the data.
+    // 4. Function Executor responds with UploadSerializedObjectResponse for each UploadSerializedObjectRequest.
```
diptanu, line 314: Is there a need for this message? The messages are delivered in order. So executor could send a bunch of of messages in sequence. UploadSerializedObjectRequest, UploadSerializedObjectRequest, RunTaskAllocationsRequest. The handler on the FE won't process these messages out of order.

src/tensorlake/function_executor/proto/function_executor.proto
```diff
@@ -161,6 +166,130 @@ message InfoResponse {
...
+    // The ID of the serialized object this chunk belongs to.
+    optional SerializedObjectID id = 1;
+    // The chunk data.
+    optional bytes data = 2;
+}
+
+// Used for uploading a serialized object from initiating side to the accepting side.
+message UploadSerializedObjectRequest {
+    oneof message {
+        SerializedObjectManifest manifest = 1;
```
diptanu, line 193: I am assuming we are going to send a stream of these UploadSerializedObjectRequest, there would be a lot of repetion of SerializedObjectManifest objects

src/tensorlake/function_executor/proto/function_executor.proto
```diff
@@ -161,6 +166,130 @@ message InfoResponse {
...
+        SerializedObjectChunk chunk = 2;
+    }
+}
+
+// The accepting side sends a response per request.
+message UploadSerializedObjectResponse {
```
diptanu, line 199: What is the point of responding to every chunk request? Grpc doesn't return a status for every message sent on a bi-directional stream?

src/tensorlake/function_executor/proto/function_executor.proto
eabatalov: Yes, this is the idea. Otherwise we won't be able to implement concurrent function runs (this is not about batching).

src/tensorlake/function_executor/proto/function_executor.proto
eabatalov: If the streams get interrupted for any reason and client re-joins the session then it won't know which parts of the file were uploaded successfully. This can be addressed by re-uploading everything again. But this is questionable decision.

src/tensorlake/function_executor/proto/function_executor.proto
eabatalov: Not sure what repetition means. One manifest per serialized object and then chunks. Chunks only contain SO id and data. They only share SO id with the manifest.

src/tensorlake/function_executor/proto/function_executor.proto
eabatalov: Sure it doesn't, it's just two streams, one in both directions, not bidirectional rpcs.

eabatalov:
This is not needed anymore, we have a different protocol implemented.

PR #250, opened 2025-07-12
FE stream protocol

diptanu:
A proposal for streaming IO between Executors and FE to support large files, health checks and progress updates.


diptanu:
Closing this as @eabatalov is working on this separately

PR #253, opened 2025-07-14
RFC DON'T MERGE implementation of Function Executor side S3 direct get/put

eabatalov:
Need to keep SerializedObject around for:
* InvocationState, we're passing its serialized data back and worth to Server and we don't use S3 for this.
* Graph ZIP. We use host side graph caching so we don't download it directly from S3 in Function Executor.

The FE tests are passing with local blob store, S3 blob store implementation TBD.


src/tensorlake/function_executor/proto/function_executor.proto
```diff
@@ -125,17 +153,15 @@ enum TaskFailureReason {
...
-    optional Metrics metrics = 7;
-    optional TaskOutcomeCode outcome_code = 8;
-    optional TaskFailureReason failure_reason = 9;
+    optional bool is_reducer = 4;
+    optional Metrics metrics = 5;
+    optional TaskOutcomeCode outcome_code = 6;
+    optional TaskFailureReason failure_reason = 7;
     // User payload for invocation error if task failed with invocation error.
-    optional SerializedObject invocation_error_output = 10;
+    optional SerializedObjectInsideBLOB invocation_error_output = 8;
```
diptanu, line 164: Not a comment but just a clarification. We are storing the error from the functions inside the URL for outputs? Makes sense to do it that way

diptanu (commented):
Perfect :)

src/tensorlake/function_executor/proto/function_executor.proto
eabatalov: It's an S3 object, this allows to remove limits on InvocationError.message size and let users dump any json they want there. Server already stores this as DataPayload. Might not be what you expected but this is flexible and works. Good enough for now imo.

eabatalov: These InvocationError.message strings are also customer data and we handle it consistently (in S3 blobs). I think this is more important for long term to be consistent.

eabatalov: Also answering your question, I guess I see what you meant.
Yes, we store InvocationError.message in the BLOB used for function outputs.

PR #261, opened 2025-07-18
fix: ignore_sections serialization

diptanu:
There was a bug in serializing the ingore_sections field in ParseOptions


diptanu:
Merging this since this fixes a bug in the last released tensorlake package

PR #264, opened 2025-07-21
Update tensorlake sdk to v1 api

diptanu:
Updated tensorlake sdk to use the new Indexify server 

This needs to be merged and released after Indexify PR for the new API is merged. Indexify server is backwards compatible to older versions of tensorlake sdk so we can safely release the new server without releasing this package.


src/tensorlake/cli/_common.py
```diff
@@ -12,7 +12,7 @@
     VERSION = "unknown"
 
 
-from tensorlake.http_client import TensorlakeClient
+from tensorlake.functions_sdk.http_client import TensorlakeClient
```
eabatalov, line 15: This breaks all tests across indexify, PE, doc ai, are you sure you want to do this right now?

diptanu: Yeah this is fine. Better to do this now and fix these things before making a public release

j3m7 (approved):

PR #265, opened 2025-07-21
feat: add get_dataset_data method

miguelhrocha:
### Description

Add SDK operation to get the data from a dataset.

Document AI tests:
https://github.com/tensorlakeai/tensorlake/actions/runs/16426628706/job/46418401135


src/tensorlake/documentai/_datasets.py
```diff
@@ -416,6 +417,112 @@ async def update_dataset_async(
...
         return Dataset.model_validate(resp.json())
 
+    def get_dataset_data(
+        self,
+        dataset: Dataset,
+        cursor: Optional[str] = None,
+        direction: Optional[PaginationDirection] = None,
+        dataset_name: Optional[str] = None,
+        limit: Optional[int] = None,
+        file_name: Optional[str] = None,
```
diptanu: What is the point of having a file name here? Looks like you want to be able to list parsed data. Is this like a filter?

src/tensorlake/documentai/_datasets.py
```diff
@@ -416,6 +417,112 @@ async def update_dataset_async(
...
+        status: Optional[ParseStatus] = None,
```
diptanu: Same here, looks like these are all filters. May be create a filter object to make this clear?

diptanu (changes requested):

src/tensorlake/documentai/_datasets.py
```diff
@@ -416,6 +417,50 @@ async def update_dataset_async(
         resp = await self._arequest("PUT", f"/datasets/{dataset.dataset_id}", json=body)
         return Dataset.model_validate(resp.json())
 
+    def get_dataset_data(
+        self,
+        dataset: Dataset,
```
drguthals: Why send the dataset object instead of the dataset ID? That seems to break all of our other conventions.

drguthals (approved):

PR #271, opened 2025-07-22
RFC protocol for FE S3 direct and no function stdout, stderr capturing

eabatalov:
Need to keep SerializedObject around for:
* InvocationState, we're passing its serialized data back and worth to Server and we don't use S3 for this.
* Graph ZIP. We use host side graph caching so we don't download it directly from S3 in Function Executor.

The FE protocol now explicitly separates read only blobs and write only blobs. This is because in S3 we have to work with them very differently. We can read a blob of any size from S3 but to upload a blob we need to chunk it into 5 GB chunks. To support this Executor provides a list of presigned URIs per chunk, Executor controls how many chunks it creates so we can control the max blob size in the system easily. Later when we have a streaming based protocol FE can request the presigned chunks from Executor on demand.

We discussed that Executor will be capturing full FE stdout, stderr via file descriptors or local RPC to e.g. Docker daemon. Executor will treat FE stdout, stderr as customer data and upload/stream it where applicable. To implement this design we change the following in FE:

* Stop capturing function stdout, stderr so everything it logs goes directly to FE stdout, stderr.
* Add a new module for FE logging and add the FE log in FE diagnostics messages included in initialize and run task responses. These FE logs are guaranteed to contain no customer data so we can log them on Executor side for easier debugging of what's happening in FE. Later when we have streaming protocol we can send these diagnostic logs to Executor at the moment they were produced so it's easier to debug with them.
* In near term Executor would still be uploading FE stdout, stderr to allocs' stdout, stderr S3 blobs but in longer term we would change it. stdout, stderr Redis stream consumer would parse the stream and cut it into individual allocations' stdoutputs. I'm emmiting these events to FE stdout already for the moment when we need to use them.


miguelhrocha (approved):

eabatalov:
The test failures are expected as this PR depends on Indexify update.

PR #275, opened 2025-07-23
Add config file support

earhart:


pyproject.toml
```diff
@@ -21,6 +21,8 @@ click = "8.2.1"
 retry = "^0.9.2"
 aiofiles = "^24.1.0"
 tqdm = "^4.67.1"
+tomli = "^2.2.1"
```
diptanu: Can you see if we can just use a text file? Adding dependencies into this package makes it difficult for customers to build applications with tensorlake, as they are tied into our dependencies. 

Eventually we can get rid of - python-magic, docker, nanoid, pyyaml and aiofiles, from here.

earhart: Done

diptanu (approved):

PR #283, opened 2025-07-24
Feat: Parse structured extraction also accepts single object

ilbum:
## Description

- Parse accepts both a single `StructuredExtractionOptions` object and of `StructuredExtractionOptions`
- Adds test case for single object structured extraction parse request
- Removes deprecated `ParseRequestOptions`

### DIFF
--- a/src/tensorlake/documentai/_parse.py
+++ b/src/tensorlake/documentai/_parse.py
@@ -4,7 +4,7 @@ import asyncio
 import inspect
 import json
 import time
-from typing import Any, Dict, List, Mapping, Optional
+from typing import Any, Dict, List, Mapping, Optional, Union
 
 from pydantic import BaseModel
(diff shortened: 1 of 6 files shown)


src/tensorlake/documentai/_parse.py
```diff
@@ -497,10 +499,17 @@ def _create_parse_req(
             pc.model_dump(exclude_none=True) for pc in page_classifications
         ]
 
+    print("structured_extraction_options:", structured_extraction_options)
```
diptanu: Remove print

diptanu (approved):
Remove the print before merging

src/tensorlake/documentai/_parse.py
ilbum: Done

PR #284, opened 2025-07-25
Add async task RPCs & progress reports

earhart:


src/tensorlake/function_executor/handlers/run_function/handler.py
```diff
@@ -27,6 +28,7 @@ def __init__(
         function_stderr: io.StringIO,
         graph_metadata: ComputeGraphMetadata,
         logger: Any,
+        progress_reporter: ProgressReporter | None,
```
eabatalov: Nit. logger is always the last argument in all the function so far :P

eabatalov: Please use Optional[ProgressReporter]

src/tensorlake/function_executor/proto/function_executor.proto
```diff
@@ -93,6 +96,12 @@ message InvocationStateResponse {
     }
 }
 
+message ListTasksRequest {}
+
+message ListTasksResponse {
+  repeated RunTaskResponse tasks = 1;
```
eabatalov: RunTaskResponse is a result of running a finished task, it's not a task. It's not a correct data model to use RunTaskResponse instead of Task object here.

src/tensorlake/function_executor/proto/function_executor.proto
```diff
@@ -123,6 +132,18 @@ enum TaskFailureReason {
     TASK_FAILURE_REASON_INVOCATION_ERROR = 3;
 }
 
+message ProgressUpdate {
+    float current = 1;
+    float total = 2;
```
eabatalov, line 137: Why float?

src/tensorlake/function_executor/proto/function_executor.proto
```diff
@@ -173,7 +206,18 @@ service FunctionExecutor {
...
+    // This is logically equivalent to start_task / await_task / delete_task.
     rpc run_task(RunTaskRequest) returns (RunTaskResponse);
+    // Lists the currently-running tasks.
+    rpc list_tasks(ListTasksRequest) returns (ListTasksResponse);
+    // Starts the task specified in the request.
+    rpc start_task(RunTaskRequest) returns (StartTaskResponse);
+    // Waits for the indicated task to reach a completed state
+    // (outcome != unknown), returning a stream of task progress
+    // updates.
+    rpc await_task(AwaitTaskRequest) returns (stream TaskStreamResponse);
```
eabatalov: Maybe task_events() returns (stream TaskEvent);

src/tensorlake/function_executor/service.py
```diff
@@ -41,13 +44,34 @@
...
 )
 from .proto.function_executor_pb2_grpc import FunctionExecutorServicer
 from .std_outputs_capture import flush_logs, read_till_the_end
 
 
+class _TaskExecution:
+    def __init__(self, request: RunTaskRequest):
+        self.request = request
+        self.thread: threading.Thread | None = None
+        self.response: RunTaskResponse = RunTaskResponse(
```
eabatalov: Please use Optional[], it's used everywhere else in python code base.

src/tensorlake/function_executor/service.py
```diff
@@ -230,17 +255,20 @@ def run_task(
...
-            invocation_state=ProxiedInvocationState(
-                request.task_id, self._invocation_state_proxy_server
-            ),
-            function_wrapper=self._function_wrapper,
-            function_stdout=self._function_stdout,
-            function_stderr=self._function_stderr,
-            graph_metadata=self._graph_metadata,
-            logger=self._logger,
-        ).run()
+    def run_task(
```
eabatalov, line 325: Nice backward compatibility :)

src/tensorlake/function_executor/service.py
```diff
@@ -256,3 +284,165 @@ def get_info(
...
+        with self._tasks_lock:
+            task_execution = self._running_tasks.get(task_id)
+
+            if task_execution is None:
+                context.abort(
+                    grpc.StatusCode.NOT_FOUND,
+                    f"Task {task_id} not found",
+                )
+
+            with task_execution.updated:
```
eabatalov: Shouldn't we just delete the task here if it's finished and return an error otherwise?
Host Executor already deals with these outcomes and failure reasons and it's not easy. Why do we need to do this here?

src/tensorlake/function_executor/service.py
```diff
@@ -256,3 +284,165 @@ def get_info(
...
+                    task_execution.response.outcome_code
+                    == TaskOutcomeCode.TASK_OUTCOME_CODE_UNKNOWN
+                ):
+                    task_execution.response.outcome_code = (
+                        TaskOutcomeCode.TASK_OUTCOME_CODE_FAILURE
+                    )
+                    # TODO: Consider adding a Cancelled reason here
+                    task_execution.response.failure_reason = (
+                        TaskFailureReason.TASK_FAILURE_REASON_FUNCTION_ERROR
+                    )
```
eabatalov, line 511: If we delete unfinished task we should just return error. We can't stop a running task in FE cause it might be running customer code

src/tensorlake/function_executor/service.py
```diff
@@ -41,13 +44,34 @@
...
+    TaskOutcomeCode,
+    TaskStreamResponse,
 )
 from .proto.function_executor_pb2_grpc import FunctionExecutorServicer
 from .std_outputs_capture import flush_logs, read_till_the_end
 
 
+class _TaskExecution:
+    def __init__(self, request: RunTaskRequest):
+        self.request = request
```
eabatalov: If we copy request fields into this object:
```
namespace=task_info.allocation.task.namespace,
        graph_name=task_info.allocation.task.graph_name,
        graph_version=task_info.allocation.task.graph_version,
        function_name=task_info.allocation.task.function_name,
        graph_invocation_id=task_info.allocation.task.graph_invocation_id,
        task_id=task_info.allocation.task.id,
        allocation_id=task_info.allocation.allocation_id,
        function_input=task_info.input,
```

Then it can become a Task object . Otherwise if we want to store the request then yes TaskExecution is more applicable I agree there.

src/tensorlake/function_executor/handlers/run_function/handler.py
earhart: Done

src/tensorlake/function_executor/service.py
```diff
@@ -256,3 +284,165 @@ def get_info(
...
+        if task_execution is None:
+            context.abort(
+                grpc.StatusCode.NOT_FOUND,
+                f"Task {task_id} not found",
+            )
+
+        # Stream progress updates until task completes
+        response = None
+        while True:
+            with task_execution.updated:
```
eabatalov: Can we just await until response is not None? The response generally shouldn't be available until function completes.

src/tensorlake/function_executor/service.py
```diff
@@ -256,3 +284,165 @@ def get_info(
...
+                with task_execution.updated:
+                    task_execution.progress = ProgressUpdate(
+                        current=progress.current, total=progress.total
+                    )
+                    task_execution.updated.notify_all()
+                # sleep(0) here momentarily releases the GIL, giving
+                # other threads a chance to run.
+                time.sleep(0)
+
+                # TODO: Consider implementing task cancellation here.
```
eabatalov: Can you clarify what is this about? We can't cancel customer code cleanely.

src/tensorlake/function_executor/service.py
```diff
@@ -256,3 +284,165 @@ def get_info(
...
+
+            def progress_reporter(progress: Progress):
+                with task_execution.updated:
+                    task_execution.progress = ProgressUpdate(
+                        current=progress.current, total=progress.total
+                    )
+                    task_execution.updated.notify_all()
+                # sleep(0) here momentarily releases the GIL, giving
+                # other threads a chance to run.
+                time.sleep(0)
```
eabatalov: Let's not keep time.sleep it in the code base unless we have a very clear reason why, otherwise after years ppl would be worried to remove it cause they don't know what's going to happen.

src/tensorlake/function_executor/service.py
```diff
@@ -256,3 +284,165 @@ def get_info(
...
+                ),
+                function_wrapper=self._function_wrapper,
+                function_stdout=self._function_stdout,
+                function_stderr=self._function_stderr,
+                graph_metadata=self._graph_metadata,
+                logger=self._logger,
+                progress_reporter=progress_reporter,
+            ).run()
+
+            if response.outcome_code == TaskOutcomeCode.TASK_OUTCOME_CODE_UNKNOWN:
```
eabatalov: Executor will handle this correctly because TASK_OUTCOME_CODE_UNKNOWN is a possible value that can be returned by FE, there's no need to change it here to something else.

src/tensorlake/function_executor/service.py
```diff
@@ -256,3 +284,165 @@ def get_info(
...
+            task_id=run_request.task_id,
+            outcome_code=TaskOutcomeCode.TASK_OUTCOME_CODE_FAILURE,
+            failure_reason=TaskFailureReason.TASK_FAILURE_REASON_INTERNAL_ERROR,
+            stderr="",
+            stdout="",
+        )
+
+        try:
+
+            def progress_reporter(progress: Progress):
```
eabatalov: `def progress_reporter` can be moved outside of the try block, making try block smaller makes it much more clear what is it for (RunTaskHandler)

src/tensorlake/function_executor/service.py
```diff
@@ -256,3 +284,165 @@ def get_info(
...
+                response.failure_reason = (
+                    TaskFailureReason.TASK_FAILURE_REASON_INTERNAL_ERROR
+                )
+
+        except Exception as e:
+            # Handle any errors by creating a failed task result
+            self._logger.error(
+                "task execution failed in background thread",
+                task_id=run_request.task_id,
+                error=str(e),
```
eabatalov: exc_info=e.

This enables exception formatting configured for the logger. exc_info is used in all code paths.

src/tensorlake/function_executor/service.py
```diff
@@ -256,3 +284,165 @@ def get_info(
...
+
+            if response.outcome_code == TaskOutcomeCode.TASK_OUTCOME_CODE_UNKNOWN:
+                response.outcome_code = TaskOutcomeCode.TASK_OUTCOME_CODE_FAILURE
+                response.failure_reason = (
+                    TaskFailureReason.TASK_FAILURE_REASON_INTERNAL_ERROR
+                )
+
+        except Exception as e:
+            # Handle any errors by creating a failed task result
+            self._logger.error(
```
eabatalov, line 393: This method _execute_task_in_thread needs to create its own local variable `logger` and use it, see how run task handler does this here https://github.com/tensorlakeai/tensorlake/blob/main/src/tensorlake/function_executor/handlers/run_function/handler.py#L33

src/tensorlake/function_executor/service.py
```diff
@@ -256,3 +284,165 @@ def get_info(
         self, request: InfoRequest, context: grpc.ServicerContext
     ) -> InfoResponse:
         return InfoResponse(**info_response_kv_args())
+
+    def _execute_task_in_thread(
+        self, run_request: RunTaskRequest, task_execution: _TaskExecution
+    ):
+        response = RunTaskResponse(
```
eabatalov: Isn't this variable initialization redundant? It's set on all code paths below.
I usually do in this case:
```
response: Optional[RanTaskResponse] = None
```

src/tensorlake/function_executor/service.py
```diff
@@ -256,3 +284,165 @@ def get_info(
...
+
+        except Exception as e:
+            # Handle any errors by creating a failed task result
+            self._logger.error(
+                "task execution failed in background thread",
+                task_id=run_request.task_id,
+                error=str(e),
+            )
+
+            response.stderr = (str(e),)
```
eabatalov: Currently in main branch this stderr should not be set because it's only used for captured function stderr. FE logs go straight to FE stdout and we'll see them there. Otherwise we're leaking FE logs into customer code logs here.

src/tensorlake/function_executor/service.py
earhart: Okay, but longer-term let's change that now that we're on Python 3.10.

src/tensorlake/function_executor/proto/function_executor.proto
eabatalov: One more thing I'm wondering about:

await_task RPC returns stream from server FE to client Executor and right it's sufficient to run a task cause we don't need to send a response back to the task from Executor to FE. This is not exactly true though cause we have get/set context request-response over a separate bidirectional stream. So if we need to add a feature where we need to send requests and responses back and forth then we need to either change this RPC or add a new one. Looks like a more long term approach here is to have a bidirectional stream in await_task. But this will require a change of this RCP semantic. It's not an await anymore but more like task_stream() which Executor is required to connect to to ensure that a task it just started using start_task can complete cause Executor needs to do some stuff for this like run get/set request context.

eabatalov: An alternative here is to send replies back from Executor to FE over a regular RPC and then no changes needed to the `await_task` RPC. But this looks like a workaround to me if we send a request from FE using a different approach.

src/tensorlake/function_executor/proto/function_executor.proto
earhart: Rebuilt. (The goal was to evolve the code to where we wanted it to be instead of doing everything all at once, but I'm fine doing everything all at once :-)

src/tensorlake/function_executor/proto/function_executor.proto
earhart: No particular reason; we needed something. Float gets us 23 bits of precision, plenty for most integer purposes.

src/tensorlake/function_executor/proto/function_executor.proto
earhart: IMHO, we _can_ reimplement GRPC on top of GRPC, but it's complicated and fragile - GRPC provides a lot of useful functionality for us. So instead of using a bidirectional stream for communicating requests back to the executor, I'd much rather pass the FE a URI to contact for its outgoing calls, and run a separate service.

If we really, really want a bidirectional stream in order to offer a backlink on the same channel, we can add another RPC to explicitly handle that case, so that the logic that's calling the FE to run a task can stay simple.

src/tensorlake/function_executor/service.py
earhart: I didn't want to get into a situation where we have an un-deletable task; if the Executor says "Delete this", I think it's the job of a correct FE to make the task go away, and it's not the executor's job to track that the delete failed and that it needs to try again sometime.

src/tensorlake/function_executor/service.py
earhart: We might not be able to immediately stop the task code, but we can complete await_task() calls and make the task appear to go away from the PoV of the executor.

src/tensorlake/function_executor/service.py
earhart: We want to return the intermediate progress updates.

src/tensorlake/function_executor/service.py
earhart: We can treat the progress update as a potential cancellation point, throwing an exception.

src/tensorlake/function_executor/service.py
earhart: I documented the reason why :-)
Yeah, don't remove the sleep here! I don't know that it's guaranteed that notifying the condition variable will release the GIL, and the sleep is harmless.

src/tensorlake/function_executor/service.py
earhart: Outcome == unknown is the zero state of the field; the executor of course needs to deal with it, but we should report a final outcome.

src/tensorlake/function_executor/service.py
earhart: Why?

I've actually been meaning to ask about this - in other Python projects I've never done this passing around loggers thing; it's always been "Create a file-level logger and use it everywhere" -- i.e. what the Python logging package says is idiomatic logger usage.  But maybe the idiomatic approach isn't right for our codebase...? :-). So, what's the rationale?

src/tensorlake/function_executor/service.py
earhart: It's not set on all code paths, but it's not obvious, which is why I like to ensure sensible values for all variables that are accessed in the exception path before the exception path could be invoked.

src/tensorlake/function_executor/service.py
earhart: Done

src/tensorlake/function_executor/service.py
earhart: Done (good catch, dunno what I was thinking there :-)

src/tensorlake/function_executor/service.py
earhart: Done

src/tensorlake/documentai/_datasets.py
```diff
@@ -198,9 +198,9 @@ def parse_dataset_file(
         body = _create_dataset_parse_req(file, page_range, labels, mime_type)
 
         did = dataset.dataset_id if isinstance(dataset, Dataset) else dataset
-        parse_id = self._request(
-            "POST", f"/datasets/{did}/parse", json=body
-        ).json()["parse_id"]
+        parse_id = self._request("POST", f"/datasets/{did}/parse", json=body).json()[
```
diptanu: Remove these changes from the PR please

src/tensorlake/function_executor/proto/function_executor.proto
```diff
@@ -138,6 +159,54 @@ message RunTaskResponse {
     optional SerializedObject invocation_error_output = 10;
 }
 
+// TaskRequest contains the input data for task execution
+message TaskRequest {
```
diptanu: Can we rename this to FunctionInputs?

src/tensorlake/function_executor/proto/function_executor.proto
```diff
@@ -138,6 +159,54 @@ message RunTaskResponse {
...
+    optional SerializedObject function_input = 1;
+    optional SerializedObject function_init_value = 2;
+}
+
+// TaskResult contains the execution outcome and outputs of a task
+message TaskResult {
+    repeated SerializedObject function_outputs = 1;
+    // Edges that the function wants the invocation to be routed to.
+    repeated string next_functions = 2;
+    optional string stdout = 3;
```
diptanu, line 173: Aren't these getting moved into pre-signed urls? May be use SerializedObjects for these?

src/tensorlake/function_executor/proto/function_executor.proto
```diff
@@ -138,6 +159,54 @@ message RunTaskResponse {
...
+    optional SerializedObject invocation_error_output = 9;
+}
+
+// Task represents a task with its metadata and execution result
+message Task {
+    optional string task_id = 1;
+    optional string namespace = 2;
+    optional string graph_name = 3;
+    optional string graph_version = 4;
+    optional string function_name = 5;
```
diptanu, line 189: Why is the function name going to the FE? If these are used for reconciling state of the FE if the executor loses connection to the FE or if it's restarted, isn't task id enough?

src/tensorlake/function_executor/proto/function_executor.proto
```diff
@@ -173,7 +242,18 @@ service FunctionExecutor {
     rpc initialize_invocation_state_server(stream InvocationStateResponse) returns (stream InvocationStateRequest);
     // Executes the task defined in the request.
     // Multiple tasks can be running in parallel.
+    // This is logically equivalent to start_task / await_task / delete_task.
     rpc run_task(RunTaskRequest) returns (RunTaskResponse);
+    // Lists the currently-running tasks.
+    rpc list_tasks(ListTasksRequest) returns (ListTasksResponse);
+    // Creates and starts the task specified in the request.
+    rpc create_task(CreateTaskRequest) returns (Task);
```
diptanu, line 250: Is create_task going to replace run_task? Can you share sequence of calls between executor and FE under the following scenarios - 
1. Executor has to start a new task 
2. Executor has to start a new task while there is already a task running 
3. Grpc connection between the executor and FE breaks but both are alive 
4. Executor is restarted and all state in the executor is lost. The server will at some stream the desired state, etc. 
5. FE is restarted while the executor is still running.

src/tensorlake/documentai/_datasets.py
earhart: Done.  :-). (Thanks for the nudge, I've been running into this for days :-)

src/tensorlake/function_executor/proto/function_executor.proto
earhart: Done

src/tensorlake/function_executor/proto/function_executor.proto
earhart: That's Eugene's change.  For the current change, we want exactly the same fields that we currently get from run_task() / RunTaskRequest.

src/tensorlake/function_executor/proto/function_executor.proto
earhart: The function name is here because it's used in RunTaskRequest, and the Task message is also used in the CreateTaskRequest - this message type isn't solely for reconnecting after restart.

src/tensorlake/function_executor/proto/function_executor.proto
earhart: Yes - `create_task()`/`await_task()`/`delete_task()` replaces `run_task()`.  There's an example of how to use this in https://github.com/tensorlakeai/indexify/pull/1642, but the basic idea is that we switch to using REST semantics with `Task` as our resource type - so:
* To start a task, the executor uses `create_task()`
* To find out what's going on with a task, the executor subscribes to it with `await_task()`; this is an idempotent call, does not insert the task if it doesn't already exist, and always returns at least one message (the final task output).
* When the executor's commited all the task's outputs to durable storage, it uses `delete_task()` to tell the FE that it's safe to drop its knowledge of the task.
* The executor is free to start multiple tasks in parallel or not; that's completely at the executor's discretion.
* If a restarted executor doesn't know the state of the world, it should wait until it has the state of the world; once it does, it can use `list_tasks()` to reconcile the FE state.
* If the executor gets an unexpected result from the FE (the FE restarted, the FE is evil, the FE is broken...), it can take whatever actions it deems appropriate - I expect it should mark the FE as broken, tear it down, and report task failure to the server, letting the server decide what to do.

diptanu: This makes sense. Thanks.

src/tensorlake/function_executor/proto/function_executor.proto
eabatalov: Floats are never "precise" afaik because they represent numbers using "steps". I undestand if we used float to represent fractions of something but we have current and total here so we can represent the full fraction with ints without loosing any precision due to floats. This is why I'd prefer ints or better uints here because we don't expect negative progress or total.

src/tensorlake/function_executor/service.py
eabatalov: Because the caller passes a logger bound with the caller context and then the callee adds any extra callee specific context to the logger so there's no duplication or assumptions by callee about what's in the context and the context is always complete for each code path.

We even started to do the same on the Server side but it takes a while to migrate to this approach.

src/tensorlake/function_executor/service.py
eabatalov: What's the reason of having code paths that might not return a value that represents results of this code path?
Such code is much harder for code reader to understand.

src/tensorlake/function_executor/service.py
eabatalov: Exactly because of this the log context in this line is incomplete, it doesn't include alloc id and invocation id

src/tensorlake/function_executor/service.py
```diff
@@ -41,13 +44,87 @@
...
+    if run_request.HasField("function_input"):
+        request.function_input.CopyFrom(run_request.function_input)
+
+    if run_request.HasField("function_init_value"):
+        request.function_init_value.CopyFrom(run_request.function_init_value)
+
+    return request
+
+
+def _run_task_request_to_task(run_request: RunTaskRequest) -> Task:
```
eabatalov, line 75: It seems to me like this data model would suite the domain better:

```
class Task
    task_execution: _TaskExecution
    input: FunctionInput
    result: Optional[TaskResult]
```

TaskExecution object here being the root object doesn't seem optimal as a task exists longer than it's being executed.

src/tensorlake/function_executor/service.py
eabatalov: I mean why do this 

```
if (
                    task_execution.task.result.outcome_code
                    != TaskOutcomeCode.TASK_OUTCOME_CODE_UNKNOWN
                ):
```

instead of checking for an optional result?

eabatalov: I again think that it's wrong to check TaskOutcomeCode in FE. TaskOutcomeCode is part of result,  it's not a task state that tracks its full lifecycle.

src/tensorlake/function_executor/service.py
eabatalov: There're undeletable tasks by design. E.g. this task_execution.updated can block forever because the user function can block forever. Given that I don't understand why all this extra logic and task await is needed here. Also this design forces Executor to use function deadline for this delete_task RPC and extend it if there are progress updates from the function.

src/tensorlake/function_executor/service.py
eabatalov: The task appears to go away from Executor PoV after delete_task RPC finished successfully. If delete_task blocks forever or Executor client deadline exceeded then it didn't delete successfully.

src/tensorlake/function_executor/service.py
```diff
@@ -256,3 +364,172 @@ def get_info(
...
+            with task_execution.updated:
+                if (
+                    task_execution.task.result.outcome_code
+                    != TaskOutcomeCode.TASK_OUTCOME_CODE_UNKNOWN
+                ):
+                    # Use TaskResult directly
+                    final_result = task_execution.task.result
+                    break
+
+                task_execution.updated.wait()
```
eabatalov: Would this result in Executor not getting a progress update if it was added while Executor was reconnecting to await_task?

src/tensorlake/function_executor/proto/function_executor.proto
eabatalov: Please confirm if libs like https://github.com/tqdm/tqdm use floats, then it's clear why

src/tensorlake/function_executor/service.py
```diff
@@ -41,13 +44,87 @@
...
 from .proto.function_executor_pb2_grpc import FunctionExecutorServicer
 from .std_outputs_capture import flush_logs, read_till_the_end
 
 
+def _run_task_request_to_function_inputs(run_request: RunTaskRequest) -> FunctionInputs:
+    """Convert RunTaskRequest to FunctionInputs."""
+    request = FunctionInputs()
+
+    if run_request.HasField("function_input"):
+        request.function_input.CopyFrom(run_request.function_input)
```
eabatalov, line 67: This is a required field, there's helper MessageValidator class that can do this validation.

src/tensorlake/function_executor/service.py
```diff
@@ -41,13 +44,87 @@
...
+    return response
+
+
+class _TaskExecution:
+    def __init__(self, task: Task):
+        self.task = task
+        self.thread: threading.Thread | None = None
+        # Initialize with empty result using TaskResult
+        self.task.result.CopyFrom(
+            TaskResult(
```
eabatalov: Why not just None? Even from code lines perspective this is shorted and it's less code to maintain when something changes in TaskResult.

src/tensorlake/function_executor/service.py
```diff
@@ -201,46 +281,74 @@ def initialize_invocation_state_server(
...
-    ) -> RunTaskResponse:
+    def _validate_task(self, task: Task):
         # Customer function code never raises an exception because we catch all of them and add
         # their details to the response. We can only get an exception here if our own code failed.
         # If our code raises an exception the grpc framework converts it into GRPC_STATUS_UNKNOWN
         # error with the exception message. Differentiating errors is not needed for now.
-        RunTaskRequestValidator(request=request).check()
+
+        # Validate required fields
+        if not task.namespace:
```
eabatalov: Why not using existing MessageValidator class?

src/tensorlake/function_executor/handlers/run_function/function_inputs_loader.py
```diff
@@ -22,8 +22,8 @@ def __init__(
 
 
 class FunctionInputsLoader:
-    def __init__(self, request: RunTaskRequest):
-        self._request = request
+    def __init__(self, task: Task):
+        self._task = task
```
eabatalov, line 26: Nit. Passing _task.request here is sufficient. func inputs loader doesn't need to know about anything else besides the function inputs.

src/tensorlake/function_executor/service.py
eabatalov:
```
class TaskInfo
    task: Task
    input: FunctionInput
    result: Optional[TaskResult]
    # Task execution state
    # ....
```

src/tensorlake/function_executor/service.py
```diff
@@ -256,3 +364,177 @@ def get_info(
...
+                task_execution.updated.notify_all()
+            # sleep(0) here momentarily releases the GIL, giving
+            # other threads a chance to run.
+            time.sleep(0)
+
+            # TODO: Consider implementing task cancellation here.
+
+        try:
+            # Run the task handler
+            result = RunTaskHandler(
```
eabatalov, line 376: task_info.result = RunTaskHandler().run()

src/tensorlake/function_executor/service.py
```diff
@@ -256,3 +364,177 @@ def get_info(
...
+            ).run()
+
+            if result.outcome_code == TaskOutcomeCode.TASK_OUTCOME_CODE_UNKNOWN:
+                result.outcome_code = TaskOutcomeCode.TASK_OUTCOME_CODE_FAILURE
+                result.failure_reason = (
+                    TaskFailureReason.TASK_FAILURE_REASON_INTERNAL_ERROR
+                )
+
+        except Exception as e:
+            # Handle any errors by creating a failed task result
```
eabatalov, line 392: "Logging the error, we already created internal error result which we set in finally block"

src/tensorlake/function_executor/service.py
```diff
@@ -256,3 +364,177 @@ def get_info(
...
+                function_wrapper=self._function_wrapper,
+                function_stdout=self._function_stdout,
+                function_stderr=self._function_stderr,
+                graph_metadata=self._graph_metadata,
+                progress_reporter=progress_reporter,
+                logger=self._logger,
+            ).run()
+
+            if result.outcome_code == TaskOutcomeCode.TASK_OUTCOME_CODE_UNKNOWN:
+                result.outcome_code = TaskOutcomeCode.TASK_OUTCOME_CODE_FAILURE
```
eabatalov: Please add an error log line explaining that there's an unexpected service state/bug here.

src/tensorlake/function_executor/proto/function_executor.proto
earhart: Confirmed; TQDM takes either, progressbar2 only takes floats (but int should just cast over).

src/tensorlake/function_executor/service.py
earhart: Done

src/tensorlake/function_executor/service.py
earhart: Done

src/tensorlake/function_executor/service.py
earhart: Done

src/tensorlake/function_executor/service.py
earhart: Done - split into TaskInfo and TaskExecution

src/tensorlake/function_executor/service.py
earhart: Makes sense - done

src/tensorlake/function_executor/service.py
earhart: Done

src/tensorlake/function_executor/service.py
earhart: Done - we'll now always immediately report _something_.

src/tensorlake/function_executor/service.py
earhart: This code will be removed pretty soon, and this structure lets us do validation in one place.

src/tensorlake/function_executor/service.py
earhart: Style.  :-). Done.

src/tensorlake/function_executor/service.py
earhart: Done

src/tensorlake/function_executor/handlers/run_function/function_inputs_loader.py
earhart: It's also getting a few other fields from the Task.

src/tensorlake/function_executor/service.py
earhart: Done (well, with a CopyFrom).

src/tensorlake/function_executor/service.py
earhart: Done (via simplifying the code here)

src/tensorlake/function_executor/service.py
earhart: I just removed it - we'll count on the invariants.

diptanu (approved):

src/tensorlake/function_executor/service.py
```diff
@@ -41,13 +44,87 @@
...
+        self.thread: threading.Thread | None = None
+        self.updated = threading.Condition()
+        self.progress = ProgressUpdate(current=0.0, total=1.0)
+
+
+class _TaskInfo:
+    def __init__(self, task: Task):
+        self.task = task
+        self.execution = _TaskExecution()
+        self.task.result.Clear()
```
eabatalov, line 125: Nit. Why do we clear result here?

src/tensorlake/function_executor/service.py
```diff
@@ -201,46 +281,67 @@ def initialize_invocation_state_server(
...
+
+        try:
+            for response in self.await_task(
+                AwaitTaskRequest(task_id=request.task_id), context
+            ):
+                last_response = response
+        finally:
+            self.delete_task(DeleteTaskRequest(task_id=request.task_id), context)
+
+        assert last_response.WhichOneof("response") == "task_result"
```
eabatalov, line 340: We'll remove this code soon but we never use Python asserts in production code. It's not necessary to use them and no one knows about them and how they work. We use error logs for unexpected situations.

src/tensorlake/function_executor/service.py
```diff
@@ -256,3 +357,159 @@ def get_info(
...
+            ).run()
+
+            task_info.task.result.CopyFrom(result)
+
+        except BaseException as e:
+            # Handle any errors by creating a failed task result
+            self._logger.error(
+                "task execution failed in background thread",
+                task_id=task_info.task.task_id,
+                exc_info=e,
```
eabatalov, line 396: log tags are missing (invocation_id, allocation_id), see the run task handler logger. task_info.logger would do the job and it can be passed around with task_info to other code paths. We can gradually migrate them to use it.

src/tensorlake/function_executor/proto/function_executor.proto
```diff
@@ -138,6 +159,54 @@ message RunTaskResponse {
...
+// Task represents a task with its metadata and execution result
+message Task {
+    optional string task_id = 1;
+    optional string namespace = 2;
+    optional string graph_name = 3;
+    optional string graph_version = 4;
+    optional string function_name = 5;
+    optional string graph_invocation_id = 6;
+    optional string allocation_id = 7;
+    // Input data for task execution (not returned when getting/listing tasks)
```
eabatalov, line 192: Should request and result be here if they are always None? If it's indeed always None then let's delete them to make protocol and code more clear.

src/tensorlake/function_executor/service.py
```diff
@@ -256,3 +357,159 @@ def get_info(
...
+            )
+
+        # Stream progress updates until task completes
+        final_result = None
+        sent_first_progress = False
+        while True:
+            with task_info.execution.updated:
+                if task_info.execution.complete:
+                    # Use TaskResult directly
+                    final_result = task_info.task.result
```
eabatalov, line 476: If task_info.task.result is None then Executor will treat this as internal error right?

src/tensorlake/function_executor/service.py
earhart: In case it was passed in by the caller. (Shouldn't happen, but it's nice to be clear about it.)

src/tensorlake/function_executor/service.py
earhart: Noted - but yeah, we're removing this code anyway, this is just a quick way to throw an exception if the state is weird.

src/tensorlake/function_executor/proto/function_executor.proto
earhart: They're not always None - request is set on create, response is set when the task completes.
An alternative might be to treat the task execution as the resource being represented.

src/tensorlake/function_executor/service.py
earhart: It had better - it can't depend on the function executor to behave correctly.

src/tensorlake/function_executor/service.py
earhart: SGTM

src/tensorlake/function_executor/proto/function_executor.proto
eabatalov: There's no connectivity from FE to host. Only from host to FE. This will require extra container networking setup work including security and FE authorization. But we also have alternatives to this so not a big deal okay.

src/tensorlake/function_executor/service.py
eabatalov: Our classic discussion about invariants :)

src/tensorlake/function_executor/proto/function_executor.proto
eabatalov: I mean it seem like we never return a Task from FE RPCs with non None request and result. Then why are they here? It seems like something again about preferences. But I mean... It's dead code isn't it and we have to clear these fields when returning Task object. This doesn't look clean to me.

src/tensorlake/function_executor/service.py
eabatalov: I guess you mean yes, lgtm in this case.

src/tensorlake/function_executor/proto/function_executor.proto
eabatalov: I understand we'd need to change quite a bit of code  on FE side to address this and imo it's not worth it at this point. But I'd like to understand if I'm missing something or not here.

PR #298, opened 2025-07-30
feat: print status update for better deployment UX

miguelhrocha:
### Description

Improve the SDK experience by showing a different message depending on the build status


diptanu (approved):
Can you add a screenshot for PRs like this? Hard to tell how the cli looks from just reading the code

PR #301, opened 2025-07-31
feat: cancel build on keyboard interrupt

miguelhrocha:
### Description

Catch when a user interrupts the deploy command and cancel the build on best-effort

### DIFF
--- a/src/tensorlake/builder/client_v2.py
+++ b/src/tensorlake/builder/client_v2.py
@@ -20,6 +20,7 @@ from dataclasses import dataclass
 from typing import Dict, Optional
 
 import aiofiles
+import asyncio
 import click
 import httpx
 from httpx_sse import aconnect_sse
@@ -196,7 +197,16 @@ class ImageBuilderV2Client:
         build = BuildInfo.model_validate(res.json())
 
         click.secho(f"Build ID: {build.id}", fg="green")
-        return await self.stream_logs(build)
+        try:
+            return await self.stream_logs(build)
+        except (asyncio.CancelledError, KeyboardInterrupt, click.Abort):
+            try:
+                click.secho(f"Cancelling remote build {build.id}...", fg="yellow")
+                await self._cancel_build(build)
+                click.secho(f"Cancelled build {build.id}", fg="yellow")
+            except Exception as e:
+                click.secho(f"Failed to cancel build {build.id}: {e}", fg="red")
+            raise
 
     async def stream_logs(self, build: BuildInfo) -> BuildInfo:
         """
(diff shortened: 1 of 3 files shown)


src/tensorlake/builder/client_v2.py
```diff
@@ -195,8 +196,25 @@ async def build(self, context: BuildContext, image: Image) -> BuildInfo:
         res.raise_for_status()
         build = BuildInfo.model_validate(res.json())
 
+        click.secho(f"Starting build for image {image.image_name} ...", fg="green")
         click.secho(f"Build ID: {build.id}", fg="green")
-        return await self.stream_logs(build)
+        try:
+            return await self.stream_logs(build)
+        except (asyncio.CancelledError, KeyboardInterrupt, click.Abort):
+            try:
```
eabatalov: Would be easier to read if this whole try block under except(KeyboardInterrupt) was moved into self._cancel_build().

I mean try inside except is not something we see often.

src/tensorlake/builder/client_v2.py
```diff
@@ -195,8 +196,25 @@ async def build(self, context: BuildContext, image: Image) -> BuildInfo:
...
+                await self._cancel_build(build, image)
+                click.secho(
+                    f"Cancelled build for image {image.image_name}", fg="yellow"
+                )
+            except Exception as e:
+                click.secho(
+                    f"Failed to cancel build for image {image.image_name}: {e}",
+                    fg="red",
+                )
+            raise
```
eabatalov, line 217: Why do we need to re-raise? Would this result in a clean error message to customer?

src/tensorlake/builder/client_v2.py
```diff
@@ -195,8 +196,25 @@ async def build(self, context: BuildContext, image: Image) -> BuildInfo:
         res.raise_for_status()
         build = BuildInfo.model_validate(res.json())
 
+        click.secho(f"Starting build for image {image.image_name} ...", fg="green")
         click.secho(f"Build ID: {build.id}", fg="green")
-        return await self.stream_logs(build)
+        try:
+            return await self.stream_logs(build)
+        except (asyncio.CancelledError, KeyboardInterrupt, click.Abort):
```
eabatalov, line 215: Could you please explain in a comment why all these exception types are handled here? Are all of them for Ctrl+C or something else? If only ctrl+C then why it's not just one exception?

Also we're using this in Executor for handling ctrl+C fyi:

```
# get the event loop
loop = asyncio.get_running_loop()
# add a custom signal handler
loop.add_signal_handler(signal.SIGINT, signal_handler)
```

You most probably tested that this PR works but I'm a bit surprised that there's multiple ways to handle that signal in async Python.

src/tensorlake/builder/client_v2.py
```diff
@@ -207,29 +225,33 @@ async def stream_logs(self, build: BuildInfo) -> BuildInfo:
...
+                                fg="black",
+                                dim=True,
+                            )
+                        elif log_entry.stream == "stderr":
+                            click.secho(log_entry.message, fg="red", err=True)
+                        elif log_entry.stream == "info":
+                            click.secho(
+                                f"{log_entry.timestamp}: {log_entry.message}", fg="blue"
+                            )
+        except asyncio.CancelledError:
```
eabatalov: If we just re-raise the exception then this except is not needed.

src/tensorlake/builder/client_v2.py
miguelhrocha: Yeah, having these exceptions is what provides a better UX. I didn't know about the `signal_handler`, but that doesn't abort the click command as opposed to the signal loop handler

src/tensorlake/builder/client_v2.py
miguelhrocha: So it stops the loop. Otherwise, it just stops the build that is currently processing, raising it halts the whole deploy process

PR #302, opened 2025-08-01
Add label filter constraints

earhart:


eabatalov (changes requested):
How customers are going to use this feature? How are they going to reason about what to put into these constraints?

E.g. what is "Executor" in tensorlake SDK?

earhart:
> How customers are going to use this feature? How are they going to reason about what to put into these constraints?
> 
> E.g. what is "Executor" in tensorlake SDK?

<nods> We'll definitely need to document the feature, and we'll need to document the tags we're supporting; I'm not sure at all what the terminology should be. But I think there's a general notion here of being able to run some code remotely "somewhere"; this is just a mechanism for describing aspects of "somewhere".

Do you have any specific suggestions for changes here?

eabatalov:
i.e. the only SDK customer visible label I can think about right now is region. But on standup we discussed that even this is not needed in public API any time soon.

src/tensorlake/functions_sdk/functions.py
```diff
@@ -134,6 +182,7 @@ class TensorlakeCompute:
     memory: float = _DEFAULT_MEMORY_GB
     ephemeral_disk: float = _DEFAULT_EPHEMERAL_DISK_GB
     gpu: Optional[Union[str, List[str]]] = _DEFAULT_GPU
+    placement_constraints: Optional[LabelsFilter] = None
```
diptanu: Should be region, looks like the docs say it's region as well.

src/tensorlake/functions_sdk/functions.py
```diff
@@ -203,10 +252,15 @@ def tensorlake_function(
     memory: float = _DEFAULT_MEMORY_GB,
     ephemeral_disk: float = _DEFAULT_EPHEMERAL_DISK_GB,
     gpu: Optional[Union[str, List[str]]] = _DEFAULT_GPU,
+    region: Optional[str] = None,
     next: Optional[Union["TensorlakeCompute", List["TensorlakeCompute"]]] = None,
     cacheable: bool = False,
 ):
     def construct(fn):
+        placement_constraints = None
```
diptanu: This should be region, and in `ComputeFunctionDefinition` you can map regions to LabelFilters

src/tensorlake/__init__.py
```diff
@@ -17,6 +18,7 @@
     "Graph",
     "GraphRequestContext",
     "Image",
+    "LabelsFilter",
```
diptanu: This doesn't need to be exported for the time being.

diptanu (approved):

src/tensorlake/functions_sdk/functions.py
```diff
@@ -115,6 +115,8 @@ class TensorlakeCompute:
         gpu (Optional[str]): GPU(s) available to the function. No GPU is allocated by default.
                              The value should be a string "GPU_MODEL:COUNT" representing the GPU model and the number of GPUs.
                              See supported GPU models and counts in Tensorlake Cloud documentation.
+        region (Optional[str]): The region where this function should be executed (e.g., "us-west", "eu-central").
+                               If specified, the function will only run on executors in that region.
```
diptanu: Change this comment to **run anywhere** globally

eabatalov (commented):
Overall LGTM, but let's be clear on supported
list of regions and make region required with
a default non None region.

PR #308, opened 2025-08-02
Fix block_until_done usage

earhart:


src/tensorlake/functions_sdk/http_client.py
```diff
@@ -384,6 +384,7 @@ def _call(
     def call_stream(
         self,
         graph: str,
+        block_until_done: bool = False,
```
diptanu: This is not needed anymore. We made stream always be blocking until the request ends. The HTTP Accept header changes the behavior of making a request. If it’s text/event-stream it blocks, if its application/json it returns a json object

diptanu: Since there is no description I am not sure what is the context of the PR. Was there  bug somewhere?

diptanu (changes requested):

src/tensorlake/functions_sdk/http_client.py
earhart: Yeah, the code was using `block_until_done`, but it wasn't defined.

PR #331, opened 2025-08-14
UX changes for Graph List, Dict, Tuple return values

eabatalov:
* Don't start map operation if a list is returned from function. Only start a map operation if user returns a Map(...) object explicitly.
* Don't convert a tuple returned from a function into positional arguments of the next function(s).
* Don't convert a dict returned from a function into kwargs of the next function(s).

Consequently:

* Each function has only one "input" parameter (in addition to accumulator and ctx).
* graph.run APIs accept a single argument `request` which is passed into the graph's start function "input" parameter. It doesn't matter how this parameter is called in the function. All parameter binding is positional now.
* graph.run `request` parameter has to be supplied, it doesn't have a default None value. This is because we'll have to map that default None value to something in graph HTTP endpoint too. I just decided to not provide this functionality to not support this corner case in UX.

I think with these changes we're more friendly towards calling graphs via http API with a clear request payload without any Python magic happening in between.


diptanu:
Regarding Map, I wonder if we implement the feature we talked about this morning - 

return FunctionName.call(value), would we want to represent Map something like this return FunctionName.map(values)

PR #332, opened 2025-08-15
RFC: New proposed UX for Workflows SDK

eabatalov:


new_ux.py
```diff
@@ -0,0 +1,186 @@
...
+# @tensorlake.graph_api(...)
+class TestGraphFunction3:
+    # The class must have an empty constructor.
+    # This is because when this class is called by other functions using `TestGraphFunction3().run(...)` we don't support
+    # different constructor arguments for different requests. In the future we can support it by e.g. hashing constructor
+    # args and kwargs and adding this to the function name but this is too complex to support right now.
+
+    # The decorator makes actual class constructor body empty when ppl call TestGraphFunction3(). When however we're running
+    # a graph the original __init__ is actually called.
+    def __init__(self):
```
diptanu, line 81: During implementation we should validate if these constructors take more than one argument.

new_ux.py
```diff
@@ -0,0 +1,186 @@
...
+
+        time.sleep(100)
+        self.magic_number: int = 2
+
+    # This is not a @staticmethod because a static method doesn't have self. And I assume that ppl are much less used to using static methods.
+    # So we want to give them a friendly instance method here.
+    #
+    # Not using __call__(self) here because I assume that people are not generally familiar with operator overloading in Python and also
+    # because it'd result in an odd syntax at call sites: `TestGraphFunction3()(ctx, d, ix)`.
+    def run(self, ctx: tensorlake.RequestContext, d: dict, ix: int):
```
diptanu, line 95: You think we annotate functions in classes with @tensorlake_function? Feels more natural to annotate functions than classes with `tensorlake_function`

eabatalov: Let's go with the same set of decorators for functions and classes initially. It's just less to support in SDK interface and also has some pros like ppl needing to remember less decorator names. It looks counter intuitive in a way that a class is decorated as a function but from graph perspective they are all the same. Class is just a function with an init hook. So we're not lying much here from the domain perspective. So let's say if we have `class InferenceModel` then it's still a function in the graph :)

eabatalov: If we get user or llm feedback that this is confusing then we can add a separate set of decorators for classes, no problem about it.

eabatalov: Actually I got a better idea... Let's just apply the decorator to the def run() method and allow any name for it. This will feel even more natural for ppl and also it'll remove the confusion between classes and functions.

new_ux.py
```diff
@@ -0,0 +1,266 @@
...
+# Server will need to be aware about function calls but not about a lot of things. Just function name. The rest (kwargs, args, etc) is
+# SDK side concern and Server has no visibility into that data.
+
+
+# input_encoder will be "json" by default. When deserializing an input from a json string we'll try a heuristic approach so e.g.
+# a "{'foo': 'bar'}" string is not deserialized into a regular Python dict() object but it's deserialized into data type used in the
+# function parameter type annotation if it has a constructor from dict() that we deserialized from json.
+#
+# API function must have a single user supplied argument caller `request`, this is HTTP calling convention. If the first argument
+# is called `ctx` then RequestContext is injected. User has to pass it to all functions that it calls if they have ctx parameter too,
```
calavera, line 36: Have you considered putting the `ctx` as the last argument instead of the first one? I've seen other frameworks that usually put that kind of optional parameter at the end. What I like about that approach is that the required arguments are always in the same position, and it's easier to create a mental model about positioning arguments. It might also be easier to implement since you don't have to take into account the context's position to parse the rest of the arguments.

new_ux.py
calavera: I'm guessing you're using `cls` because `class` is a keyword. Is there any other word that we could use without having to contract it? I always find hard to remember these kind of contractions. If we want to make the api more natural for people, maybe an easier to remember word is better.

new_ux.py
```diff
@@ -0,0 +1,266 @@
...
+
+# Overall goal:
+# Make the coding experience as close to native Python as possible. This would result in easier onboarding and
+# better DevEX. Also LLMs and IDE support will be much better.
+#
+# Overall idea:
+# User don't need to specify edges in the graph, all routing is dynamic now.
+# Use regular function calls using () operator to chain functions in the graph together.
+# A @tensorlake_function calls other Tensorlake functions by returning a list of function call closures (basically futures).
+# When a @tensorlake_function returns a non future value it means that this function actually produced output data.
```
calavera, line 13: Is this a common pattern in Python? I'm wondering if this can make it hard for people to create or debug graphs if the behavior depends on the semantics of returning values. 

I also noticed that somewhere there is a note about creating a JS SDK. Do we need to care about future languages and their similarities for this new design?

new_ux.py
```diff
@@ -0,0 +1,266 @@
...
+# There's also a PoC (see `get_called_tensorlake_computes`) on now to detect this automatically via AST traversal but I think
+# that making every @tensorlake_function in the code base callable in the graph actually makes sense as it's more natural Python
+# coding experience.
+
+# Some technical preliminary details of the implementation:
+# @tensorlake_function returns an instance of Callable that knows all the function attributes.
+# Using a Callable with () operator would result in better IDE and LLM support and much more natural coding experience.
+# The () call returns TensorlakeFunctionCall object that knows the function name and args, kwargs this is the future I mentioned above.
+#
+# Server will need to be aware about function calls but not about a lot of things. Just function name. The rest (kwargs, args, etc) is
```
diptanu, line 27: Are you thinking to encode the kwargs, args, etc, in the blob?

new_ux.py
eabatalov: It's not a common pattern in Python but it's the best we could come with so far. If you have any other ideas, then please share them.

The idea with any language we support is to provide an SDK that is closest to regular development in the language while allowing to build flexible Indexify computation graphs on the fly. There's no other known limitations per SDK.

new_ux.py
eabatalov: I don't think this is necessary because the strict function signature only applies to `@tensorlake.graph_api` functions. They have very limited signature both for `ctx` and `request` parameters. The rest of the functions get ctx at the point of function call and ctx can be in any place in function signature or can be ommited.

new_ux.py
eabatalov: Yes, because `class` is a keyword. I don't have any other ideas for the name, if you have some the please share. The only other alternative I have is `tensorlake.function_class` but this looks like mixing terms to me which causes more harm than improvement here imo.

new_ux.py
eabatalov: Important limitation of Indexify compute graph is that its functions has to finish when returning the next functions. This limitation is going to apply to all functions in all SDK languages.

Later we'd support `yield` Python operator which would feel more natural to ppl who code in Python. At the same time yield operator is a slightly more advanced thing and I'm not sure how many ppl are familiar with its use. So `yield` is not the main coding approach in Python SDK. Returning lists is much more basic language construct and this is why it's a primary usage scenario.

new_ux.py
calavera: some suggestions, I don't really know if any of them are any better 😅 
- container
- object
- structure
- struct
- definition
- bucket
- type
- wrapper

I noticed that Temporal uses `@workflow.defn`, so this might not matter 🤷

new_ux.py
calavera: That makes sense. Thanks for the explanation

src/tensorlake/functions_sdk/functions.py
```diff
@@ -243,6 +245,59 @@ def construct(fn):
     return construct
 
 
+def graph_api(graph_name: str, graph_description: str, version: str) -> Callable:
```
calavera: What do you think about dropping the `_api` suffix in the name of this decorator, and the `graph_` prefix in the parameters?

```
def graph(name: str, description: str, version: str) -> Callable:
```

new_ux.py
eabatalov: Thank you for the ideas. I quite like object but at the same time it seems a bit confusing to call something which is a class as an object right at the place where `class` is used. I'd stick to `cls` in this case. Consistent term but with some shortcomings in its naming :) 

Yep I saw other serverless Python products using shortened names too.

src/tensorlake/functions_sdk/functions.py
eabatalov: After some thinking about this I think it'd be right to split `tensorlake.Graph` class and `tensorlake.api_function` decorator. Cause API is an important term here, it puts restrictions on the function so it's callable via HTTP. At the same time a Graph is a separate thing from its entrypoint function. E.g. multiple Graphs can use the same API entrypoints. Lemme rework this a bit.

new_ux.py
eabatalov: Yep, it'll be transparent to Server, only SDK knows how these are encoded.

eabatalov:
Closing this PR because we know have close to working local implementation https://github.com/tensorlakeai/tensorlake/pull/343/files

It misses some thing like batching as we don't have this feature yet. I took a note of these in my Google Drive.

PR #343, opened 2025-08-21
Local mode implementation of new Workflows UX

eabatalov:
Run:

```
poetry run python ./tests/workflows/test_simple_graph.py
poetry run python ./tests/workflows/test_complex_graph.py
poetry run python ./tests/workflows/test_fake_pdf_parse_workflow.py
```

to execute the example graph locally.

This change doesn't affect existing SDK code as all the code here is completely independent from existing SDK code.

### DIFF
--- a/src/tensorlake/workflows/ast/ast_node.py
+++ b/src/tensorlake/workflows/ast/ast_node.py
@@ -64,3 +64,7 @@ class ASTNode:
             raise ValueError(
                 f"Old child with id {old_child.id} is not a child of parent node with id {self.id}"
             )
+
+    def add_child(self, child: "ASTNode") -> None:
+        self.children[child.id] = child
+        child.parent = self
--- a/src/tensorlake/workflows/ast/function_call_node.py
+++ b/src/tensorlake/workflows/ast/function_call_node.py
@@ -30,7 +30,6 @@ class RegularFunctionCallMetadata(BaseModel):
     kwargs: Dict[str, ArgumentMetadata]
 
     def serialize(self) -> bytes:
-        # Using pickle here because it's faster than JSON serialization according to Internet.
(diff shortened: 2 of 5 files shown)


src/tensorlake/workflows/interface/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+    RequestProgress,
+    RequestState,
+)
+from .run import local_run, remote_run
+
+__all__ = [
+    "api",
+    "cls",
+    "function",
+    "local_run",
```
calavera: What do you think about this idea that just crossed my mind:

What if instead of having `local_run` and `remote_run`, we have one only `run` method with a parameter `remote=False`. That might give people easier control flow, see this example:

```python
request = tensorlake.run(func(payload)) // run this locally
request = tensorlake.run(func(payload), remote=True) // run this remotely

import os
remote = True if os.getenv("TENSORLAKE_REMOTE") is not None
request = tensorlake.run(func(payload), remote=remote)
```

/cc @diptanu

diptanu: I like this suggestion because it will require no code change to go from local to remote. I believe the current API was designed to make it easy for people to make remote and local runs more explicit.

eabatalov: This is definitely an interesting angle but I'm on the fence about this tbh. I'll keep explicit API and add this new run method you suggested and see later what users like more.
I suspect that the use cases and preferences for both would be quite different depending on users.

eabatalov: Done.

calavera: I like that idea 👍

PR #359, opened 2025-09-29
Add subcommand to fetch application logs from the telemetry service.

calavera:
This change adds two new subcommands under application and request to be able to fetch application logs for them.

The application subcommand accepts additional filters besides the request, like the function and container where the function runs.

Formats:

I ended up going a little bit overboard with formatting, but I think it gives people flexibility. These are the, not 2, not 3, but 4!!!!! format options:

**Compact formatting**

This is the default format. It uses two lines to make it nicely to read in laptop screens as well as big screens.

image

**Expanded formatting**

This is like the compact format, but it shows the `logAttributes` that a user, or us, might have added to the log.

image

**Long formatting**

This prints everything in one line of long text.

image

**JSON formatting**

This prints each log entry as a pretty json object.

image

### DIFF
--- a/src/tensorlake/applications/remote/api_client.py
+++ b/src/tensorlake/applications/remote/api_client.py
@@ -146,7 +146,6 @@ class LogEntry(BaseModel):
 class LogsPayload(BaseModel):
     logs: list[LogEntry]
     next_token: str | None = Field(default=None, alias="nextToken")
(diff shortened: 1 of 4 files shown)


calavera:
Btw, the formats are also light theme compatible, for example, the default one is completely readable in white background.

image

diptanu (approved):

PR #378, opened 2025-10-09
Scope credentials by endpoint.

calavera:
This allow people to have credentials configured for different endpoints:

- Save credentials by endpoint. So you can have dev, prod, and on-prem configured.
- Save in a toml file, so both config files use the same formatting.
- Use a proper toml parser that create valid toml.
- Add endpoint to the `auth status` command.

### DIFF
--- a/src/tensorlake/cli/deploy.py
+++ b/src/tensorlake/cli/deploy.py
@@ -54,7 +54,7 @@ def deploy(
     except Exception as e:
         click.echo(
             f"Failed to load the application file, please check the error message: {e}",
-            err=True
+            err=True,
         )
         traceback.print_exception(e)
         raise click.Abort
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -25,6 +25,7 @@ tqdm = "^4.67.1"
 # Required only by Function Executor
 grpcio = "1.75.1"
 grpcio-tools = "1.75.1"
+tomlkit = "^0.13.3"
```
eabatalov, line 28: Adding each dependency to tensorlake needs to be carefully considered because it forces our users to use this dependency with this version and its dependencies too.

Also this is clearly not "# Required only by Function Executor" as line 25 states

src/tensorlake/cli/_configuration.py
```diff
@@ -0,0 +1,125 @@
...
+from tomlkit import document, dumps, parse, table
+
+CONFIG_DIR = Path.home() / ".config" / "tensorlake"
+
+CONFIG_FILE = CONFIG_DIR / ".tensorlake_config"
+
+CREDENTIALS_PATH = CONFIG_DIR / "credentials.toml"
+
+# Legacy credentials file (pre-endpoint-scoping)
+LEGACY_CREDENTIALS_PATH = CONFIG_DIR / "credentials.json"
```
eabatalov, line 15: I think we don't need to keep backward compatibility of this feature as it's not used anywhere really.

PR #380, opened 2025-10-10
Blocking, non-blocking and deferred function calls support

eabatalov:
Local runner and FE now supports them.

CI is failing because required Indexify release. All Tensorlake tests are passing locally.

### DIFF
--- a/src/tensorlake/applications/interface/awaitables.py
+++ b/src/tensorlake/applications/interface/awaitables.py
@@ -386,7 +386,7 @@ def make_reduce_operation_awaitable(
     function_name: str,
     items: List[Any | Awaitable] | AwaitableList,
     initial: Any | Awaitable | _InitialMissingType,
-) -> ReduceOperationAwaitable:
+) -> ReduceOperationAwaitable | Awaitable | Any:
     inputs: List[Any | Awaitable] = None
     if isinstance(items, AwaitableList):
         inputs = list(items.items)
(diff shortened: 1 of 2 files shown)


tests/applications/test_fake_pdf_parse_workflow.py
```diff
@@ -130,6 +130,15 @@ def run(self, chunk_embeddings: ChunkEmbeddings) -> None:
             )
 
 
+@function()
+def watch_pdf_updates(url: str, page_range: str) -> None:
```
calavera: what happens if this function fails? Do we stop the timer? Do we notify the user somehow? 

Do we recover from a failure starting with the last output this called produced, or from the first one that started the initial timer?

eabatalov: Yes, call_later is just a single function call, if it fails then it doesn't schedule then next delayed call.
Once we have this implemented we can think about more advanced scheduling of function calls where there's a schedule object that generates these function calls every X hours or on every Tuesday 8 AM PST.

tests/applications/test_fake_pdf_parse_workflow.py
```diff
@@ -61,27 +62,30 @@ class ResponsePayload(BaseModel):
...
+    # in a non-blocking way.
     response: FakePDFParseResult = fake_parse_pdf_service_call(
         file=payload.url,
         page_range=payload.page_range,
     )
-    # To tell SDK to deconstruct the futures list and track the FunctionCalls inside of it we have
-    # to wrap it into something we can recognize when building AST, i.e. tl_map.
-    chunk_embeddings = tl_map(
+    # Use map operation running in background as argument to other function calls.
+    chunk_embeddings: Future = tl_map.aio.call(
```
eabatalov: TODO: consider

Future f1 = tl_map.future(*args, **kwargs)
Future f2 = tl_map.later(100, *args, **kwargs)
Future f3 = tl_map.schedule(Schedule(...), *args, **kwargs)
f1.resolve()
f2.resolve()
f3.resolve()

tl_map()
tl_map.future_call()
tl_map.future_call_later()
tl_map.schedule(Schedule(...), args, kwargs)

Issue #385, opened 2025-10-13
CLI Behavior


diptanu:
The CLI needs to be refactored to be more ergonomic and logical. My proposal - 

When someone downloads tensorlake, they need to set the scope of their project/org in two ways, either using an API key, or using a PAT. 

API Keys can be set as `export TENSORLAKE_API_KEY=xxx` it connects the current shell to a specific project

`tensorlake login`
alternate to setting an api key. Connects the current environment to a org and allows the users to manipulate any project in the org. 

`tensorlake deploy <path_to_file>` 
Deploys the application on the file to the current project. If the user has not set an API Key, we should list all the projects and let the user select the project where they want to deploy the application. If there is only one project there is no need to ask the user to select a project, just logging the project where the app is getting deployed is enough.

When a user selects the project where they are deploying the application we should remember the selection in a .tensorlake/config.json folder in the current directory so that they don't have to select the same project again. Most likely they will keep deploying to the same project.

`tensorlake ls` 
lists all the applications in the current project. If the user is not using an API key, they have to login and select the project. Store the selection in a config file in the folder 

`tensorlake new`
creates a new template for an application 

`tensorlake logs <request id> --application <app name> --function <Function>` 
shows the logs for request id of an application. If the app name is not provided we ask for the app name by showing them a list of apps. We do the same for project and org for personal token. Once we have the information, we write it in .tensorlake/ folder in the current folder. If the function name is provided we show logs from the function otherwise we show all the logs of the functions>

`tensorlake parse`
continues to work for parsing documents from the cli. Its a little oddball to have it here but I think we can live with the compromise for the time being. 

`tensoralke config` - Suggest we remove this. We should create config files for the current environment in .tensorlake folder. Instead of it being a global folder it can be in the folder where the user is currently in, presumably the application folder. 

There are a lot of APIs/features not covered here, I think for application management, deployment and some troubleshooting this is enough. The UI can help users in other areas such as deleting apps, requests, seeing events, usage, etc

PR #388, opened 2025-10-15
Revamp CLI user experience

RaeesBhatti:
Fixes: https://github.com/tensorlakeai/tensorlake/issues/385

### DIFF
--- a/src/tensorlake/cli/auth.py
+++ b/src/tensorlake/cli/auth.py
@@ -96,20 +96,21 @@ def run_login_flow(ctx: Context, auto_init: bool = True) -> str:
 
     click.echo("We're going to open a web browser for you to enter a one-time code.")
     click.echo(f"Your code is: {user_code}")
+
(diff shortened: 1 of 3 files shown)


src/tensorlake/cli/_configuration.py
```diff
@@ -14,6 +14,9 @@
 # Legacy credentials file (pre-endpoint-scoping)
 LEGACY_CREDENTIALS_PATH = CONFIG_DIR / "credentials.json"
 
+# Local project configuration file
+LOCAL_CONFIG_FILE = Path.cwd() / ".tensorlake.toml"
```
calavera: should we make this into a directory, `.tensorkale/config.toml`? It will probably be easier to make any changes in the future if we need to that way.

If I look at this from past experiences, I like what GitHub did (all config files inside a directory from the start), more than what we did at Netlify (one file in the root, and then adding a directory later when we needed to dump more files in the repo). 🙈

eabatalov (approved):

PR #403, opened 2025-10-21
Separate success and failure event logs.

calavera:
This makes it more clear for customers that the initialization failed. The failure event also includes part of the stack trace that's relevant for users to see the cause of the failure.

### DIFF
--- a/tests/function_executor/test_run_allocation.py
+++ b/tests/function_executor/test_run_allocation.py
@@ -601,9 +601,7 @@ class TestRunAllocation(unittest.TestCase):
         # Check FE events in stdout
         self.assertIn("function_executor_initialization_started", fe_stdout)
         self.assertIn("function_executor_initialization_failed", fe_stdout)
-
-        # Check function output to stderr
-        self.assertIn("This function fails on initialization", process.read_stderr())
+        self.assertIn("This function fails on initialization", fe_stdout)
 
 
 if __name__ == "__main__":


src/tensorlake/function_executor/service.py
```diff
@@ -198,7 +199,7 @@ def initialize(
                 duration_sec=f"{time.monotonic() - start_time:.3f}",
                 # Don't log the exception to FE log as it contains customer data
             )
-            log_user_event_initialization_finished(event_details, success=False)
+            log_user_event_initialization_failed(event_details, error=e)
             # Print the exception to stderr so customer can see it there.
             traceback.print_exception(e)
```
eabatalov: We're already printing the full backtrace here, do we need to remove this line and the comment above it?

src/tensorlake/function_executor/user_events.py
```diff
@@ -51,6 +49,49 @@ def log_user_event_initialization_finished(
...
+            "namespace": details.namespace,
+            "application": details.application_name,
+            "application_version": details.application_version,
+            "function": details.function_name,
+            "error": clean_stack_trace(error),
+        }
+    )
+
+
+def clean_stack_trace(error: BaseException) -> list[str]:
```
diptanu: I am guessing it wasn't possible to surround the code which loads the user code with a try-catch?

eabatalov: Please:

* Make this function non-public by adding _ prefix.
* Make it clear in the name and comments that stack trace cleanup logic is only applicable to `importlib.import_module(function_zip_manifest.module_import_name)` line. If I understood this correctly.

What if we failed to initialize not on `importlib.import_module(function_zip_manifest.module_import_name)`? What's going to happen with this backtrace?

calavera: good call, I'll make it private and document the behavior. If we fail somewhere else, we won't return anything here because it's an unknown behavior.

calavera: we already have a try catch on that, but it's surrounded with other stack trace information from python's standard library. We cannot do a try/catch inside the stdlib.

src/tensorlake/function_executor/service.py
calavera: I'd like to see how the event data looks like in dev before removing it. It's easier for me to compare the behaviors when I have DataDog and our UI side by side. If there is nothing unexpected, I'll remove it later.

src/tensorlake/function_executor/user_events.py
```diff
@@ -51,6 +49,52 @@ def log_user_event_initialization_finished(
...
+    )
+
+
+def _clean_bootstrap_stack_trace(error: BaseException) -> list[str]:
+    """
+    Store only the lines of the stack relevant to the user's code.
+    Everything before the last line containing "importlib._bootstrap" is removed.
+
+    If the error was not generated by the `importlib._bootstrap` module,
+    we return an empty list because it's an undefined behavior.
```
eabatalov: Do we end up printing empty backtrace if initialization failed not due to user code module loading? Isn't it a wrong behavior?

eabatalov: Also what if this "frozen importlib._bootstrap" heuristic stops working?

src/tensorlake/function_executor/service.py
eabatalov: Let's actually leave a TODO for removing this to not forget.

src/tensorlake/function_executor/user_events.py
eabatalov: I think Diptanu meant exactly this line https://github.com/tensorlakeai/tensorlake/blob/main/src/tensorlake/function_executor/service.py#L173. It's definitely possible to surround it with try catch.

src/tensorlake/function_executor/service.py
```diff
@@ -437,6 +438,29 @@ def delete_allocation(
...
+            "function executor service initialization failed",
+            reason="failed to load customer function",
+            duration_sec=f"{time.monotonic() - start_time:.3f}",
+            # Don't log the exception to FE log as it contains customer data
+        )
+        log_user_event_initialization_failed(
+            event_details, error=error if display_error else None
+        )
+        # Print the exception to stderr so customer can see it there.
+        traceback.print_exception(error)
```
eabatalov: If we don't display the error then how is user or us are going to debug it once we remove this print_exception call?

src/tensorlake/function_executor/service.py
```diff
@@ -192,18 +203,8 @@ def initialize(
...
-                # Don't log the exception to FE log as it contains customer data
-            )
-            log_user_event_initialization_finished(event_details, success=False)
-            # Print the exception to stderr so customer can see it there.
-            traceback.print_exception(e)
-            return InitializeResponse(
-                outcome_code=InitializationOutcomeCode.INITIALIZATION_OUTCOME_CODE_FAILURE,
-                failure_reason=InitializationFailureReason.INITIALIZATION_FAILURE_REASON_FUNCTION_ERROR,
+            return self._handle_code_initialization_exception(
+                start_time, event_details, e, display_error=False
```
eabatalov: Whos' going to debug this error if we're not going to display it?

src/tensorlake/function_executor/service.py
```diff
@@ -165,13 +166,23 @@ def initialize(
                         f"Available functions: {list(code_zip_manifest.functions.keys())}"
                     )
                 )
+        except BaseException as e:
+            return self._handle_code_initialization_exception(
+                start_time, event_details, e, display_error=False
```
eabatalov: Same question here

src/tensorlake/function_executor/service.py
eabatalov: This try catch section calls customer code on `self._function_instance_arg = create_self_instance(
                    self._function.function_config.class_name
                )` line

src/tensorlake/function_executor/service.py
calavera: the errors that are not displaying should be considered service errors at this point. Customers should not have to care about those errors, these would be the exceptions we don't show customers:
- Failing to extract the zip manifest.
- Failing to create the function instance

eabatalov: This directly contradicts existing code (it doesn't log these "service errors" into FE logs) and this contradicts code comments like `# Process user controlled input in a try-except block to not treat errors here as our # internal platform errors.`.

Please update everything if you want to change this model.

However, please keep in mind that the manifest and etc are artefacts generated on user computers. If this is not a good enough justfication then we can change this.

Failing to create the function instance is definitely not correct to treat as service error because function class constructor runs customer code which i.e. can load GBs of data into GPU.

src/tensorlake/function_executor/service.py
calavera: in that case, we should report it. I've updated the implementation to report all exceptions in that original try/catch block

src/tensorlake/function_executor/service.py
calavera: good to know. I don't think I want to change the model. We'll report all errors as user errors.

src/tensorlake/function_executor/service.py
calavera: I've removed it so we don't have to worry about it anymore.

eabatalov (approved):

PR #404, opened 2025-10-22
feat: include pydantic model schema

ilbum:
## Description

- `_type_hint_json_schema()` to output pydantic model schema
- simplified function deployment output logic

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "tensorlake"
-version = "0.2.68"
+version = "0.2.69"
 description = "Tensorlake SDK for Document Ingestion API and Serverless Applications"
 readme = "README.md"
 authors = ["Tensorlake Inc. <support@tensorlake.ai>"]


src/tensorlake/applications/remote/manifests/function.py
```diff
@@ -103,10 +110,14 @@ def _type_hint_json_schema(type_hint) -> Dict[str, str]:
     if type_hint == Any:
         return {"type": "string", "description": "Any type"}
 
-    # Handle typing generics like List, Dict, etc. first
+    # Handle Pydantic BaseModel first
+    if inspect.isclass(type_hint) and issubclass(type_hint, BaseModel):
+        if hasattr(type_hint, "model_json_schema"):
+            return type_hint.model_json_schema()
```
eabatalov, line 118: That a nice method. I'm wondering if in case of non-model type hint we can create a Model class dynamically and just call its model_json_schema()["field_name"]. No need to implement this right now, just as an idea.

src/tensorlake/applications/remote/manifests/function.py
```diff
@@ -18,6 +13,18 @@
 )
 from .function_resources import resources_for_function
 
+TYPE_TO_JSON_SCHEMA = {
```
eabatalov: Is there any standard for json schema, i.e. an RFC that we can link here?

src/tensorlake/cli/deploy.py
```diff
@@ -138,63 +139,18 @@ def _deploy_applications(
...
-                type_name = getattr(param_annotation, "__name__", str(param_annotation))
-                param_type = f"<{type_name}>"
-        else:
-            param_type = "<value>"
-
-        click.echo(f"Deployed application: {fn_config.function_name}\n")
-        click.echo(
-            f"""To invoke the application, use the following curl command:
-curl -X POST {auth.base_url}/v1/namespaces/{auth.namespace}/applications/{fn_config.function_name} \\
+    for function in functions:
```
eabatalov: Use `for application in filter_applications(function)`. application type is Function.

src/tensorlake/cli/deploy.py
```diff
@@ -138,63 +139,18 @@ def _deploy_applications(
...
+        if function.application_config is not None:
```
eabatalov: Instead of using this internal SDK detail please use `for application in filter_applications(function)`. application type is Function.

eabatalov: There's `deployed_apps = filter_applications(functions)` deleted above, just use that.

src/tensorlake/applications/remote/manifests/function.py
```diff
@@ -259,3 +268,13 @@ def create_function_manifest(
         placement_constraints=placement_constraints,
         max_concurrency=function.function_config.max_concurrency,
     )
+
+
+def get_function_input_types(function: Function) -> str:
```
eabatalov: Instead of adding this function and using internal SDK things please reuse existing interface of this file in the CLI:

```
app_func_manifest: FunctionManifest = create_function_manifest(application_function)
func_name = app_func_manifest.name
# Do this check because we're going to support parameterless application functions soon.
if len(app_func_manifest.parameters) > 0:
    param_type  = app_func_manifest.parameters[0].data_type
```

src/tensorlake/cli/deploy.py
```diff
@@ -138,63 +139,18 @@ def _deploy_applications(
...
   -H "Content-Type: application/json" \\
-  -d '{param_type}'
+  -d '{get_function_input_types(function)}'
 """,
-        )
-    except StopIteration:
-        click.echo("Successfully deployed application")
-        click.echo("Error generating curl command\n", err=True)
+            )
+            break
```
eabatalov: There can be multiple apps in the user's code. We need to print a curl command for each application function.

src/tensorlake/cli/deploy.py
```diff
@@ -138,63 +139,18 @@ def _deploy_applications(
...
+                f"""To invoke the application, use the following curl command:
+curl -X POST {auth.base_url}/v1/namespaces/{auth.namespace}/applications/{func_name} \\
   -H "Authorization: Bearer $TENSORLAKE_API_KEY" \\
   -H "accept: application/json" \\
   -H "Content-Type: application/json" \\
-  -d '{param_type}'
+  -d '{get_function_input_types(function)}'
 """,
-        )
-    except StopIteration:
```
eabatalov, line 204: If this code path doesn't have any obvious bugs then it shouldn't fail, I agree that exception handling is not needed here for this reason. I'd leave a comment about this.

src/tensorlake/applications/remote/manifests/function.py
ilbum: Addressed

src/tensorlake/cli/deploy.py
ilbum: Sounds good, done

src/tensorlake/applications/remote/manifests/function.py
ilbum: Nice thank you, addressed this and left a TODO note

eabatalov (approved):
Really nice, thank you!

PR #409, opened 2025-10-24
Add new `Function.remote` and `Function.local` interfaces.

calavera:
Use these methods as entry points to the runners instead of `run_remote_application` and `run_local_application`. This is simiar to Modal's approach to expose these interfaces: https://modal.com/docs/reference/modal.Function#local

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -3,11 +3,9 @@ name: Tests
 on:
   push:
     branches:
-      - 'main'
+      - "main"
   pull_request:
     types: [opened, synchronize, reopened, labeled, unlabeled]
-    branches:
-      - 'main'
 
 permissions:
   id-token: write
@@ -27,7 +25,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v5
         with:
-          python-version: '3.10'
+          python-version: "3.10"
           cache: "pip"
 
       - name: Install Poetry


calavera:
This is creating a circular dependency between the `Function` and the `Runners` because the runners import the Function interface:

```
    File "/home/runner/work/tensorlake/tensorlake/src/tensorlake/applications/local/runner.py", line 13, in <module>
      from ..function.application_call import (
    File "/home/runner/work/tensorlake/tensorlake/src/tensorlake/applications/function/application_call.py", line 4, in <module>
      from ..interface.function import Function
  ImportError: cannot import name 'Function' from partially initialized module 'tensorlake.applications.interface.function' 
```

reference_app/reference_app.py
```diff
@@ -44,6 +41,6 @@ def reducer(total: Total, new: Total) -> Total:
 
 
 if __name__ == "__main__":
-    request: Request = run_remote_application(sequence_summer, 10)
+    request: Request = sequence_summer.remote(10)
```
eabatalov, line 44: All our tests need to get updated. But let's first decide on the new interface and if we keep the previous one.

src/tensorlake/applications/interface/function.py
```diff
@@ -118,6 +120,28 @@ def awaitable(self) -> "FunctionAwaitablesFactory":
...
+
+    # FIXME/TODO.
+    # Temporary fake remote implementation for tests to pass until remote runner gets reimplemented.
+    def remote(self, payload: Any) -> Request:
+        """Run the function remotely.
+        This function is executed on the Tensorlake environment.
+        """
+        return self.local(payload)
+
+    def run(self, payload: Any, remote: bool) -> Request:
```
eabatalov, line 138: We have function.awaitable(..).run() with a very different semantic already. As I only added the original run_application(..., remote:bool) functions only for easier testing I'd suggest to remove it in this PR and only leave local() and remote() - their names don't intersect with anything else.

src/tensorlake/applications/interface/function.py
```diff
@@ -118,6 +120,28 @@ def awaitable(self) -> "FunctionAwaitablesFactory":
         """Returns function factory for creating awaitables."""
         return self._awaitables_factory
 
+    def local(self, payload: Any) -> Request:
```
eabatalov, line 123: One thing we need to decide is if we return Request or actual output (app function return value) here.
It'd be much more consistent with the rest of Function.something methods to return the actual values.
At the same time in some cases users would want to know the request ID.

I'm think that maybe we need to make these new functions more consistent with how ppl run functions.
i.e.:

```
Function.local(...) -> returns the app function return value the same way as Function(...) called inside a Tensorlake Function code.
Function.local.request(...) -> returns the app request, similar to Function.awaitable.map(...) or Function.awaitable.reduce()
```

There are also other options on how to do this. Please share what you think.

tests/applications/test_complex_graph.py
```diff
@@ -94,8 +94,8 @@ def store_sum_as_file(total: int) -> File:
 class TestComplexGraph(unittest.TestCase):
     def test_local_api_call_of_complex_graph_produces_expected_outputs(self):
         for function in ["test_graph_api_reduce", "test_graph_api_fan_in"]:
-            request = run_local_application(
-                function,
+            func: Function = get_function(function)
+            request: Request = func.remote(
```
eabatalov, line 98: Oooops, this is actually a problem... When running a function remotely our SDK client should not require user to have the function (or application) definition awailable to them. They need to be able to call the application by name.

Also get_function() is not part of SDK interface - users are not going to use it and we shouldn't require them to use get_function() to call a function remotely.

We either need an alternative interface in SDK for calling app remotely by their names or abandon this idea... What do you think?

calavera: We could do something similar to what Modal does: `f = modal.Function.from_name("my-app", "function")`. See this example: https://modal.com/docs/reference/modal.Function#update_autoscaler

src/tensorlake/applications/interface/function.py
calavera: Instead of having subpaths, what do you think about using `invoke`, `spawn`, or `run` as the method name to start the request without waiting for the value? The api would be the following:

```
Function.local(...) -> returns the output of executing the function locally, blocking execution until the function completes.
Function.remote(...) -> returns the output of executing the function remotely, blocking execution until the function completes.
Function.spawn(...) -> calls the function remotely, but does not wait for it to complete, returning an awaitable object.
```

I'm not sure if having a local awaitable function call is very useful. If you think about what people will use local calls will be for testing most of the time.

eabatalov:
```
Function.spawn(...) -> calls the function remotely, but does not wait for it to complete, returning an awaitable object.
```

Did you mean "returning a Request object.". We can make it awaitable. The most important thing that it has Request ID, that it's not just some opaque awaitable.

tests/applications/test_complex_graph.py
eabatalov: This is interesting but in our case it's going to be `Application.from_name()` because we don't allow to call non-application functions. ppl only use application name to call it remotely.
But we don't have Application class either. Doesn't look much better than a function like `run_remote_application`.

src/tensorlake/applications/interface/function.py
calavera: yeah, something that returns a request object. That's probably the easiest path, we can make it awaitable later if it makes sense.

tests/applications/test_complex_graph.py
calavera: > But we don't have Application class either. Doesn't look much better than a function like run_remote_application.

Now that you mention it, what if we move all these methods to an `Application` class? That would probably make more sense, since as you say, we don't allow running non-application functions. That would also get rid of the circular dependency loop.

eabatalov: We don't have Application class because ppl don't need to create it.
They create Application by adding `@application()` decorator to their function.
There's no need to add it. Forcing ppl to create an Application class instances makes many things more complicated. i.e. they'll have to do it in the same module where there application function is defined, otherwise in certain cases things won't work.

PR #413, opened 2025-10-30
Reduce number of FEs required to run complex graph test by 2x

eabatalov:
Otherwise we run out of capacity to creeate FEs on test Executor in some environments like CI.

The CI is failing because Indexify is not released yet.

### DIFF
--- a/tests/applications/test_complex_graph.py
+++ b/tests/applications/test_complex_graph.py
@@ -96,24 +96,24 @@ class TestComplexGraph(unittest.TestCase):
         for function in ["test_graph_api_reduce", "test_graph_api_fan_in"]:
             request = run_local_application(
                 function,
-                TestGraphRequestPayload(numbers=[str(i) for i in range(10, 20)]),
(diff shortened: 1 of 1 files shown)


tests/applications/test_complex_graph.py
```diff
@@ -96,24 +96,24 @@ def test_local_api_call_of_complex_graph_produces_expected_outputs(self):
...
             request = run_local_application(
                 function,
-                TestGraphRequestPayload(numbers=[str(i) for i in range(10, 20)]),
+                TestGraphRequestPayload(numbers=[str(i) for i in range(10, 15)]),
             )
 
             file: File = request.output()
             self.assertEqual(file.content_type, "text/plain; charset=UTF-8")
-            self.assertEqual(file.content, b"Total sum: 280")
+            self.assertEqual(file.content, b"Total sum: 116")
```
miguelhrocha, line 104: Could you add a comment explaining why 116 is, and where this value comes from?

PR #428, opened 2025-11-05
Wrap function call user errors into an event.

calavera:
Instead of printing the raw backtrace into the log, print a structured JSON event. This will be more friendly for people to debug when they go see the remote logs.

I've also added a bunch of tests for the serializers to help people understand their behavior.


diptanu (approved):

eabatalov (approved):
Agreed, we should report user code failures using user evens module.

PR #430, opened 2025-11-05
Add helpers to our SDK to help users write nice structured logs.

calavera:
This will make logging a more integrated experience for our users, and will create nice JSON payloads that they can see in our UI later.

I'll follow up with docs on how to use these helpers.

### DIFF
--- a/src/tensorlake/cli/deploy.py
+++ b/src/tensorlake/cli/deploy.py
@@ -150,12 +150,17 @@ def _deploy_applications(
             # TODO: update after parameterless function support
             if len(app_func_manifest.parameters) > 0:
                 param_type = app_func_manifest.parameters[0].data_type
+                type = (
+                    param_type["type"]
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -26,6 +26,7 @@ tqdm = "^4.67.1"
 grpcio = "1.75.1"
 grpcio-tools = "1.75.1"
 tomlkit = "^0.13.3"
+structlog = "^25.5.0"
```
diptanu: Can we vendor this dependency?

calavera: I don't know how vendoring works in python. I think this will make the experience much better. To be honest, I was just going to put this configuration in the docs, but it took me a few tries to get it right, so I thought it'd be better to provide it directly for users.

calavera: done

PR #432, opened 2025-11-10
Revert "Add helpers to our SDK to help users write nice structured logs."

eabatalov:
Reverts tensorlakeai/tensorlake#430

Reasoning:

* No need to vendor structlog. We don't use structlog, ppl can install structlog as any other python dependency.
* We can provide example structlog configuration in our public docs if ppl request it.
* If we provide a better UX for a particular format of structured logging then we can update the public docs with the structlog configuration and users can use it if they want.
* We're not maintaining an "approved" logging library for people. This is not a responsibility of Tensorlake Apps SDK.


eabatalov:
We'll decide on this later.

PR #433, opened 2025-11-10
Resolve the cloud URL based on the API url.

calavera:
If the API url is set to https://api.tensorlake.dev, we want the cloud url to be automatically set to https://cloud.tensorlake.dev without having to specify that in the configuration.

This change checks the API url to resolve the right cloud url. That way, when you run `tensorlake login` with an instance that's not production, you can get redirected to the right cloud interface directly, otherwise you always end up in https://cloud.tensorlake.ai.

### DIFF
--- a/tests/cli/test_cloud_url.py
+++ b/tests/cli/test_cloud_url.py
@@ -203,15 +203,15 @@ class TestCloudURLIntegration(unittest.TestCase):
                 self.assertEqual(ctx.cloud_url, url)
 
     def test_cloud_url_different_from_base_url(self):
-        """Test that cloud_url and base_url are independent"""
+        """Test that cloud_url and api_url are independent"""
         ctx = Context.default(
             api_url="https://api.example.com",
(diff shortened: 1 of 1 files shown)


src/tensorlake/cli/_common.py
```diff
@@ -89,7 +89,7 @@ def client(self) -> httpx.Client:
                 )
 
             self._client = httpx.Client(
-                base_url=self.base_url, headers=headers, event_hooks=HTTP_EVENT_HOOKS
+                api_url=self.api_url, headers=headers, event_hooks=HTTP_EVENT_HOOKS
```
RaeesBhatti: `httpx.Client()` expects `base_url` as the parameter name, not `api_url`. This will cause a runtime error.

calavera: fixed

RaeesBhatti (approved):
Approved

PR #435, opened 2025-11-12
Added page_dimensions in output

cooleel:


src/tensorlake/documentai/models/_results.py
```diff
@@ -99,6 +99,7 @@ class Page(BaseModel):
     """
 
     dimensions: Optional[List[int]] = None
+    page_dimensions: Optional[Dict[str, int]] = None
```
miguelhrocha:
```suggestion
    page_dimensions: Optional[PageDimensions] = None
```

I'd propose to have a custom object instead of a generic Dict, so our customers can have better type hints.

cooleel: Updated @miguelhrocha

PR #437, opened 2025-11-12
Export progress updates

calavera:
The goal of these changes is to export allocation progress updates generated inside FE in order to expose them from an SSE api. The `update` function in `RequestCtx` should also allow users to add additional information like a message, or extra attributes that clients can consume later to present detailed information about the progress update.

TODO:
- [x] The `RequestCtx::update` function should allow additional data.
- [x] Progress updates should be added to the logs, so users can correlate logs that happen between progress updates with the updates.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "tensorlake"
-version = "0.2.80"
+version = "0.2.81"
 description = "Tensorlake SDK for Document Ingestion API and Serverless Applications"
 readme = "README.md"
 authors = ["Tensorlake Inc. <support@tensorlake.ai>"]


src/tensorlake/applications/interface/request_context.py
```diff
@@ -42,12 +42,16 @@ def counter(self, name: str, value: int = 1):
...
+    ) -> None:
         """Update the progress of the current Tensorlake Function call execution.
 
         Args:
-            current: Current request progress value
-            total: Total request progress value
+            current: Current function call execution step
+            total: Total number of function call execution steps
+            message: Optional message to display with the progress update
+            kwargs: Additional keyword arguments to pass to the progress update
```
eabatalov, line 64: How are the kwargs are going to be serialized and deserialized? Or they won't be?

eabatalov, line 64: How are the kwargs are going to be serialized and deserialized? Or they won't be?

src/tensorlake/function_executor/allocation_runner/allocation_runner.py
```diff
@@ -702,12 +703,30 @@ class ProxiedAllocationProgress(FunctionProgress):
...
         # sleep(0) here momentarily releases the GIL, giving other
         # FE threads a chance to run before returning back to customer code that
         # might never return GIL. i.e. allowing the FE to handle incoming RPCs,
         # report back allocation state updates, etc.
         # NB: this was never tested to fix anything in practice but nice to have
         # this just in case.
         time.sleep(0)
+
+
+def print_progress_update(
```
eabatalov: Please mark private using ynderscore

src/tensorlake/function_executor/allocation_runner/allocation_runner.py
```diff
@@ -702,12 +703,30 @@ class ProxiedAllocationProgress(FunctionProgress):
     def __init__(self, allocation_runner: AllocationRunner):
         self._allocation_runner: AllocationRunner = allocation_runner
 
-    def update(self, current: float, total: float) -> None:
+    def update(
+        self, current: float, total: float, message: str | None = None, **kwargs
+    ) -> None:
         self._allocation_runner._allocation_state.update_progress(current, total)
+        print_progress_update(current, total, message, **kwargs)
```
eabatalov: We'll be updating fe protocol to send kwargs and message to executor? Otherwise not sure how it'd get into server unless we want ti make log saver talk to server.

src/tensorlake/function_executor/allocation_runner/allocation_runner.py
calavera: done

src/tensorlake/function_executor/allocation_runner/allocation_runner.py
calavera: log saver takes care of putting the events in the right place.

eabatalov:
Please address the following issues in this PR:

* Our json serializer is not used so pydantic models are not supported.
* SerializationError exception is not documented in progress.update doc string.
* No tests for the new user visible functionality.
* Exceptions raised in json.dumps() are raised directly to user and these are not inheriting from TensorlakeError which the existing docstring suggests.

src/tensorlake/applications/interface/request_context.py
calavera: they are serialized as json in https://github.com/tensorlakeai/tensorlake/blob/main/src/tensorlake/function_executor/cloud_events.py#L9-L20

eabatalov: The linked code is wrong.
it needs to raise SDK SerializationError to user because we failed to serialize user object there. It's not an internal error.
Also please don't change FE logger, I have important fixes for it related to cloud events serialization there. But it's FE logger only.

PR #438, opened 2025-11-13
feat: add cmd, entrypoint and pip_install directives to Image DSL

miguelhrocha:
### Description

When OSS users need to set up a specific function for their image, they need to override the CMD or the ENTRYPOINT command.

The `pip_install` directive is just a convenience, as most of the RUN directives are used for pip install, and this will make the Image declaration less verbose.

### DIFF
--- a/src/tensorlake/applications/interface/image.py
+++ b/src/tensorlake/applications/interface/image.py
@@ -27,7 +27,7 @@ class _ImageBuildOperation:
 
 class Image:
     """
-    Represents a Docker image to be built for application functions.
+    Represents a Docker image to be used for application functions.
     """
 
     def __init__(
@@ -130,23 +130,11 @@ class Image:
             )
         )
 
-    def pip_install(self, packages: str | List[str]) -> "Image":
-        """
-        Install Python packages using pip in the image. Equivalent to running 'pip install' in a RUN command.
-
-        Args:
-            packages (str | List[str]): The package or list of packages to install.
-        """
-
-        pkgs = packages if isinstance(packages, list) else [packages]
-        install_cmd = f"pip install {' '.join(pkgs)}"
-        return self.run(install_cmd)
-
     def cmd(self, command: str | List[str]) -> "Image":
         """
         Set the default command for the image. Equivalent to the Dockerfile CMD command.
 
-        Setting the CMD in the image is not necessary when using TensorLake, as the application
+        Setting the CMD in the image is not necessary when using Tensorlake, as the application
         function will override it. However, it can be useful for local testing, or custom images.
 
         Args:
@@ -164,7 +152,7 @@ class Image:
         """
         Set the entrypoint for the image. Equivalent to the Dockerfile ENTRYPOINT command.
 
-        Setting the ENTRYPOINT in the image is not necessary when using TensorLake, as the application
+        Setting the ENTRYPOINT in the image is not necessary when using Tensorlake, as the application
         function will override it. However, it can be useful for local testing, or custom images
 
         Args:


src/tensorlake/applications/image.py
```diff
@@ -77,13 +77,26 @@ def dockerfile_content(img: Image) -> str:
...
 
+    cmd_op = None
+    entrypoint_op = None
     for op in img._build_operations:
+        if op.type == _ImageBuildOperationType.CMD:
+            cmd_op = op
+        elif op.type == _ImageBuildOperationType.ENTRYPOINT:
+            entrypoint_op = op
+
         dockerfile_lines.append(_render_build_op(op))
```
eabatalov, line 88: Why we don't use _render_build_op for the new commands?

eabatalov: I mean why don't call _render_build_op inside the loop and then dockerfile_lines.append()

src/tensorlake/applications/interface/image.py
```diff
@@ -24,6 +26,10 @@ class _ImageBuildOperation:
 
 
 class Image:
+    """
+    Represents a Docker image to be built for application functions.
```
eabatalov: "to be built" -> "to be used"

src/tensorlake/applications/interface/image.py
```diff
@@ -87,6 +130,54 @@ def run(
             )
         )
 
+    def pip_install(self, packages: str | List[str]) -> "Image":
```
eabatalov: I'm not sure if this is a good idea because other python package managers exist and which one is available depends on the base image. But this is not a big concern either. So no problem I think.

Worth adding support later for requirements.txt and other python dependency formats.

src/tensorlake/applications/interface/image.py
```diff
@@ -87,6 +130,54 @@ def run(
...
+
+        pkgs = packages if isinstance(packages, list) else [packages]
+        install_cmd = f"pip install {' '.join(pkgs)}"
+        return self.run(install_cmd)
+
+    def cmd(self, command: str | List[str]) -> "Image":
+        """
+        Set the default command for the image. Equivalent to the Dockerfile CMD command.
+
+        Setting the CMD in the image is not necessary when using TensorLake, as the application
```
eabatalov: Are we TensorLake or Tensorlake? :)

src/tensorlake/applications/interface/image.py
```diff
@@ -87,6 +130,54 @@ def run(
...
+
+        Args:
+            packages (str | List[str]): The package or list of packages to install.
+        """
+
+        pkgs = packages if isinstance(packages, list) else [packages]
+        install_cmd = f"pip install {' '.join(pkgs)}"
+        return self.run(install_cmd)
+
+    def cmd(self, command: str | List[str]) -> "Image":
```
eabatalov, line 133: Not great that we're adding OSS only functionality to SDK. Can we move this to Indexify as an extended image functionality?

src/tensorlake/applications/image.py
miguelhrocha: The reason is that I wanted to keep the CMD and ENTRYPOINT as the last lines of the Dockerfile for caching purposes. Without this, there won't be guarantees that they are the last ones.

src/tensorlake/applications/interface/image.py
miguelhrocha: The main issue is that we want to avoid diverging Document AI images between cloud and prod. If I understand what you are proposing, this would end up in having different image definitions for both of these, right?

src/tensorlake/applications/interface/image.py
miguelhrocha: I've removed it to reduce the change radius of this PR. We can add custom package manager directives in a later PR

src/tensorlake/applications/image.py
eabatalov: Can you explain this? Why the original order of instructions is not going to be cached? Can you add a comment explaining it in the source?

src/tensorlake/applications/interface/image.py
eabatalov: Cloud and on-prem. Got it.

I think what we're doing here is very questionable. If users use entrypoint or cmd then their apps just won't work. This is not the right design for SDK.

PR #439, opened 2025-11-14
Validate application on deployment and local run

eabatalov:
All the validations currently implemented will fail application execution if not addressed. So they are necessary. This is why they are blocking app deployment and local runs until the validation errors are addressed.

The validations ensure correct usage of decorators, function/class definitions, and type hints. In particular:

* Ensures all decorators are called with parentheses (e.g., @function(), @application(), @cls()).
* Detects duplicate function definitions across different files.
* Detects duplicate class definitions across different files.
* Validates that all objects that have @cls():
** Are actual classes (not functions).
** Have a proper __init__ signature (first parameter is 'self',
   no extra parameters).
** Are defined at the module level (not nested).
* Checks that at least one application function is defined with @application().
* Validates all objects decorated with @function():
** Are not classes decorated as functions.
** Are defined at the module level (not nested).
* Application functions have both @application() and @function() decorators.
* Application functions have correct parameter count (one for functions, two for methods).
* Application functions have type hints for input and return types.
* Methods are not defined in nested classes.
* Methods have @cls() decorator on their class.
* Methods have 'self' as the first parameter.

Example error outputs:

```
poetry run tensorlake deploy ./reference_app/reference_app.py
⚙️  Preparing deployment for applications from ./reference_app/reference_app.py
‼️  Error: reference_app:32:processor: 
@cls decorator is missing its parenthesis. Please replace it with @cls().

‼️  Error: reference_app:32:processor: 
@function decorator is missing its parenthesis. Please replace it with @function().

‼️  Error: 
No application function is defined. Please add at least one application function by adding @application() decorator to it.

‼️  Deployment aborted due to code validation errors, please address them before deploying.
Aborted!
```

Extensive set of tests is added for the validation because it turned out that it's really tricky to get them right. One coding
error easily results in multiple irrelevant error messages because application model gets into an inconsistent state. So I spent a bunch of time on obtaining the required information to make reliable validations. All the tests verify that there are no false positive validation messages.

### DIFF
--- a/tests/applications/validation/class/missing_decorator_call/test_missing_all.py
+++ b/tests/applications/validation/class/missing_decorator_call/test_missing_all.py
@@ -33,7 +33,7 @@ class TestMissingDecoratorCall(unittest.TestCase):
             if "@cls" in validation_message.message:
                 self.assertEqual(
                     validation_message.message,
-                    "Class is missing () after @cls decorator. Please use @cls() syntax.",
+                    "@cls decorator is missing its parenthesis. Please replace it with @cls().",
                 )
                 self.assertIsNotNone(validation_message.details)
                 class_details: ClassDetails = validation_message.details
@@ -56,7 +56,7 @@ class TestMissingDecoratorCall(unittest.TestCase):
             elif "@application" in validation_message.message:
                 self.assertEqual(
                     validation_message.message,
-                    "@application decorator is missing (). Please use @application() syntax.",
+                    "@application decorator is missing its parenthesis. Please replace it with @application().",
                 )
                 self.assertIsNotNone(validation_message.details)
                 details: FunctionDetails = validation_message.details
@@ -87,7 +87,7 @@ class TestMissingDecoratorCall(unittest.TestCase):
             elif "@function" in validation_message.message:
                 self.assertEqual(
                     validation_message.message,
-                    "@function decorator is missing (). Please use @function() syntax.",
+                    "@function decorator is missing its parenthesis. Please replace it with @function().",
                 )
                 self.assertIsNotNone(validation_message.details)
                 details: FunctionDetails = validation_message.details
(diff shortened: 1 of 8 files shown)


tests/applications/validation/class/missing_decorator_call/test_missing_all.py
```diff
@@ -0,0 +1,123 @@
...
+        for validation_message in validation_messages:
+            self.assertEqual(
+                validation_message.severity,
+                ValidationMessageSeverity.ERROR,
+            )
+
+            if "@cls" in validation_message.message:
+                self.assertEqual(
+                    validation_message.message,
+                    "Class is missing () after @cls decorator. Please use @cls() syntax.",
```
calavera: I'm concerned that people will see `missing ()` as an incomplete message because it's not usual to see parenthesis mentioned that way. Can these messages have a consistent format? A small suggestion:

```
The @cls decorator is missing its parenthesis. Please replace it with `@cls()`.
The @application decorator is missing its parenthesis. Please replace it with `@application()`.
The @function decorator is missing its parenthesis. Please replace it with `@function()`.
```

eabatalov: Yep, good idea, thank you.

pyproject.toml
```diff
@@ -35,13 +34,25 @@ function-executor = "tensorlake.function_executor.main:main"
 requires = ["poetry==2.0.0"]
 build-backend = "poetry.core.masonry.api"
 
+[tool.black]
```
calavera, line 37: This syntax might be better because they add to the default excluded directories:

```
[tool.black]
extend_exclude = "vendor"

[tool.isort]
profile = "black"
extend_skip = ["vendor"]
```

tests/applications/validation/class/missing_decorator_call/test_missing_all.py
eabatalov: Applied this comment with a couple changes:
* Didn't use `` because I don't use them in other messages.
* Didn't use "The" article because I don't use it in other messages.

This can be changed if we need to though (required updating 20+ tests).
I preferred consistency, this is why and was too lazy to change all tests.

PR #457, opened 2025-12-02
RequestContextHTTPClient + Server implementation in local mode

eabatalov:
RequestContextHTTPClient is a picklable RequestContext implementation that provides
all RequestContext functionality in any Python thread and subprocess.

It exchanges small http requests with an HTTP Server which implements:
presigning of read/write blobs, commiting multipart uploads to blob store,
adding request metrics, sending function progress updates. The BLOBs
are not sent to the Server. BLOB store client is directly invoked
by RequestContextHTTPClient.

User code can use RequestContextHTTPClient in a thread or subprocess by passing the result of RequestContext.get()
as an argument when creating a Thread or Process. This usage scenario is demonstrated
in the added tests.

When user passes RequestContextHTTPClient into a Thread the context is passed by reference.
RequestContextHTTPClient is thread safe so this is okay. When user passes RequestContextHTTPClient
into a Process, Python runtime pickles RequestContextHTTPClient and then unpickles its clone
in the subprocess so the subprocess uses a clone of the original RequestContextHTTPClient.
RequestContextHTTPClient has all the data required to reestablish http connection to the
HTTP Server and tell the Server which allocation and request the http requests are for.

This PR adds implementation of the HTTP Server in local mode.
Remote (FE) side HTTP Server implementation will be done in a
separate PR. New integration tests are covering all RequestContext
functionaility working correctly when accessed from a user function,
a thread and a subprocess.

Each LocalRunner (one per request) uses its own HTTP Server. Each
HTTP Server uses a unique ephemeral port where the LocalRunner
request's RequestContextHTTPClient connects to. So multiple local runners
can exist at the same time without any problems.

---

> [!NOTE]
> Introduces a picklable RequestContext HTTP client/server for local mode with thread/process support, refactors BLOB store and internal logging, and updates FE/local runner and tests accordingly.
> 
> - **Request Context (Local mode)**:
>   - Add `RequestContextHTTPClient` (picklable, thread-safe) and lightweight local HTTP server/router with handlers for `progress`, `metrics`, and request state (`prepare_read`, `prepare_write`, `commit_write`).
>   - Local handlers print progress/metrics and serve file-backed BLOBs for request state.
> - **BLOB Store**:
>   - Introduce SDK `BLOB`/`BLOBChunk` pydantic models and `BLOBStore` with local FS and S3 backends; parallel chunked `get/put` with `InternalLogger`.
>   - Remove old local blob store; add local `SerializedValueStore` wrapper.
> - **Logging**:
>   - Replace `FunctionExecutorLogger` with picklable `InternalLogger`; unify usage across SDK/FE.
> - **Local Runner**:
>   - Use per-request local HTTP server + `RequestContextHTTPClient`, new value store, cleanup lifecycle; set multiprocessing start method via `setup_multiprocessing`.
> - **Function Executor**:
>   - Migrate to new `InternalLogger`/`BLOBStore`; add blob proto↔SDK converters; validate progress attributes; minor proto doc tweak.
> - **Runtime hooks & multiprocessing**:
>   - Add `_raise_multiprocessing_usage_error`; initialize spawn method; improved SDK usage errors outside function threads.
> - **Cloud events**:
>   - Move CloudEvents helpers under `applications` and update imports; clarify JSON serialization errors.
> - **Tests**:
>   - Add/adjust tests for blob store (local/S3), request context (metrics/progress/state/request_id) in function, thread, and subprocess scenarios; update benchmarks.
> - **Version**: bump to `0.2.89`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e487a2344f9557f66be86da91e199cfac6d53dba. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/tensorlake/function_executor/main.py
```diff
@@ -52,7 +52,7 @@ def main():
 
     # Set "spawn" method because grpc Server doesn't support forking. exec is required.
     # "spawn" is supported by all OSes: Windows, Linux, MacOS.
-    multiprocessing.set_start_method(method="spawn", force=True)
+    multiprocessing.set_start_method(method="spawn")
```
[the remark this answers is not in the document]
eabatalov: This is intentional as no other library code calls this.

src/tensorlake/applications/cloud_events.py
```diff
@@ -0,0 +1,46 @@
...
+def print_cloud_event(
+    event: dict[str, Any],
+    type: str = "ai.tensorlake.event",
+    source: str = "/tensorlake/function_executor/events",
+) -> None:
+    """
+    Takes a dictionary representing an event produced by the executor, wraps it in a CloudEvent and prints it to stdout.
+    """
+    try:
+        print(json.dumps(new_cloud_event(event)), flush=True)
```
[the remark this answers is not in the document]
eabatalov: This is bug in existing code, let's fix it.

tests/applications/request_context/test_mpt_progress.py
```diff
@@ -0,0 +1,104 @@
...
+
+@application()
+@function()
+def func_update_progress(values: tuple[int, int]) -> str:
+    ctx: RequestContext = RequestContext.get()
+    ctx.progress.update(current=values[0], total=values[1])
+    return "success"
+    q: mt_queue.SimpleQueue = mt_queue.SimpleQueue()
+    update_progress_worker(ctx, values, q)
+    return "success" if q.get() is None else "failure"
```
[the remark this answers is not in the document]
eabatalov: Fixed

src/tensorlake/applications/request_context/http_server/server.py
```diff
@@ -0,0 +1,52 @@
...
+    def stop(self):
+        """Stops the HTTP server and releases its resources.
+
+        Blocks until the server is fully stopped.
+        Does nothing if the server is not running.
+        """
+        if self._running:
+            self._running = False
+            # self._running must be checked, otherwise shutdown() will deadlock.
+            self._httpd.shutdown()
```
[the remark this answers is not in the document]
eabatalov: This is okay, we call server.start() and stop() from the same thread always.

src/tensorlake/applications/local/request_context/handlers/request_state/prepare_read.py
```diff
@@ -0,0 +1,37 @@
...
+class LocalPrepareReadHandler(BasePrepareReadHandler):
+    def __init__(self, request_state_dir_path: str):
+        super().__init__()
+        self._request_state_dir_path: str = request_state_dir_path
+
+    def _handle(self, request: PrepareReadRequest) -> PrepareReadResponse:
+        file_path: str = os.path.join(
+            self._request_state_dir_path,
+            request.state_key,
+        )
```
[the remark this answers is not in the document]
eabatalov: Fixed

src/tensorlake/applications/blob_store/blob.py
```diff
@@ -0,0 +1,22 @@
...
+    # URI of the chunk.
+    # S3 URI if the data is stored in S3.
+    # Starts with "file://"" prefix if the data is stored on a local file system.
+    uri: str
+    # Actual size of chunk data if the BLOB has data.
+    # Requested chunk size if the BLOB has no data yet (can be used for data upload).
+    size: int
+    # None if the BLOB has data.
+    # ETag of the chunk data (generated by BLOB store) if the BLOB chunk was used to upload data.
+    etag: str | None
```
diptanu, line 14: This is for CAS updates? Can you please do a quick check to see how azure and gcp supports cas? I wonder if we are tying ourselves too hard to S3.

src/tensorlake/applications/blob_store/blob.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class BLOB(BaseModel):
+    # ID of the BLOB
+    id: str
+    # Ordered chunks of the BLOB if the BLOB has data.
+    # Ordered chunks of the BLOB that can be used to upload data if the BLOB has no data yet.
+    chunks: list[BLOBChunk]
```
diptanu, line 22: This is to support large payloads which don't fit in a single key?

src/tensorlake/applications/blob_store/blob_store.py
```diff
@@ -1,11 +1,11 @@
 from concurrent.futures import FIRST_EXCEPTION, Future, ThreadPoolExecutor, wait
 from dataclasses import dataclass
-from typing import List
```
diptanu, line 3: Why was this removed? Looks like a bunch of List[T] was replaced with list[T] Isn't List the right annotations for lists?

src/tensorlake/applications/blob_store/blob_store.py
```diff
@@ -1,11 +1,11 @@
...
+from typing import Any
 
-from tensorlake.applications import InternalError
+from tensorlake.applications.interface.exceptions import InternalError
+from tensorlake.applications.internal_logger import InternalLogger
```
diptanu, line 6: Just a rename to make it more clear that this is for internal logging?

src/tensorlake/applications/interface/request_context.py
```diff
@@ -32,14 +32,14 @@ def get(self, key: str, default: Any | None = None) -> Any | None:
 class RequestMetrics:
     """Abstract interface for reporting application request metrics."""
 
-    def timer(self, name: str, value: float):
+    def timer(self, name: str, value: float) -> None:
```
diptanu, line 35: We didn't really end up using these interfaces in production. Looks like serverless platforms just let people send metrics to an OTEL endpoint using whatever Otel client users like. May be we should just delete these metrics interfaces.

src/tensorlake/applications/request_context/http_server/router.py
```diff
@@ -0,0 +1,101 @@
+import http.server
+
+from tensorlake.applications.internal_logger import InternalLogger
+
+from .handlers.handler import Handler, Request, Response
+from .route import Route
+
+
+class Router(http.server.BaseHTTPRequestHandler):
```
diptanu, line 9: Feel like we have created a HTTP framework inside the FE.  Isn't there any lightweight web server framework without any external dependencies we could have vendored?

src/tensorlake/applications/blob_store/blob.py
eabatalov: Yep, we're handling all the request context state values in dataplane the same way as function outputs. So there's not limits on request context state value sizes. As we already have all executor APIs for this provided to FE, there's no extra cost in having this.

src/tensorlake/applications/blob_store/blob.py
eabatalov: We don't use CAS at the moment. This etag is required for multipart uploads to S3 which we're using to mitigate 5 GB max S3 upload size limit. Each chunk (part) is uploaded by FE to S3 and we get an etag for each uploaded part. To commit a multipart upload to a BLOB we need to provide to S3 an index and etag of each uploaded part comprising the final uploaded BLOB.
There's no CAS even with etags used this way because there can be any number of concurrent multipart uploads created for the same Object Key and the last writer still wins (the last commited multipart upload).

src/tensorlake/applications/blob_store/blob.py
eabatalov: The key is the same though. The chunks are addressing the problem that max single S3 upload is 5 GB (and different in different object store implementations). Also the chunks are uploaded and read in parallel. So it increases speed quite a lot ~2x at least iirc.

src/tensorlake/applications/blob_store/blob_store.py
eabatalov: Not anymore, Python now supports some regular types as type annotations. Like tuple[], dict[], list[].

src/tensorlake/applications/blob_store/blob_store.py
eabatalov: The
> from tensorlake.applications import InternalError
to
> from tensorlake.applications.interface.exceptions import InternalError

is just importing the same thing without waiting for whole Applications SDK initialization to finish. As BLOB store was moved from FE into apps SDK it now can't take a dependency on fully initialized Apps SDK because initilizing BLOB store is now part of initializing apps SDK.

src/tensorlake/applications/interface/request_context.py
eabatalov: Possibly, but let's think about it separately. Having metrics supported out of the box might be handy but then we need to provide full UX for them which we won't be priroriting yet. Let's discuss separately.

src/tensorlake/applications/request_context/http_server/router.py
eabatalov: Yep, let's vendor something if we more feature out of it. Cause the features we implement here is just routing.

PR #470, opened 2025-12-11
Add try catch and flush to internal logger message format and print

eabatalov:
This is to make sure that none of the internal logger code paths fail silently. Also flush makes logging more reliable in case if i.e. FE process crashes immediately after logging a line.


calavera (commented):
good change

calavera (approved):

PR #475, opened 2025-12-17
IPC request context for remote mode (Function Executor)

eabatalov:
Each FE creates an HTTP server and an httpx.Client for it on FE initialization. Then both are reused during FE lifetime to call and implement RequestContext methods for all allocations running in the FE.

Also unify request progress and metrics printing code in FE and local mode. This helps keeping the printed messages consistent in both modes.

Also stop sending user metrics to executor. We'll use the same approach as we use for user logs: print them as cloud events so all the user metrics are visible to users in request events. Later if we want we can use otel for this - no SDK interface changes needed for this.

All the tests that are using RequestContext in multi-threaded and multiprocessing modes are now running on both local and remote modes.


src/tensorlake/function_executor/allocation_runner/allocation_runner.py
```diff
@@ -245,6 +254,38 @@ def deliver_allocation_update(self, update: AllocationUpdate) -> None:
...
+        if isinstance(operation, PrepareReadRequest):
+            return self._request_state.prepare_read(operation)
+        elif isinstance(operation, PrepareWriteRequest):
+            return self._request_state.prepare_write(operation)
+        elif isinstance(operation, CommitWriteRequest):
+            return self._request_state.commit_write(operation)
+        elif isinstance(operation, FunctionProgressUpdateRequest):
+            return self._allocation_progress.update(operation)
+        else:
+            raise InternalError(
```
calavera: should this be a `RuntimeError` instead of `InternalError`?

eabatalov: Sounds good. In this particular case we don't need to use SDK exceptions because these are not raised to user code here. So using RuntimeError is fine and a bit less misleading, cause ppl might assume we actually raise to user code here if they see SDK Exception (InternalError).

PR #477, opened 2025-12-18
Support async function in exponential_backoff decorator

eabatalov:
The decorator never retried async functions because it just returned coroutine returned by the async function without awaiting for it.

Also fix retry logic to not count the first function execution attempt as a retry. Previously 3 max retries meant 3 max total function execution, now it means 4 max total function executions. This looks more correct to me.

### DIFF
--- a/src/tensorlake/utils/retries.py
+++ b/src/tensorlake/utils/retries.py
@@ -1,7 +1,9 @@
+import asyncio
+import inspect
 import random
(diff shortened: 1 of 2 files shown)


sgirones (approved):

eabatalov:
Let's merge as is not to spend more time on this subtile thing.

PR #480, opened 2025-12-19
WIP: fix: deduplicate app functions for image building

daniel-white:
change to use a set instead of a list to track duplicate functions within a single application

### DIFF
--- a/src/tensorlake/applications/image.py
+++ b/src/tensorlake/applications/image.py
@@ -23,7 +23,7 @@ _SDK_VERSION: str = importlib.metadata.version("tensorlake")
 class ImageInformation:
     image: Image
     # Functions that are using the image.
-    functions: List[Function]
+    functions: set[Function]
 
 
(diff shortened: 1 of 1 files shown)


src/tensorlake/applications/image.py
```diff
@@ -23,7 +23,7 @@
 class ImageInformation:
     image: Image
     # Functions that are using the image.
-    functions: List[Function]
+    functions: set[Function]
```
daniel-white, line 26: is this a breaking change? should i change this back?

eabatalov: This won't solve the problem afaik, because we need to dedup images, not Functions in Dict[Image, ImageInformation] returned by image_infos() and they are already deduped, see line 35.

Besides that Functions are not intended to be stored in sets, they don't have == operator or hash implemented.

What needs to be done on image builder side afaik is to not build image per function but instead build image once and then associate it with every function that uses it.

PR #486, opened 2025-12-28
change application call semantics

### LINKED ISSUE #485: Improve ergonomics of Application Functions
Claude Code and other coding agents make mistakes all the time for the application functions because it's pretty unusual for any framework to have the limitation of having exactly one input parameter as the entrypoint. We have to extend this so that the very first thing an agent writes doesn't lead to an error. 
## Current Behavior

Application functions currently require **exactly one parameter** (the request input). The following patterns are not supported:

```python
# Multiple parameters 
@application()
@function()
def process_order(item: Item, user: dict, importance: int) -> str:
    ...

# No parameters 
@application()
@function()
def health_check() -> str:
    return "OK"
```

Users must wrap all inputs into a single object:

```python
class ProcessOrderRequest(BaseModel):
    item: Item
    user: dict
    importance: int

@application()
@function()
def process_order(request: ProcessOrderRequest) -> str:
    ...
```

## Proposed Behavior

Allow application functions to accept **zero, one, or multiple parameters**, mapping them to a JSON request body:

### Multiple parameters

```python
@application()
@function()
def process_order(item: Item, user: dict, importance: int) -> str:
    ...
```

Request body:

```json
{
    "item": {
        "name": "Foo",
        "description": "The pretender",
        "price": 42.0,
        "tax": 3.2
    },
    "user": {
        "username": "dave",
        "full_name": "Dave Grohl"
    },
    "importance": 5
}
```

### Single parameter (backward compatible)

```python
@application()
@function()
def create_item(item: Item) -> str:
    ...
```

Request body:

```json
{
    "name": "Foo",
    "description": "The pretender",
    "price": 42.0,
    "tax": 3.2
}
```

### No parameters

```python
@application()
@function()
def health_check() -> str:
    return "OK"
```

Request body: `{}` or no body required

This follows the convention how FastAPI handles body of requests.

diptanu:
Make application calling more flexible

Fixes #485 

Bunch of changes in how we deal with types in function args and do validation, need careful review before merging this. 

---

> [!NOTE]
> Introduces flexible application call semantics and aligns local/remote execution and validation.
> 
> - **Application calls**: Support `0/1/many` parameters; map JSON payload objects to `args/kwargs`, exclude class `self`, skip variadics; coerce dicts to Pydantic models via serializer; robust single-arg deserialization; helper utilities added (`_get_application_param_count`, `_coerce_payload_to_kwargs`, `_deserialize_to_type`).
> - **Runners**: Local runner branches by param count; multi-arg payloads coerced to proper types; Remote runner converts Pydantic instances to dicts for `pickle` serializer to avoid class reference issues.
> - **Function Executor**: Reconstructs application call args for `0/1/many` params; inserts `self` when needed; passes `function` into reconstruction; keeps SDK-created calls unchanged.
> - **Validation**: Remove "exactly one/two param" constraint; now require type hints on all application parameters; keep return type-hint requirement.
> - **Tests**: Add comprehensive integration tests for zero/single/multi-param functions (incl. defaults, class methods, JSON/pickle/Pydantic), and update validation tests; remove obsolete tests enforcing old signature rules.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6c30fdbda4e369f97822b0e68bac3fb41185e14e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/tensorlake/applications/function/type_hints.py
```diff
@@ -134,3 +134,36 @@ def _resolve_union_type_hint(union_type_hint: Any) -> List[Any]:
         str | FunctionCall -> [str, FunctionCall]
     """
     return [t for t in get_args(union_type_hint)]
+
+
+def coerce_to_type(value: Any, type_hint: Any) -> Any:
```
eabatalov: We have Pydantic support in JSON serializer already, what this function adds on top of that?

src/tensorlake/applications/local/runner.py
```diff
@@ -217,9 +219,23 @@ def _run(self) -> LocalRequest:
                 payload=serialized_payload,
                 payload_content_type=payload_metadata.content_type,
             )
-            app_function_call_awaitable: FunctionCallAwaitable = self._app.awaitable(
-                payload
-            )
+            # Create awaitable based on number of parameters
+            param_count = _get_application_param_count(self._app)
```
eabatalov, line 223: This code path needs to be extracted into a common function reused between FE and local runner.

src/tensorlake/applications/remote/runner.py
```diff
@@ -33,10 +56,17 @@ def run(self) -> Request:
             app_manifest.entrypoint.input_serializer
         )
 
+        # For pickle serializer, convert Pydantic models to dicts to avoid
+        # class reference issues across process boundaries. The FE will
+        # reconstruct Pydantic models using coerce_to_type and type hints.
+        payload_to_serialize = self._payload
+        if isinstance(input_serializer, PickleUserDataSerializer):
+            payload_to_serialize = _convert_pydantic_to_dict(self._payload)
```
eabatalov, line 64: Shouldn't we just create a kwargs dict and serialized it when we call a function?

src/tensorlake/applications/remote/runner.py
```diff
@@ -1,15 +1,38 @@
...
 
 
+def _convert_pydantic_to_dict(obj: Any) -> Any:
+    """Recursively converts Pydantic models to dicts.
+
+    This is used for pickle serialization to avoid class reference issues
+    when sending Pydantic instances across process boundaries.
+    """
+    if isinstance(obj, pydantic.BaseModel):
+        return obj.model_dump()
```
eabatalov, line 26: We handle all special Pydantic scenarios in json use data serializer. We need to try to keep all such code in common places like json serializer to not create code duplications.

src/tensorlake/applications/validation/validate.py
```diff
@@ -452,37 +452,23 @@ def _validate_application_function(
...
-    if signature_is_valid:
-        # Warning: if you want to delete this or reduce severity then add a test that verifies that things work without type hints.
-        request_input_type_hints: list[Any] = function_arg_type_hint(function, -1)
-        if len(request_input_type_hints) == 0:
+    # Get parameters excluding 'self' for class methods
+    params = list(signature.parameters.values())
+    if (
+        function_details.class_name is not None
+        and len(params) > 0
+        and params[0].name == "self"
```
eabatalov: Nit. Let's just remove the first parameter from validation. No need to assume "self", this is not strictly required.

src/tensorlake/function_executor/allocation_runner/allocation_runner.py
```diff
@@ -745,10 +748,12 @@ def _run_allocation(self) -> AllocationResult:
 
         # This is internal FE code.
         args, kwargs = reconstruct_function_call_args(
+            function=self._function,
             function_call_metadata=function_call_metadata,
             arg_values=arg_values,
-            function_instance_arg=self._function_instance_arg,
```
eabatalov, line 750: Why we remove function_instance_arg? reconstruct_function_call_args returns a fully reconstructed args, kwargs already with needing any extra modifications by caller.

src/tensorlake/function_executor/allocation_runner/sdk_algorithms.py
```diff
@@ -454,29 +454,67 @@ def validate_and_deserialize_function_call_metadata(
...
 def reconstruct_function_call_args(
+    function: Function,
     function_call_metadata: FunctionCallMetadata | ReduceOperationMetadata | None,
     arg_values: Dict[str, Value],
-    function_instance_arg: Any | None,
 ) -> tuple[List[Any], Dict[str, Any]]:
-    """Returns function call args and kwargs reconstructed from arg_values."""
+    """Returns function call args and kwargs reconstructed from arg_values.
+
+    For class methods, inserts the class instance as the first argument (self).
```
eabatalov: This is not true anymore after this change. Also this comment doesn't add much value.

src/tensorlake/function_executor/allocation_runner/sdk_algorithms.py
```diff
@@ -454,29 +454,67 @@ def validate_and_deserialize_function_call_metadata(
...
             arg_values=arg_values,
         )
 
-    if function_instance_arg is not None:
-        set_self_arg(args, function_instance_arg)
-
     return args, kwargs
 
 
+def _reconstruct_application_function_call_args(
```
eabatalov, line 490: Can we reuse this logic between local runner and FE?

src/tensorlake/function_executor/allocation_runner/sdk_algorithms.py
```diff
@@ -454,29 +454,67 @@ def validate_and_deserialize_function_call_metadata(
...
+    Handles three cases:
+    - Zero parameters: returns ([], {})
+    - Single parameter (backward compatible): returns ([payload], {})
+    - Multiple parameters: returns ([], {param_name: value, ...})
+    """
+    signature: inspect.Signature = function_signature(function)
+    params = list(signature.parameters.values())
+
+    # Exclude 'self' parameter for class methods
+    if len(params) > 0 and params[0].name == "self":
```
eabatalov: Just skip the first positional parameter if the function is a class method.

src/tensorlake/function_executor/allocation_runner/sdk_algorithms.py
```diff
@@ -454,29 +454,67 @@ def validate_and_deserialize_function_call_metadata(
...
+    # Zero parameters case
+    if len(params) == 0:
+        return [], {}
+
+    # Single parameter case (backward compatible)
+    if len(params) == 1:
+        return [payload], {}
+
+    # Multiple parameters case - map payload dict keys to kwargs
+    if not isinstance(payload, dict):
```
eabatalov: The payload is coming from user supplied json string. This should be SDKUsageError with a clear explanation on how we got there and what user needs to change in their json payload. We also need test for this as it's going to be a very common error scenario.

src/tensorlake/applications/remote/runner.py
```diff
@@ -33,10 +56,17 @@ def run(self) -> Request:
             app_manifest.entrypoint.input_serializer
         )
 
+        # For pickle serializer, convert Pydantic models to dicts to avoid
```
eabatalov, line 59: There's no need to add any special cases here for Pydantic because we use json serialization for app functions.

src/tensorlake/applications/function/application_call.py
```diff
@@ -1,11 +1,51 @@
...
+
+    # Exclude 'self' parameter for class methods
+    if len(params) > 0 and params[0].name == "self":
+        params = params[1:]
+
+    kwargs: Dict[str, Any] = {}
+    for param in params:
+        if param.name in payload:
+            raw_value = payload[param.name]
+            kwargs[param.name] = coerce_to_type(raw_value, param.annotation)
```
eabatalov: Let's not modify original values so we don't need to modify them back. Relying on accurate type hints by users unnecessary without a real necessity would result in more complex behavior which is hard to understand for us and user and to debug.

eabatalov: I mean, If a user supplied value is a dict then we just map its values to app function parameters by parameter names. Nothing else is needed afaik.

src/tensorlake/applications/function/application_call.py
```diff
@@ -1,11 +1,51 @@
...
+        elif param.default is not inspect.Parameter.empty:
+            kwargs[param.name] = param.default
+        else:
+            raise SDKUsageError(
```
eabatalov, line 157: We can just call the function and see it fails. It's going to be a very regular Python error same as when user calls a function without passing all required parameters. I'm not happy about us duplicating regular Python logic in our code base.

eabatalov (changes requested):
General comment:

I think 50% of this PR https://github.com/tensorlakeai/tensorlake/pull/486/files is just not need plus all the new tests.

My understanding is that all we needed to do is:
1. If app function has > 1 parameter then try to deserialize the payload into a dict. If it fails then SDKUsageError with clear explanation on what happened and what needs to be done by user to fix it.
2. For each app function parameter name, get a value from the dict with same key. If the value is a dict itself then serialize it back to json and encode it into bytes and deserialize it with app function's user data serializer + app function's parameter type hint. (the encode into bytes part can be fixed but I won't call it a priority cause json objects are typically small). This is for json. If pickle is used then just use the value from the dict as they are already deserialized into correct types. A proper fix here is to incorporate this logic into user data serializers, i.e. add a method to them which accepts type hints per dict key name and does all this magic.
3. Call the app function with the resulting deserialized kwargs.

This will support default parameters and missing parameters in the dict too with native Python error messages which are easy to google about and understand.
With this simple logic we'll have much less to debug and wonder what's going on.

For json deserializer implementation we'd need ideally something like "leave json dict values as is - as json strings" algorithm. It might be possible to implement this using the customizations availble using https://docs.python.org/3/library/json.html#json.load parameters.

diptanu:
**Re: Lines 26, 59, 64 - Pydantic special-casing in remote/runner.py**

This code is intentionally placed here rather than in the pickle serializer because:

1. **App functions support pickle** via `@application(input_deserializer="pickle")`. While JSON is the default, pickle is a valid option.

2. **Pickle + Pydantic across processes fails** due to class reference issues. When user sends `Item(name="foo")`, pickle stores the class as `tests.module.Item`, but FE's `__main__` can't find it.

3. **We can't put this in the pickle serializer** because that would break request state. The pickle serializer is used for more than app payloads - request state stores Pydantic models and expects them back as Pydantic, not dicts.

4. **RemoteRunner is the right scope** - this conversion only affects app function payloads when using pickle. It's targeted, not global.

**Alternative:** If we want to remove this, we could:
- Remove pickle support for app functions entirely
- Or document that pickle users must send dicts, not Pydantic instances

Let me know if you'd prefer either alternative.

src/tensorlake/applications/local/runner.py
diptanu: Addressed

eabatalov:
This PR still violates lots of layering and scoping in SDK.
I'll take a look into this on Monday to see if the original layering and scoping can be preserved without too much refactoring.

eabatalov:
Superseded by https://github.com/tensorlakeai/tensorlake/pull/502

PR #502, opened 2026-01-08
Multi-parameter application function calls

eabatalov:
Application function calls now support any number of positional and keyword arguments, with default argument values applied if some arguments are not passed.

SDK run application functions which are part of SDK interface are updated with new backward compatible signature:

```
def run_local_application(application: Function | str, *args, **kwargs) -> Request:
def run_remote_application(application: Function | str, *args, **kwargs) -> Request:
```

In remote mode the API client passes empty HTTP body to Server if application function is called without any parameters. If application function is called with a single position arg then the current approach is used where the arg is serialized into HTTP body and HTTP content type is the same as serialized arg content type.
In all other cases multipart HTTP request is used with each arg serialized as a separate part in the request body.  A parser with zero data copy and bytes.find() used for substring search is implemented for multipart request bodies. bytes.find() is more than 1000x faster than searching substrings in Python because it's fully implemented in C and uses fast substring search algorithms.

These first two call modes (no parameters and one positional parameter) allow users to use non-multipart (simpler) curl commands.

Field names in multipart mode are used to map body parts to application function parameters. If a field name is a number then it's an index of positional argument of the function. If a field name is not a number then it's name of a keyword argument of the function. The mapping logic is permissive, i.e. it ignores non- existent keyword arguments. This allows to simplify code changes for users. I.e. they can send a request with a new arg in it while the request not failing if it land on application function of previous version that doesn't have the arg added yet. Web frameworks typically ignore parts that are not consumed by user code too.

For multypart and empty body calls to work no Server changes are required at the moment because Server just forwards the HTTP bodies with their content type to FE.

FE also now supports mode (turned on using "message/http" content-type) where Server can forward full HTTP request to FE. In this case FE parses full HTTP request and its headers become available to it. These headers would be used in the future for future SDK features. For now we just leave this capability in FE to be able to not fail user requests once Server changes it behavior and starts forwarding full HTTP requests.

I ran all integ tests both with current Server and Server that forwards full http request. All pass. I also added an FE test that uses HTTP request forwarding mode to call application function just to make sure that we don't regress on implementation of this feature while we're not using it on Server side.

Added tests that check all these scenarios.
All the commong new code is shared between local and remote runners.

Also added missing test coverage for regular and application function calls that accept File objects as arguments and return File objects.

### DIFF
--- a/src/tensorlake/applications/applications.py
+++ b/src/tensorlake/applications/applications.py
@@ -1,8 +1,11 @@
-from typing import Any, Generator, Iterator
+from typing import Generator, Iterator
 
 from .interface.function import Function, _is_application_function
 from .interface.request import Request
-from .interface.run import run_local_application, run_remote_application
+from .interface.run import (
+    run_local_application,
+    run_remote_application,
+)
 
 # Internal utilities for working with applications.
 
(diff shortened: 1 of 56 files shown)


src/tensorlake/applications/applications.py
```diff
@@ -18,13 +21,15 @@ def filter_applications(
         yield function
 
 
-def run_application(application: Function | str, payload: Any, remote: bool) -> Request:
+def run_application(
```
eabatalov, line 24: This function is not part of SDK interface, it's only used for our testing. This is why it's okay to change it in non backward compatible way.

PR #503, opened 2026-01-11
Implement a new Logger exposed in the applications interface.

### LINKED ISSUE #476: Remove structlog and expose InternalLogger as Logger for applications
Now that we have internal_logger.py in the applications module, it might be a good idea to remove the structlog module and vendored package and expose our own logger to applications directly. That would help us reduce package size and duplicated code in the SDK.

calavera:
This replaces the vendored structlog package with a lighter module. It still keeps the most used features from structlog that we use in applications:

- Structured logging as JSON objects.
- Context aware to add more information as the application progresses.
- Print backtraces as json strings for better output.

Fixes #476

### DIFF
--- a/src/tensorlake/applications/internal_logger.py
+++ b/src/tensorlake/applications/internal_logger.py
@@ -25,7 +25,7 @@ class InternalLogger:
         self,
         context: Dict[str, Any],
         destination: LOG_FILE,
-        as_cloud_event: bool = True,
+        as_cloud_event: bool,
     ):
         self._context: Dict[str, Any] = context
         self._destination: InternalLogger.LOG_FILE = destination
@@ -52,7 +52,7 @@ class InternalLogger:
         self.__init__(
             context=state["context"],
             destination=state["destination"],
-            as_cloud_event=state.get("as_cloud_event", True),
+            as_cloud_event=state["as_cloud_event"],
         )
 
     @classmethod
@@ -64,6 +64,7 @@ class InternalLogger:
         return InternalLogger(
             context=kwargs,
             destination=InternalLogger.LOG_FILE.STDOUT,
+            as_cloud_event=True,
         )
 
     def bind(self, **kwargs) -> "InternalLogger":
@@ -147,7 +148,12 @@ class InternalLogger:
         context["event"] = message
 
         if "exc_info" in context:
-            exc_info: tuple | None = context["exc_info"]
+            exc_info: (
+                tuple[type[BaseException], BaseException, Any]
+                | BaseException
+                | bool
+                | None
+            ) = context["exc_info"]
             # Handle exc_info=True (capture current exception from sys.exc_info())
             if exc_info is True:
                 exc_info = sys.exc_info()
--- a/tests/applications/test_structlogger.py
+++ b/tests/applications/test_structlogger.py
@@ -61,6 +61,15 @@ def test_logging_error_exception(payload: DummyPayload) -> str:
     return "error_exception"
 
 
+@application()
+@function(description="test logging with bind")
+def test_logging_bind(payload: DummyPayload) -> str:
+    logger = Logger.get_logger(module="test")
+    bound_logger = logger.bind(request_id="123", user="test_user")
+    bound_logger.info("info with bound context")
+    return "bind"
+
+
 class TestInternalLoggerLogging(unittest.TestCase):
     def setUp(self):
         self.captured_output = io.StringIO()
@@ -134,6 +143,22 @@ class TestInternalLoggerLogging(unittest.TestCase):
         self.assertEqual(log_entry["level"], "error")
         self.assertEqual(log_entry["event"], "error with exception")
         self.assertIn("exception", log_entry)
+
+    def test_bind_log(self):
+        request: Request = run_local_application(
+            test_logging_bind,
+            DummyPayload(),
+        )
+        self.assertEqual(request.output(), "bind")
+        output = self.captured_output.getvalue()
+        lines = output.strip().split("\n")
+        self.assertEqual(len(lines), 1)
+        log_entry = json.loads(lines[0])
+        self.assertEqual(log_entry["level"], "info")
+        self.assertEqual(log_entry["event"], "info with bound context")
+        self.assertEqual(log_entry["module"], "test")
+        self.assertEqual(log_entry["request_id"], "123")
+        self.assertEqual(log_entry["user"], "test_user")
 
 
 if __name__ == "__main__":


src/tensorlake/applications/internal_logger.py
```diff
@@ -21,9 +21,17 @@ class LOG_FILE(Enum):
 
     """Picklable internal logger for use in FE and SDK."""
 
-    def __init__(self, context: Dict[str, Any], destination: LOG_FILE):
+    def __init__(
+        self,
+        context: Dict[str, Any],
+        destination: LOG_FILE,
+        _dict_traceback: bool = False,
```
eabatalov: Please don't use _argument_names and default values unless there's a really good reason to.
_ is not a known to me coding convention.
default values reduce discipline and redability in large code bases.

src/tensorlake/applications/internal_logger.py
```diff
@@ -38,13 +46,17 @@ def __getstate__(self):
         return {
             "context": self._context,
             "destination": self._destination,
+            "_dict_traceback": self._dict_traceback,
```
eabatalov: Same here about _.

src/tensorlake/applications/interface/logger.py
```diff
@@ -0,0 +1,69 @@
...
+        """Delegate all attribute access to the underlying logger."""
+        return getattr(self._logger, name)
+
+    @classmethod
+    def get_logger(cls, **kwargs) -> "Logger":
+        """Gets the root logger with the given context.
+
+        Doesn't raise any exceptions.
+        """
+        return cls(context=kwargs)
```
eabatalov, line 41: Later we can implement context pre-initialization logic in this factory function like adding allocation_id or anything else. This would work I think.

src/tensorlake/applications/interface/logger.py
```diff
@@ -0,0 +1,69 @@
...
+            destination: Log file destination (internal use)
+        """
+        if destination is None:
+            destination = InternalLogger.LOG_FILE.STDOUT
+
+        self._logger = InternalLogger(
+            context, destination, _dict_traceback=True, _as_cloud_event=False
+        )
+
+    def __getattr__(self, name):
```
eabatalov: What this is for?

src/tensorlake/applications/interface/logger.py
```diff
@@ -0,0 +1,69 @@
...
+    This logger automatically transforms exception tracebacks into a structured,
+    machine-readable dictionary format suitable for JSON output and log aggregators.
+
+    Outputs plain JSON without CloudEvent wrapping.
+    """
+
+    def __init__(
+        self,
+        context: dict,
+        destination=None,
```
eabatalov: This is a private constructor which people shouldn't call so let's either add proper type hint to destination and remove default None value or just remove destination arg here and InternalLogger.LOG_FILE.STDOUT to InternalLogger() call.

src/tensorlake/applications/internal_logger.py
```diff
@@ -135,9 +152,21 @@ def _format_message(self, level: str, message: str, **kwargs) -> str:
         context["event"] = message
 
         if "exc_info" in context:
-            context["exception"] = "".join(
-                traceback.format_exception(context["exc_info"])
-            )
+            exc_info = context["exc_info"]
```
eabatalov: Please add a type hint here

src/tensorlake/applications/internal_logger.py
```diff
@@ -147,6 +176,44 @@ def _format_message(self, level: str, message: str, **kwargs) -> str:
...
-        )
+        if self._as_cloud_event:
+            return json.dumps(
+                new_cloud_event(context, source="/tensorlake/function_executor/logger")
+            )
+        else:
+            context["timestamp"] = event_time()
+            return json.dumps(context)
+
+    def _format_exception_dict(self, exc_info: tuple) -> Dict[str, Any]:
```
eabatalov: Please add full tuple[, , ,] type hint here

src/tensorlake/applications/interface/logger.py
```diff
@@ -0,0 +1,69 @@
...
+from ..internal_logger import InternalLogger
+
+
+class Logger:
+    """Public logger for applications with dict_traceback support enabled.
+
+    This logger automatically transforms exception tracebacks into a structured,
+    machine-readable dictionary format suitable for JSON output and log aggregators.
+
+    Outputs plain JSON without CloudEvent wrapping.
```
eabatalov: What does it do that? Please explain in this comment so UX is clear for any future reader (and me).

src/tensorlake/applications/interface/logger.py
```diff
@@ -0,0 +1,69 @@
+"""Public logger for applications with structured output."""
+
+from ..internal_logger import InternalLogger
+
+
+class Logger:
+    """Public logger for applications with dict_traceback support enabled.
+
+    This logger automatically transforms exception tracebacks into a structured,
```
eabatalov: As we're already using structured logging in InternalLogger it looks to me like the use of standard traceback.format_exception there is a misalignment because it doesn't output structured exception info.

Let's just use the structured exception_info approach in all cases and remove "dict_traceback" fields everywhere. It'd result in less human readable an not pretty printed backtraces in Executor FE DD logs right now but we have to stop using DD for FE logs sooner or later anyway and it's right in long term store backtraces in structured way so they can be searched if we need to.

src/tensorlake/applications/internal_logger.py
```diff
@@ -147,6 +176,44 @@ def _format_message(self, level: str, message: str, **kwargs) -> str:
...
+        """Formats exception info as a structured dictionary.
+
+        Similar to structlog's dict_tracebacks processor, transforms exception
+        information into a machine-readable dictionary suitable for JSON output.
+        """
+        exc_type, exc_value, exc_tb = exc_info
+        if not exc_type:
+            return {}
+
+        frames = []
```
eabatalov: Please add type hints for all local vars if possible

src/tensorlake/applications/internal_logger.py
```diff
@@ -147,6 +176,44 @@ def _format_message(self, level: str, message: str, **kwargs) -> str:
...
+        tb = exc_tb
+        while tb is not None:
+            frame = tb.tb_frame
+            frames.append(
+                {
+                    "filename": frame.f_code.co_filename,
+                    "lineno": tb.tb_lineno,
+                    "name": frame.f_code.co_name,
+                    "locals": {
```
eabatalov: This is formatting local vars into strings afaik. For large variables or memory this can be huge. Are you sure it's worth doing this? This can also result in OOM due to memory usage surge.
I think this is too unsafe. There's also no way to do a partial convertion of an object into a string so I'd prefer to not do this.

src/tensorlake/applications/internal_logger.py
```diff
@@ -147,6 +176,44 @@ def _format_message(self, level: str, message: str, **kwargs) -> str:
...
+                        k: str(v)
+                        for k, v in frame.f_locals.items()
+                        if not k.startswith("_")
```
eabatalov: Prefix "_", is it some sort of convention to not print stuff?

src/tensorlake/applications/internal_logger.py
eabatalov: At the very least this shouldn't be enabled by default as an unsafe feature, I think.

calavera:
@eabatalov I've made updates to address your comments. Basically simplifying the public implementation. Please, take another look.

src/tensorlake/applications/internal_logger.py
```diff
@@ -21,9 +21,15 @@ class LOG_FILE(Enum):
 
     """Picklable internal logger for use in FE and SDK."""
 
-    def __init__(self, context: Dict[str, Any], destination: LOG_FILE):
+    def __init__(
+        self,
+        context: Dict[str, Any],
+        destination: LOG_FILE,
+        as_cloud_event: bool = True,
```
eabatalov: default parameter values :( 
It's easy to find all the call sites using IDE where IntenalLogger is created and change them instead of leaving it like this and let call sites call reader guess what's happening to this parameter.

src/tensorlake/applications/internal_logger.py
```diff
@@ -38,13 +44,15 @@ def __getstate__(self):
...
             "destination": self._destination,
+            "as_cloud_event": self._as_cloud_event,
         }
 
     def __setstate__(self, state: dict[str, Any]):
         """Set the state for unpickling."""
         self.__init__(
             context=state["context"],
             destination=state["destination"],
+            as_cloud_event=state.get("as_cloud_event", True),
```
eabatalov: Just do state["as_cloud_event"]. This should never fail, this is called in forked processes.
Unnecessary defensive programming.

src/tensorlake/applications/internal_logger.py
```diff
@@ -135,9 +147,16 @@ def _format_message(self, level: str, message: str, **kwargs) -> str:
         context["event"] = message
 
         if "exc_info" in context:
-            context["exception"] = "".join(
-                traceback.format_exception(context["exc_info"])
-            )
+            exc_info: tuple | None = context["exc_info"]
```
eabatalov: `exc_info: tuple[.., .., ..] | BaseException  | bool | None = context.get("exc_info")`

tests/applications/test_structlogger.py
```diff
@@ -6,12 +6,12 @@
 from pydantic import BaseModel
 
 from tensorlake.applications import (
+    Logger,
```
eabatalov, line 9: It seems like .bind() call is not covered. May I ask you to add a test case for it? This is mainly because we're not using Logger in our code so we don't cover it in end-to-end tests and because .bind implementation in Logger.bind() is not trivial really - it might easily regress with future code changes without us noticing.

src/tensorlake/applications/internal_logger.py
calavera: updated

src/tensorlake/applications/internal_logger.py
calavera: updated

src/tensorlake/applications/internal_logger.py
calavera: updated

tests/applications/test_structlogger.py
calavera: updated

PR #504, opened 2026-01-13
feat: adds support for image builder api v3

daniel-white:
* adds v3 endpoint suppport
* new high level builder client
* replaces existing image build

### DIFF
--- a/src/tensorlake/applications/image_builder/__init__.py
+++ b/src/tensorlake/applications/image_builder/__init__.py
@@ -1,9 +1,4 @@
-"""Builder module for building container images and application versions in Tensorlake.
-
-This module provides classes and utilities for building container images that
-contain Tensorlake functions and their dependencies. It handles the orchestration
-of image builds, streaming build logs, and managing the build lifecycle.
-"""
+"""Builder module for building container images and application versions."""
 
 import asyncio
 import os
@@ -21,10 +16,8 @@ from tensorlake.applications.image import ImageInformation, create_image_context
 from tensorlake.applications.image_builder.client_v3 import (
     ApplicationVersionBuildInfoV3,
     ApplicationVersionBuildRequestV3,
-    ClientKey,
     ImageBuilderClientV3,
     ImageBuilderClientV3Error,
-    ImageBuildId,
     ImageBuildInfoV3,
     ImageBuildLogEventV3,
     ImageBuildRequestV3,
@@ -32,60 +25,16 @@ from tensorlake.applications.image_builder.client_v3 import (
 
 
 class ImageBuildRequest:
-    """Represents a request to build a container image for Tensorlake functions.
-
-    Tensorlake functions run in function containers. This class encapsulates
-    the information needed to build a container image, including the image
-    definition (base image, build operations) and the list of functions
-    that should be included in the build.
-
-    Container images are used to install dependencies (Python packages, system
-    packages) that functions need at runtime. Images are defined using the
-    `Image` class with methods like `.run()`, `.add()`, `.copy()`, and `.env()`.
-
-    For more information on defining images, see:
-    https://docs.tensorlake.ai/applications/images
-
-    Attributes:
-        image_info: The ImageInformation object containing the image and the
-            functions that use it.
-
-    Example:
-        >>> from tensorlake.applications import Image
-        >>> from tensorlake.applications.image import ImageInformation
-        >>> # Define an image with dependencies
-        >>> image = (
-        ...     Image(name="my-pdf-parser-image", base_image="ubuntu:24.04")
-        ...     .run("apt update")
-        ...     .run("pip install langchain")
-        ... )
-        >>> # Create image info with functions
-        >>> image_info = ImageInformation(image=image, functions=[...])
-        >>> # Create a build request
-        >>> request = ImageBuildRequest(image_info)
-    """
+    """Request to build a container image."""
 
     def __init__(self, image_info: ImageInformation):
-        """
-        Initialize an ImageBuildRequest.
-
-        Args:
-            image_info: The ImageInformation object containing the image and
-                the functions that use it.
-
-        Raises:
-            ValueError: If image_info has no functions.
-        """
+        """Initialize an ImageBuildRequest."""
         if not image_info.functions:
             raise ValueError("image_info.functions cannot be empty")
         self.image_info = image_info
 
     async def _synthesize_v3_request(self) -> ImageBuildRequestV3:
-        """Synthesize a v3 image build request from this image build request.
-
-        Returns:
-            The v3 image build request.
-        """
+        """Synthesize a v3 image build request."""
         image = self.image_info.image
         function_names = [func.function_name for func in self.image_info.functions]
 
@@ -110,64 +59,10 @@ class ImageBuildRequest:
 
 
 class BuildRequest:
-    """Represents a request to build an application with multiple container images.
-
-    This class is used to construct a build request for an application, which
-    may consist of multiple container images. Each image can contain different
-    sets of functions and their dependencies.
-
-    Tensorlake applications can use multiple images to separate functions with
-    different dependency requirements. For example, you might have one image
-    for PDF processing functions (with PDF libraries) and another for ML
-    functions (with ML frameworks).
-
-    Attributes:
-        name: The name of the application being built. This should match your
-            application name in Tensorlake.
-        version: The version string for this application build (e.g., "v1.0.0",
-            "latest", "dev-2024-01-01"). Used to identify and track different
-            builds of the same application.
-        images: List of ImageBuildRequest objects, each representing a container
-            image to be built as part of this application. Each image can have
-            different base images, dependencies, and associated functions.
-
-    Example:
-        >>> from tensorlake.applications import Image
-        >>> req = BuildRequest("my-app", "v1.0.0")
-        >>>
-        >>> # Create an image for PDF processing functions
-        >>> pdf_image = (
-        ...     Image(name="pdf-processor", base_image="ubuntu:24.04")
-        ...     .run("apt update && apt install -y poppler-utils")
-        ...     .run("pip install pdfplumber")
-        ... )
-        >>> req.add_image(pdf_image, ["parse_pdf", "extract_text"])
-        >>>
-        >>> # Create an image for ML functions
-        >>> ml_image = (
-        ...     Image(name="ml-processor")
-        ...     .run("pip install torch transformers")
-        ... )
-        >>> req.add_image(ml_image, ["classify_document", "extract_entities"])
-        >>>
-        >>> print(len(req.images))
-        2
-    """
+    """Request to build an application with multiple container images."""
 
     def __init__(self, name: str, version: str):
-        """Initialize a BuildRequest.
-
-        Args:
-            name: The name of the application. This should match the application
-                name in your Tensorlake project.
-            version: The version identifier for this application build.
-                This can be any string (e.g., "v1.0.0", "latest", "dev-2024-01-01").
-                The version is used to identify and track different builds of the
-                same application.
-
-        Raises:
-            ValueError: If name or version is empty or None.
-        """
+        """Initialize a BuildRequest."""
         if not name:
             raise ValueError("name cannot be empty or None")
         if not version:
@@ -177,60 +72,13 @@ class BuildRequest:
         self.images: list[ImageBuildRequest] = []
 
     def add_image(self, image_info: ImageInformation) -> ImageBuildRequest:
-        """Add an image build request to this application version build.
-
-        This method creates an ImageBuildRequest for the given image information,
-        adds it to the list of images to be built, and returns it for further
-        configuration if needed.
-
-        The image should be defined with all necessary build operations (e.g.,
-        `.run()` commands to install dependencies) before being added to the
-        build request.
-
-        Args:
-            image_info: The ImageInformation object containing the image definition
-                and the functions that use it. This should be an ImageInformation
-                instance with an Image that has base image and build operations
-                defined (e.g., `Image(name="my-image").run("pip install package")`).
-                The functions in the ImageInformation will be packaged into containers
-                built from the image and will have access to all dependencies installed
-                in the image.
-
-        Returns:
-            The created ImageBuildRequest object, which can be used for further
-            configuration or reference.
-
-        Raises:
-            ValueError: If image_info has no functions.
-
-        Example:
-            >>> from tensorlake.applications import Image
-            >>> from tensorlake.applications.image import ImageInformation
-            >>> req = BuildRequest("my-app", "v1.0.0")
-            >>>
-            >>> # Define image with dependencies
-            >>> image = (
-            ...     Image(name="worker", base_image="ubuntu:24.04")
-            ...     .run("apt update")
-            ...     .run("pip install langchain")
-            ... )
-            >>>
-            >>> # Create image info with functions
-            >>> image_info = ImageInformation(image=image, functions=[...])
-            >>> image_req = req.add_image(image_info)
-            >>> print(len(image_req.image_infos))
-            1
-        """
+        """Add an image build request to this application version build."""
         image_req = ImageBuildRequest(image_info)
         self.images.append(image_req)
         return image_req
 
     async def _synthesize_v3_request(self) -> ApplicationVersionBuildRequestV3:
-        """Synthesize a v3 application version build request from this request.
-
-        Returns:
-            The v3 application version build request.
-        """
+        """Synthesize a v3 application version build request."""
         return ApplicationVersionBuildRequestV3(
             name=self.name,
             version=self.version,
@@ -261,15 +109,7 @@ _IMAGE_NAME_PREFIX_COLORS: list[str] = [
 
 @dataclass
 class BuildSummary:
-    """Summary of image build results.
-
-    Attributes:
-        total: Total number of image builds.
-        succeeded: Number of successful builds.
-        failed: Number of failed builds.
-        canceled: Number of canceled builds.
-        unknown: Number of builds with unknown status.
-    """
+    """Summary of image build results."""
 
     total: int
     succeeded: int
@@ -282,11 +122,7 @@ class _ImageBuildReporter:
     _instance_count = 0
 
     def __init__(self, info: ImageBuildInfoV3):
-        """Initialize the image build reporter.
-
-        Args:
-            info: The image build information to report on.
-        """
+        """Initialize the image build reporter."""
         self._info = info
         self._last_seen_status = info.status
         self._display_name = info.name if info.name != "default" else info.id[:12]
@@ -300,28 +136,24 @@ class _ImageBuildReporter:
         self._status_message_task: asyncio.Task[None] | None = None
 
     @property
-    def key(self) -> ClientKey | None:
+    def key(self) -> str | None:
         """Get the image build request key."""
         return self._info.key
 
     @property
-    def image_build_id(self) -> ImageBuildId:
+    def image_build_id(self) -> str:
         """Get the image build ID."""
         return self._info.id
 
     @property
     def last_seen_status(self) -> str:
-        """Get the last seen build status from log events."""
+        """Get the last seen build status."""
         return self._last_seen_status
 
     async def process_log_events(
         self, stream: AsyncGenerator[ImageBuildLogEventV3, None]
     ):
-        """Process and display log events from the build stream.
-
-        Args:
-            stream: Async generator of image build log events.
-        """
+        """Process and display log events from the build stream."""
         if self._finished:
             return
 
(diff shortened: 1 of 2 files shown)


src/tensorlake/builder/client_v3.py
```diff
@@ -0,0 +1,396 @@
...
+
+    def __init__(self, req: ImageBuildRequest):
+        self.key = req.key
+        self.name = req.name
+        self.description = req.description
+        self.context_tar_part_name = uuid7().hex
+        self.function_names = req.function_names
+
+
+class _ApplicationVersionBuildRequest(BaseModel):
```
calavera: please, cleanup the model definitions, it doesn't make sense to have a model that uses dataclass definitions and the same model that extends from pydantic basemodels, we should only have one.

daniel-white: cleaned up, and renamed to make it more clear which is which

src/tensorlake/applications/interface/function.py
```diff
@@ -158,6 +158,17 @@ def awaitable(self) -> "FunctionAwaitablesFactory":
         """Returns function factory for creating awaitables."""
         return self._awaitables_factory
 
+    @property
+    def function_name(self) -> str:
```
diptanu: def name(self)

that would read something like <function>.name

src/tensorlake/applications/interface/function.py
```diff
@@ -170,12 +181,7 @@ def __repr__(self) -> str:
 
     def __str__(self) -> str:
         # Shows a simple human readable representation of the Function. Used in error messages.
-        function_name: str = (
```
eabatalov, line 173: Don't change this please. This is handling cases of inconsistent misconfigured by user function objects.
When app is deployed it's fully validated and doesn't have any inconsistencies.

src/tensorlake/applications/interface/function.py
eabatalov, line 162: This new getter is most probably not needed because we're already using this information provided by SDK.
I'll figure out how you use this and write an approach that aligns with current architecture.

daniel-white: It was getting it via the private path

src/tensorlake/cli/deploy.py
```diff
@@ -8,6 +9,14 @@
 from tensorlake.applications import Function, Image, SDKUsageError, TensorlakeError
 from tensorlake.applications.applications import filter_applications
 from tensorlake.applications.image import ImageInformation, image_infos
+from tensorlake.applications.image_builder import (
+    BuildRequest,
+    ImageBuilder,
+)
+from tensorlake.applications.image_builder.client_v3 import (
+    ImageBuilderClientV3,
+    ImageBuilderClientV3Options,
```
eabatalov, line 18: Why ImageBuilderClient and ImageBuilder are in separate packages? Aren't all of this is an ImageBuilderClient stuff?

src/tensorlake/cli/deploy.py
```diff
@@ -50,25 +57,31 @@
...
-        build_service=os.getenv("TENSORLAKE_BUILD_SERVICE")
-        or f"{auth.api_url}/images/v2",
-        api_key=bearer_token,
-        organization_id=auth.organization_id if not auth.api_key else None,
-        project_id=auth.project_id if not auth.api_key else None,
+    click.echo(
+        f"⚙️  Preparing deployment for application(s) from {application_file_path}"
     )
 
+    opts = ImageBuilderClientV3Options.from_env()
```
eabatalov, line 67: Please add a type hint to opts.

src/tensorlake/cli/deploy.py
daniel-white: Separation of concerns.

src/tensorlake/cli/deploy.py
```diff
@@ -102,40 +115,46 @@ def deploy(
...
+                name=fn_config.function_name,
+                version=app_config.version,
+            )
+
+            for image_info in images.values():
+                build_req.add_image(image_info)
+
+            if not build_req.images:
+                click.secho(
+                    f"❌ No images found for application '{fn_config.function_name}'. "
```
eabatalov, line 135: This is already validated during application validation. No need to do this check again.

src/tensorlake/cli/deploy.py
```diff
@@ -102,40 +115,46 @@ def deploy(
...
-                    raise error
-                except Exception as error:
-                    click.echo(error, err=True)
-                    raise click.Abort
-
-    click.secho("\n✅ All images built successfully")
+                raise click.Abort
+
+            await builder.build(build_req)
+        except (
```
eabatalov, line 143: If we're rewriting ImageBuilder then it should provide a clear contract on what exception types it rases in which cases. We also shouldn't use click inside SDK where the new ImageBuilder was moved now as SDK doesn't depend on Click, only CLI.

src/tensorlake/cli/deploy.py
```diff
@@ -102,40 +115,46 @@ def deploy(
...
+            click.Abort,
+            click.UsageError,
+        ) as error:
+            # Re-raise cancellation exceptions so they can be handled at asyncio.run() level
+            # This avoids calling sys.exit() from within async function which causes cleanup issues
+            raise error
+        except Exception:
+            # Error message and summary are already printed by builder.build()
+            # Print final error message
+            click.secho("\n❌ Image build(s) failed", err=True, fg="red")
```
eabatalov, line 155: We swallow the Exception here (not printing it) how we and user are going to debug such errors?

src/tensorlake/applications/image_builder/__init__.py
```diff
@@ -0,0 +1,826 @@
...
+
+import asyncio
+import os
+import sys
+import tempfile
+from datetime import datetime
+from typing import AsyncGenerator
+from uuid import uuid4 as uuid
+
+import click
```
eabatalov, line 12: Click is a CLI only dependency, please don't add it into SDK.

src/tensorlake/applications/image_builder/__init__.py
```diff
@@ -0,0 +1,826 @@
...
+of image builds, streaming build logs, and managing the build lifecycle.
+"""
+
+import asyncio
+import os
+import sys
+import tempfile
+from datetime import datetime
+from typing import AsyncGenerator
+from uuid import uuid4 as uuid
```
eabatalov: We typically use nanoid (vendored module) but if you want we can use something else.

src/tensorlake/applications/image_builder/__init__.py
```diff
@@ -0,0 +1,826 @@
...
+            key=uuid().hex,
+            name=image.name,
+            description=None,
+            context_tar_content=context_tar_content,
+            function_names=function_names,
+        )
+
+
+class BuildRequest:
+    """Represents a request to build an application with multiple container images.
```
eabatalov: Do we need such verbose docstrings? Asking because these are internal non-public components.

src/tensorlake/applications/image_builder/client_v3.py
```diff
@@ -0,0 +1,1122 @@
+"""
```
eabatalov: Please reduce verbosity of docstrings and make them concise and clear.

src/tensorlake/applications/image_builder/__init__.py
```diff
@@ -0,0 +1,826 @@
...
+            >>> # Build the application images (logs streamed to console)
+            >>> try:
+            ...     await builder.build(req)
+            ...     print("Build completed successfully")
+            ... except ImageBuilderClientV3Error as e:
+            ...     print(f"Build error: {e}")
+        """
+
+        try:
+            v3_req = (
```
eabatalov, line 300: Please add type hints to all local variables to keep the code base easy to read and enforce discipline and alignment with existing coding practices in the repo.

src/tensorlake/applications/image_builder/__init__.py
```diff
@@ -0,0 +1,826 @@
...
+                await req._synthesize_v3_request()  # pylint: disable=protected-access
+            )
+            info = await self._client.build_app(v3_req)
+            reporters = {
+                image_build_info.id: _ImageBuildReporter(image_build_info)
+                for image_build_info in info.image_builds.values()
+            }
+        except ImageBuilderClientV3Error as e:
+            click.secho(str(e), err=True, fg="red")
+            raise
```
eabatalov, line 310: All these exceptions need to bubble to CLI and then logged their during click.echo.
ImageBuilder needs to document all exceptions it can raise with enough details to make exception handling by CLI meaningful. If CLI is not going to do different things depending on the exception type then just document "raises  Exception on error".

src/tensorlake/applications/image_builder/__init__.py
```diff
@@ -0,0 +1,826 @@
...
+            click.secho("🏭 Image build logs:", bold=True)
+            process_log_events_tasks = [
+                asyncio.create_task(reporter.process_log_events(log_stream))
+                for reporter, log_stream in zip(
+                    reporters.values(), log_streams, strict=True
+                )
+            ]
+            _ = await asyncio.gather(*process_log_events_tasks, return_exceptions=True)
+
+        except (asyncio.CancelledError, KeyboardInterrupt, click.Abort):
```
eabatalov, line 343: Should we remove except for these exceptions from cli code?

src/tensorlake/applications/image_builder/__init__.py
```diff
@@ -0,0 +1,826 @@
...
+            ]
+            click.echo()
+            click.secho("🏭 Image build logs:", bold=True)
+            process_log_events_tasks = [
+                asyncio.create_task(reporter.process_log_events(log_stream))
+                for reporter, log_stream in zip(
+                    reporters.values(), log_streams, strict=True
+                )
+            ]
+            _ = await asyncio.gather(*process_log_events_tasks, return_exceptions=True)
```
eabatalov, line 341: return_exceptions=True means that all exceptions raised are put into the returned list which is ignored here.
Is this correct?

src/tensorlake/applications/image_builder/__init__.py
```diff
@@ -0,0 +1,826 @@
...
+                else:
+                    info = await self._client.image_build_info(reporter.image_build_id)
+                reporter.print_final_result(info)
+                status = info.status if info is not None else reporter.last_seen_status
+                self._update_summary_from_status(summary, status)
+            except ImageBuilderClientV3Error as e:
+                self._handle_build_info_error(
+                    reporter,
+                    summary,
+                    f"Error getting final build info for {reporter.image_build_id}: {e}",
```
eabatalov, line 452: Do we need to write all these separate except blocks to add these "Error getting final build info" lines while a full backtrace and exception is available to us?

src/tensorlake/applications/image_builder/__init__.py
```diff
@@ -0,0 +1,826 @@
...
+        if summary["failed"] > 0:
+            raise RuntimeError("Image build(s) failed")
+
+    def _print_build_summary(self, summary: dict[str, int]):
+        """Print a compiler-style summary of build results.
+
+        Args:
+            summary: Dictionary with counts for total, succeeded, failed, canceled, and unknown.
+        """
+        total = summary["total"]
```
eabatalov: Please use a @dataclass or anything similar instead of untyped dicts.
Also please don't print anything except one success message for all image builds if everything succeeded.
We need to reduce cognitive load on users, not increase it by printing more build stats than they need to see.
I think all these _print_ methods can be removed unless you have a clear reason why to keep them.
Also we need to output a single final success/failure message for all applications. Thus this printing needs to be done in CLI.

src/tensorlake/applications/image_builder/__init__.py
```diff
@@ -0,0 +1,826 @@
...
+        Args:
+            image_info: The ImageInformation object containing the image and
+                the functions that use it.
+
+        Raises:
+            ValueError: If image_info has no functions.
+        """
+        if not image_info.functions:
+            raise ValueError("image_info.functions cannot be empty")
+        self.image_info = image_info
```
eabatalov, line 34: If the whole purpose of this object is to wrap ImageInformation and provide a method that transforms it, why don't just replace it with private function `def _v3_request(image_info)` ? This would remove > 100 lines of docstrings and not useful code. Also ImageBuildRequest and BuildRequest are hard to understand entities. It sounds like the same things. If we keep ImageBuildRequest class then it needs to be clear what it is just from its name.

src/tensorlake/applications/image_builder/__init__.py
```diff
@@ -0,0 +1,826 @@
...
+
+
+class BuildRequest:
+    """Represents a request to build an application with multiple container images.
+
+    This class is used to construct a build request for an application, which
+    may consist of multiple container images. Each image can contain different
+    sets of functions and their dependencies.
+
+    Tensorlake applications can use multiple images to separate functions with
```
eabatalov: We don't need to explain Tensorlake Applications UX here as this class if for internal use and we have good public docs that are responsible for explaining all of this.

src/tensorlake/applications/image_builder/__init__.py
```diff
@@ -0,0 +1,826 @@
...
+        self.image_info = image_info
+
+    async def _synthesize_v3_request(self) -> ImageBuildRequestV3:
+        """Synthesize a v3 image build request from this image build request.
+
+        Returns:
+            The v3 image build request.
+        """
+        image = self.image_info.image
+        function_names = [func.function_name for func in self.image_info.functions]
```
eabatalov, line 39: Please use `func._function_config.function_name`. It's okay because this code is part of SDK.
This is consistent with how function name is obtained currently all across SDK for properly initialized function objects. Also .function_name public property shouldn't be added to Function object because it makes it usable by users because they work with Function objects directly. We minimize API surface exposed to users to minimal to ease maintenance.

src/tensorlake/cli/deploy.py
```diff
@@ -26,15 +35,13 @@
     print_validation_messages,
     validate_loaded_applications,
 )
-from tensorlake.builder.client_v2 import BuildContext, ImageBuilderV2Client
```
eabatalov, line 29: Please delete all clients not used anymore in this PR.

src/tensorlake/applications/image_builder/__init__.py
```diff
@@ -0,0 +1,826 @@
...
+            canceled_app_version_info = await self._cancel_builds(
+                info, process_log_events_tasks
+            )
+            summary = await self._print_final_results_for_reporters(
+                reporters, canceled_app_version_info
+            )
+            self._print_build_summary(summary)
+            # Use os._exit() to bypass asyncio cleanup and avoid "unhandled exception" errors
+            # This exits immediately without triggering asyncio.run() cleanup issues
+            os._exit(0)
```
eabatalov, line 354: This looks like a hack, please fix this properly or explain why it's not possible at the moment.

src/tensorlake/applications/image_builder/__init__.py
```diff
@@ -0,0 +1,826 @@
...
+                    f"Error getting final build info for {reporter.image_build_id}: {e}",
+                )
+            except RuntimeError as e:
+                self._handle_build_info_error(
+                    reporter,
+                    summary,
+                    f"Build service error getting final build info for {reporter.image_build_id}: {e}",
+                )
+            except Exception as e:  # pylint: disable=broad-except
+                self._handle_build_info_error(
```
eabatalov, line 461: Why do we have many except blocks that do almost the same?
Also why do we call _handle_build_info_error instead of bubbling the exception up to the caller?

src/tensorlake/applications/image_builder/client_v3.py
```diff
@@ -0,0 +1,1122 @@
...
+from uuid import uuid4 as uuid
+
+import click
+import httpx
+from httpx_sse import aconnect_sse
+from pydantic import BaseModel
+
+from tensorlake.cli._common import ASYNC_HTTP_EVENT_HOOKS
+
+# Enable httpx debug logging if requested via environment variable
```
eabatalov, line 17: httpx is used across SDK and CLI.
What side effects does this have on them?
Is it right to put this code to client_v3.py?
Is this approach documented in httpx docs?

src/tensorlake/applications/image_builder/client_v3.py
```diff
@@ -0,0 +1,1122 @@
...
+    @property
+    def project_id(self) -> str | None:
+        """The project ID. Returns None if using API key authentication."""
+        # Hide project_id when using API key (not needed, API key contains this info)
+        if self._api_key is not None:
+            return None
+        return self._project_id
+
+    @classmethod
+    def from_env(cls) -> "ImageBuilderClientV3Options":
```
eabatalov, line 196: This looks to me like we're copy-pasting CLI code here.
Why do we do that? Who's going to use from_env() ?

src/tensorlake/applications/image_builder/client_v3.py
```diff
@@ -0,0 +1,1122 @@
...
+        - API Key: api_key is provided, pat is None. organization_id and project_id return None.
+          API keys already contain org/project info via introspection.
+        - PAT: pat is provided, api_key is None. organization_id and project_id are available.
+          PAT tokens don't contain org/project info, so X-Forwarded headers are needed.
+
+    Note:
+        Validation is not performed in __init__. Call validate() to validate the configuration.
+    """
+
+    _base_url: str = field(init=False)
```
eabatalov, line 59: If this is a simple @dataclass then why do we make these fields private and add lots of getter properties for it? Can we remove the getter properties?

src/tensorlake/applications/image_builder/client_v3.py
```diff
@@ -0,0 +1,1122 @@
...
+    _api_key: str | None = field(default=None, init=False)
+    _pat: str | None = field(default=None, init=False)
+    _organization_id: str | None = field(default=None, init=False)
+    _project_id: str | None = field(default=None, init=False)
```
eabatalov, line 63: What's the use for `field` here?

src/tensorlake/applications/image_builder/client_v3.py
```diff
@@ -0,0 +1,1122 @@
...
+# Sentinel object to distinguish between "not provided" and "explicitly None"
+class _NotProvided:
+    pass
+
+
+_NOT_PROVIDED = _NotProvided()
+
+
+@dataclass
+class ImageBuilderClientV3Options:
```
eabatalov, line 56: Can we remove this and just create ImageBuilderClientV3() directly in CLI?
We're essentially creating an abstraction here over concepts that ImageBuilderClient shouldn't know about like pat,api_key,organization_id, project_id. This is all managed by CLI. Long story short there's nothing ImageBuilder specific in ImageBuilderClientV3Options.

Also please extract common logic with APIClient, see https://github.com/tensorlakeai/tensorlake/blob/main/src/tensorlake/applications/remote/api_client.py#L157C9-L157C16 and  https://github.com/tensorlakeai/tensorlake/blob/main/src/tensorlake/applications/remote/api_client.py#L490 and reuse it in both this client and APIClient. This will result in a more complete refactoring that you started already.

eabatalov: If you really want to keep ImageBuilderClientV3Options then it needs to be a common non-image builder specific options object accepted by both image builder and existing API clients. The options objects needs to be defined in a common package and filled by CLI which handles configuration of all these things.
Also if you use it then existing CLI code needs to use it too where applicable.

src/tensorlake/applications/image_builder/client_v3.py
```diff
@@ -0,0 +1,1122 @@
...
+            project_id=project_id,
+        )
+
+
+# ============================================================================
+# Type Aliases (for documentation purposes)
+# ============================================================================
+
+# Type aliases for clarity (these are just str at runtime)
+ImageBuildId = str
```
eabatalov: Why do we add these type aliases instead of using str directly?

src/tensorlake/applications/image_builder/__init__.py
daniel-white: switching

src/tensorlake/applications/image_builder/__init__.py
daniel-white: lets discuss tomorrow.

src/tensorlake/applications/image_builder/__init__.py
daniel-white: felt kind of yucky reaching into its internals...

src/tensorlake/applications/image_builder/__init__.py
daniel-white: yea i thought these _were_ public

src/tensorlake/applications/image_builder/__init__.py
daniel-white: anyway to force that with a tool?

src/tensorlake/applications/image_builder/__init__.py
daniel-white: yea. im thinking we need some kind of verbose mode, otherwise showing full stack traces all the time is annoying

src/tensorlake/applications/image_builder/__init__.py
daniel-white: just different error messages. happy to learn

PR #510, opened 2026-02-02
HTTP API in FE for communicating with dataplane 

diptanu:
Added a HTTP API in the FE so that the dataplane can talk to it instead of using grpc. this will help us remove grpc from the FE completely reducing a dependency. 

Another change -> Bind mount code in the container instead of an /initialize api since it's done only once in the life of an FE.

The changes are backward compatible to the python executor


src/tensorlake/function_executor/http_server.py
```diff
@@ -0,0 +1,361 @@
...
+
+import hashlib
+import json
+import re
+import threading
+import time
+from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
+from typing import TYPE_CHECKING, Any, Dict, Optional
+
+from google.protobuf.json_format import MessageToDict, ParseDict
```
eabatalov, line 15: This is a dependency of protobuf. Is this expected?

src/tensorlake/function_executor/allocation_runner/allocation_runner.py
```diff
@@ -207,10 +207,19 @@ def __init__(
         self._output_blob_requests: Dict[str, _OutputBLOBRequestInfo] = {}
 
     def wait_allocation_state_update(
-        self, last_seen_hash: str | None
+        self, last_seen_hash: str | None, timeout: float | None = None
```
eabatalov, line 210: Why this timeout is needed? It's not needed in FE as of now, this might hind a fundamental and not correct assumption change in this PR.

src/tensorlake/function_executor/allocation_runner/allocation_state_wrapper.py
```diff
@@ -132,17 +132,35 @@ def remove_request_state_operation(self, id: str) -> None:
...
+            timeout: Maximum time to wait in seconds. If None, blocks indefinitely.
+                     If 0 or negative, returns immediately with current state.
+
+        Returns:
+            Copy of the current allocation state.
+        """
         with self._allocation_state_update_lock:
+            # If timeout is 0 or negative, return immediately
+            if timeout is not None and timeout <= 0:
+                return self._copy_state_locked()
```
eabatalov, line 151: To read current value just pass last_seen_hash=None.
Let's not add any extra meaning to this new timeout arg.

src/tensorlake/function_executor/server.py
```diff
@@ -12,18 +16,58 @@
...
-        server.start()
-        server.wait_for_termination()
+        """Runs Function Executor Service.
+
+        Starts HTTP server if http_port is configured.
+        Starts gRPC server if server_address is configured.
+        At least one must be configured.
+        """
+        # Start HTTP server if configured
+        if self._http_port is not None and self._logger is not None:
```
eabatalov, line 42: `and self._logger is not None` looks like a hack

src/tensorlake/function_executor/service.py
```diff
@@ -95,92 +96,255 @@ def __init__(self, logger: InternalLogger):
...
+        self._logger.info(
+            "initialized function executor service",
+            duration_sec=f"{time.monotonic() - start_time:.3f}",
+        )
+        log_user_event_initialization_finished(event_details)
+        return InitializeResponse(
+            outcome_code=InitializationOutcomeCode.INITIALIZATION_OUTCOME_CODE_SUCCESS,
+        )
+
+    def initialize_from_code_path(
```
eabatalov, line 148: Please share all of this logic with existing code paths, not copy-pasting.

src/tensorlake/function_executor/service.py
```diff
@@ -95,92 +96,255 @@ def __init__(self, logger: InternalLogger):
...
+            namespace=function_ref.namespace,
+            application_name=function_ref.application_name,
+            application_version=function_ref.application_version,
+            function_name=function_ref.function_name,
         )
         log_user_event_initialization_started(event_details)
 
-        self._function_ref = request.function
+        # Set up function ref and logger
+        self._setup_function_ref_and_logger(function_ref)
```
eabatalov, line 115: Unhandled exception.

src/tensorlake/function_executor/service.py
```diff
@@ -95,92 +96,255 @@ def __init__(self, logger: InternalLogger):
...
+        # Complete initialization
+        self._complete_initialization()
+        self._logger.info(
+            "initialized function executor service from code path",
+            duration_sec=f"{time.monotonic() - start_time:.3f}",
+        )
+        log_user_event_initialization_finished(event_details)
+        return True
+
+    def _setup_function_ref_and_logger(self, function_ref: FunctionRef) -> None:
```
eabatalov, line 212: This function does lots of things, doesn't look like a good design choice.
Also it's not clear why it changes so much and i.e. loses exception handling code here.

tests/function_executor/test_http_endpoints.py
```diff
@@ -0,0 +1,642 @@
+"""Tests for Function Executor HTTP endpoints.
+
+These tests verify that the HTTP endpoints work correctly alongside the gRPC service.
+Uses stdlib http.client for HTTP requests (zero external dependencies).
```
eabatalov, line 4: We can use any dev dependencies in tests, this won't affect the built package.

eabatalov (changes requested):
Overall comments:

* This PR touches only simple FE initialization and CRUD allocations API. The actual complexity is in running allocations which will require significant extra work both in FE but also a lot of work in "dataplane" (ex Executor service). Without these changes working end to end with the current quality we have in prod this PR is a dead code that adds complexity and maintenance burden.
* This PR needs more refactoring to deduplicate all the code and minimize changes in existing code paths because these changes introduce bugs.

src/tensorlake/function_executor/http_server.py
```diff
@@ -0,0 +1,361 @@
...
+        self._shutdown_event.set()
+        if self._server:
+            self._server.shutdown()
+
+    def _create_handler(self) -> type:
+        """Create a request handler class with access to the service."""
+        service = self._service
+        logger = self._logger
+
+        class Handler(BaseHTTPRequestHandler):
```
eabatalov, line 70: This approach is not great as it puts all the web framework logic including routing and response rendering/parsing into a single class. I already did splitting of functioncality in request context http server. Please either reuse it or implement separation of concerns here to not end up with one giant class handling everything.

PR #519, opened 2026-02-09
sandbox client

diptanu:
Developer Experience

  1. One-liner: create, use, auto-cleanup
```
  from tensorlake.sandbox import SandboxClient

  client = SandboxClient()  # reads TENSORLAKE_API_URL + TENSORLAKE_API_KEY from env

  with client.create_and_connect(image="python:3.12-slim") as sandbox:
      result = sandbox.run("python", args=["-c", "print('hello')"])
      print(result.stdout)   # "hello"
      print(result.exit_code) # 0
  # sandbox auto-terminates on exit
```
  2. Step-by-step: create, poll, connect, use, delete

```
  from tensorlake.sandbox import SandboxClient, SandboxStatus

  client = SandboxClient(api_url="http://localhost:8900")

  # Create
  resp = client.create(image="alpine:latest", cpus=0.5, memory_mb=256)
  print(resp.sandbox_id)  # "sbx-abc123"
  print(resp.status)      # SandboxStatus.PENDING

  # Poll until running
  info = client.get(resp.sandbox_id)
  while info.status == SandboxStatus.PENDING:
      info = client.get(resp.sandbox_id)

  # Connect and interact
  sandbox = client.connect(resp.sandbox_id)
  sandbox.write_file("/tmp/data.txt", b"hello world")
  content = sandbox.read_file("/tmp/data.txt")  # b"hello world"

  proc = sandbox.start_process("cat", args=["/tmp/data.txt"])
  stdout = sandbox.get_stdout(proc.pid)
  print(stdout.lines)  # ["hello world"]

  # Cleanup
  sandbox.close()
  client.delete(resp.sandbox_id)
```
  3. Pools: pre-warm containers for fast startup
```
  # Create a pool with 3 warm containers
  pool = client.create_pool(
      image="python:3.12-slim",
      cpus=1.0,
      memory_mb=512,
      warm_containers=3,
  )

  # Sandboxes from the pool start instantly (no cold-start)
  with client.create_and_connect(pool_id=pool.pool_id) as sandbox:
      result = sandbox.run("python", args=["-c", "import sys; print(sys.version)"])

  # Update pool config
  client.update_pool(pool.pool_id, image="python:3.12-slim", warm_containers=5)

  # List all pools
  for p in client.list_pools():
      print(f"{p.pool_id}: {p.image} ({p.warm_containers} warm)")

  # Delete pool (fails with PoolInUseError if sandboxes are active)
  client.delete_pool(pool.pool_id)
```
  4. Error handling
```
  from tensorlake.sandbox import (
      SandboxClient,
      SandboxNotFoundError,
      PoolNotFoundError,
      PoolInUseError,
      RemoteAPIError,
  )

  client = SandboxClient()

  try:
      client.get("nonexistent-id")
  except SandboxNotFoundError as e:
      print(e.sandbox_id)  # "nonexistent-id"

  try:
      client.delete_pool("pool-with-active-sandboxes")
  except PoolInUseError as e:
      print(e.pool_id)     # "pool-with-active-sandboxes"

  try:
      client.delete_pool("nonexistent-pool")
  except PoolNotFoundError as e:
      print(e.pool_id)     # "nonexistent-pool"
```
  5. Process I/O streaming (interactive)
```
  sandbox = client.connect("sbx-abc123")

  # Start an interactive process with stdin pipe
  proc = sandbox.start_process("python3", stdin_mode="pipe")

  # Write to stdin
  sandbox.write_stdin(proc.pid, b"print('from stdin')\n")
  sandbox.close_stdin(proc.pid)

  # Stream output as it arrives (SSE)
  for event in sandbox.follow_stdout(proc.pid):
      print(f"[{event.timestamp}] {event.line}")

  6. File operations

  sandbox.write_file("/app/config.json", b'{"key": "value"}')
  data = sandbox.read_file("/app/config.json")

  entries = sandbox.list_directory("/app")
  for entry in entries.entries:
      kind = "dir" if entry.is_dir else "file"
      print(f"  {kind}: {entry.name} ({entry.size} bytes)")

  sandbox.delete_file("/app/config.json")
```

### DIFF
--- a/tests/sandbox/test_lifecycle.py
+++ b/tests/sandbox/test_lifecycle.py
@@ -537,7 +537,9 @@ class TestMaxContainers(BaseSandboxTest):
         # Wait for the reconciler to create a replacement warm container
         # before creating sandbox 2, so it can claim the warm rather than
         # needing an on-demand container.
-        _poll_pool_containers(self.client, self.__class__.pool_id, min_count=2, timeout=60)
+        _poll_pool_containers(
+            self.client, self.__class__.pool_id, min_count=2, timeout=60
+        )
 
         # Sandbox 2 claims the replacement warm container.
         resp2 = self.client.claim(self.__class__.pool_id)


tests/sandbox/testing.py
```diff
@@ -0,0 +1,156 @@
...
+import subprocess
+import sys
+import tempfile
+import time
+
+
+def find_free_port() -> int:
+    """Find an available TCP port on localhost."""
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+        s.bind(("", 0))
```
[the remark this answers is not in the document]
diptanu: This is test infrastructure - doesn't matter.

src/tensorlake/sandbox/client.py
```diff
@@ -0,0 +1,544 @@
...
+
+class SandboxClient:
+    """Client for managing Tensorlake sandboxes and sandbox pools."""
+
+    def __init__(
+        self,
+        api_url: str = _API_URL_FROM_ENV,
+        api_key: str | None = _API_KEY_FROM_ENV,
+        organization_id: str | None = None,
+        project_id: str | None = None,
```
calavera, line 46: why do we have a `project_id` and a `namespace`? isn't it the same thing?

src/tensorlake/sandbox/client.py
```diff
@@ -0,0 +1,544 @@
...
+                "cpus": cpus,
+                "memory_mb": memory_mb,
+                "ephemeral_disk_mb": ephemeral_disk_mb,
+            },
+            "timeout_secs": timeout_secs,
+        }
+
+        if secret_names:
+            payload["secret_names"] = secret_names
+        if entrypoint:
```
calavera: should we check if the list is empty?

src/tensorlake/sandbox/client.py
```diff
@@ -0,0 +1,544 @@
...
+            "resources": {
+                "cpus": cpus,
+                "memory_mb": memory_mb,
+                "ephemeral_disk_mb": ephemeral_disk_mb,
+            },
+            "timeout_secs": timeout_secs,
+        }
+
+        if secret_names:
+            payload["secret_names"] = secret_names
```
calavera: should we check if the list is empty?

src/tensorlake/sandbox/client.py
```diff
@@ -0,0 +1,544 @@
...
+            "image": image,
+            "resources": {
+                "cpus": cpus,
+                "memory_mb": memory_mb,
+                "ephemeral_disk_mb": ephemeral_disk_mb,
+            },
+            "timeout_secs": timeout_secs,
+        }
+
+        if secret_names:
```
calavera: should we check if the list if empty?

src/tensorlake/sandbox/client.py
```diff
@@ -0,0 +1,544 @@
...
+            payload["secret_names"] = secret_names
+        if entrypoint:
```
calavera: should we check if the list if empty?

.github/workflows/tests.yaml
```diff
@@ -54,6 +54,14 @@ jobs:
         id: login-ecr
         uses: aws-actions/amazon-ecr-login@v2
 
+      - name: Install Indexify Dataplane
```
eabatalov: Please add separate test GH actions for Sandbox. There's no need to merge testing of these two different products.

src/tensorlake/sandbox/client.py
```diff
@@ -0,0 +1,544 @@
...
+    ListSandboxesResponse,
+    ListSandboxPoolsResponse,
+    SandboxInfo,
+    SandboxPoolInfo,
+    SandboxStatus,
+)
+
+_DEFAULT_HTTP_REQUEST_TIMEOUT_SEC = 30.0
+
+_API_URL_FROM_ENV: str = os.getenv("TENSORLAKE_API_URL", "https://api.tensorlake.ai")
```
eabatalov: Nit. This is all a copy-paste. Can we dedup this?

src/tensorlake/sandbox/client.py
```diff
@@ -0,0 +1,544 @@
...
+    CreateSandboxPoolResponse,
+    CreateSandboxResponse,
+    ListSandboxesResponse,
+    ListSandboxPoolsResponse,
+    SandboxInfo,
+    SandboxPoolInfo,
+    SandboxStatus,
+)
+
+_DEFAULT_HTTP_REQUEST_TIMEOUT_SEC = 30.0
```
eabatalov: Why the default timeout is so high? Please explain a comment for future readers.

src/tensorlake/sandbox/client.py
eabatalov: +1, if users are going to use this constructor directly then we need the one for Cloud and one for OpenSource (as class methods).

src/tensorlake/sandbox/client.py
```diff
@@ -0,0 +1,544 @@
...
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        """Context manager exit."""
+        self.close()
+
+    def close(self):
+        """Close the HTTP client."""
+        self._client.close()
+
+    def _is_localhost(self) -> bool:
```
eabatalov, line 118: Is this some sort of a hack?

src/tensorlake/sandbox/client.py
```diff
@@ -0,0 +1,544 @@
...
+
+    def _run_request(self, request: httpx.Request) -> httpx.Response:
+        self._add_auth_headers(request.headers)
+        try:
+            response = self._client.send(request)
+            response.raise_for_status()
+        except httpx.HTTPStatusError:
+            raise
+        except httpx.RequestError as e:
+            raise RemoteAPIError(0, str(e))
```
eabatalov: HTTP status 0?

src/tensorlake/sandbox/client.py
```diff
@@ -0,0 +1,544 @@
...
+        if self._project_id is not None:
+            headers["X-Forwarded-Project-Id"] = self._project_id
+
+    def _run_request(self, request: httpx.Request) -> httpx.Response:
+        self._add_auth_headers(request.headers)
+        try:
+            response = self._client.send(request)
+            response.raise_for_status()
+        except httpx.HTTPStatusError:
+            raise
```
eabatalov: Redundant except block

src/tensorlake/sandbox/client.py
```diff
@@ -0,0 +1,544 @@
...
+            headers["X-Forwarded-Organization-Id"] = self._organization_id
+        if self._project_id is not None:
+            headers["X-Forwarded-Project-Id"] = self._project_id
+
+    def _run_request(self, request: httpx.Request) -> httpx.Response:
+        self._add_auth_headers(request.headers)
+        try:
+            response = self._client.send(request)
+            response.raise_for_status()
+        except httpx.HTTPStatusError:
```
eabatalov: Is this list of exception exhaustive?

src/tensorlake/sandbox/client.py
```diff
@@ -0,0 +1,544 @@
...
+
+    def _add_auth_headers(self, headers: dict[str, str]) -> None:
+        if self._api_key is not None:
+            headers["Authorization"] = f"Bearer {self._api_key}"
+        if self._organization_id is not None:
+            headers["X-Forwarded-Organization-Id"] = self._organization_id
+        if self._project_id is not None:
+            headers["X-Forwarded-Project-Id"] = self._project_id
+
+    def _run_request(self, request: httpx.Request) -> httpx.Response:
```
eabatalov, line 141: Please document all the exceptions raised, what caller needs to expect? I see callers already wrapping this method with try-except blocks? Do we need try-except here?

eabatalov, line 141: What about retries? When Server gets deployed, all requests will be failing without retries.

src/tensorlake/sandbox/client.py
```diff
@@ -0,0 +1,544 @@
...
+            raise RemoteAPIError(0, str(e))
+        return response
+
+    def create(
+        self,
+        image: Optional[str] = None,
+        cpus: float = 1.0,
+        memory_mb: int = 512,
+        ephemeral_disk_mb: int = 1024,
+        secret_names: Optional[List[str]] = None,
```
eabatalov: Use more modern `| None` instead of Optional.
Also you can use `list[` instead of `List[`

src/tensorlake/sandbox/client.py
```diff
@@ -0,0 +1,544 @@
...
+            response.raise_for_status()
+        except httpx.HTTPStatusError:
+            raise
+        except httpx.RequestError as e:
+            raise RemoteAPIError(0, str(e))
+        return response
+
+    def create(
+        self,
+        image: Optional[str] = None,
```
eabatalov: Can we share and re-use Image object from Applications SDK (this needs to be a separate package like tensorlake.containers.Image). I'd also prefer if we had tensorlake.container.Container. We might use it later when we add local function calls by default.

src/tensorlake/sandbox/client.py
```diff
@@ -0,0 +1,544 @@
...
+        cpus: float = 1.0,
+        memory_mb: int = 512,
+        ephemeral_disk_mb: int = 1024,
+        secret_names: Optional[List[str]] = None,
+        timeout_secs: Optional[int] = None,
+        entrypoint: Optional[List[str]] = None,
+        allow_internet_access: bool = True,
+        allow_out: Optional[List[str]] = None,
```
eabatalov: Can we use a more clear data model class for all these rules?
If not then let's clearly document the string format (ideally some pre-existing format instead of adding ours).

src/tensorlake/sandbox/client.py
```diff
@@ -0,0 +1,544 @@
...
+        """
+        try:
+            response = self._run_request(
+                self._client.build_request(
+                    "GET",
+                    url=self._endpoint_url(f"sandboxes/{sandbox_id}"),
+                )
+            )
+            return SandboxInfo(**response.json())
+        except httpx.HTTPStatusError as e:
```
eabatalov, line 302: The list of exceptions needs to be exhaustive in all cases, i.e. raise InternalError if an unexpected Exception is raised by the client. The SDK interface needs to work according to its docstring without raising unexpected exception types.

src/tensorlake/sandbox/client.py
```diff
@@ -0,0 +1,544 @@
...
+                )
+            )
+        except httpx.HTTPStatusError as e:
+            if e.response.status_code == 404:
+                raise SandboxNotFoundError(sandbox_id)
+            raise RemoteAPIError(e.response.status_code, e.response.text)
+
+    def create_pool(
+        self,
+        image: str,
```
eabatalov, line 354: Same questions about using a more robust data model instead of a long list of parameters with basic data types.

src/tensorlake/sandbox/client.py
```diff
@@ -0,0 +1,544 @@
...
+            max_containers: Maximum number of containers in pool
+            warm_containers: Number of warm containers to maintain
+
+        Returns:
+            CreateSandboxPoolResponse with pool_id and namespace
+
+        Raises:
+            RemoteAPIError: If the API request fails
+        """
+        payload = {
```
eabatalov: Why we're making dicts instead of using model classes for request payloads?

src/tensorlake/sandbox/client.py
eabatalov: David means:

```
if secret_names is not None:
```

We need to use `is None` is not None` everywhere because it's safer and shows intent better

src/tensorlake/sandbox/client.py
```diff
@@ -0,0 +1,544 @@
...
+            RemoteAPIError: If the API request fails
+        """
+        try:
+            response = self._run_request(
+                self._client.build_request(
+                    "GET",
+                    url=self._endpoint_url("sandbox-pools"),
+                )
+            )
+            data = ListSandboxPoolsResponse(**response.json())
```
eabatalov: This is not the right way of instantiating Model classes because this will fail if response json contains an unknown yet field for the ModelClass definition. Use validate json.

src/tensorlake/sandbox/sandbox.py
```diff
@@ -0,0 +1,368 @@
...
+    ListDirectoryResponse,
+    ListProcessesResponse,
+    OutputEvent,
+    OutputResponse,
+    ProcessInfo,
+    ProcessStatus,
+    SendSignalResponse,
+)
+
+if TYPE_CHECKING:
```
eabatalov, line 32: Unnecessary premature optimization imo

src/tensorlake/sandbox/sandbox.py
```diff
@@ -0,0 +1,368 @@
...
+
+        parsed = urlparse(proxy_url)
+        if parsed.hostname in ("localhost", "127.0.0.1"):
+            base_url = proxy_url.rstrip("/")
+            headers["Host"] = f"{sandbox_id}.local"
+        else:
+            port_part = f":{parsed.port}" if parsed.port else ""
+            base_url = f"{parsed.scheme}://{sandbox_id}.{parsed.hostname}{port_part}"
+
+        self._client: httpx.Client = httpx.Client(
```
eabatalov, line 78: Why do we create a new client instead of reusing existing one? Besides code design considerations httpx.client does http connection pooling which might be really handy.

src/tensorlake/sandbox/sandbox.py
```diff
@@ -0,0 +1,368 @@
...
+            base_url=base_url,
+            headers=headers,
+            timeout=30.0,
```
eabatalov: Why 30 sec? This needs to be explained in comments.

src/tensorlake/sandbox/sandbox.py
```diff
@@ -0,0 +1,368 @@
...
+
+    @property
+    def sandbox_id(self) -> str:
+        return self._sandbox_id
+
+    def __enter__(self) -> Sandbox:
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        if self._lifecycle_client is not None:
```
eabatalov: Is there any more clear logic we can use? Like just a bool flag with a clear name?

src/tensorlake/sandbox/sandbox.py
```diff
@@ -0,0 +1,368 @@
...
+        else:
+            self.close()
+
+    def close(self):
+        """Close the HTTP connection. The sandbox keeps running."""
+        self._client.close()
+
+    def terminate(self):
+        """Terminate the sandbox and close the connection."""
+        sandbox_id = self._sandbox_id
```
eabatalov: Why do we copy sandbox_id into a local var?

src/tensorlake/sandbox/sandbox.py
```diff
@@ -0,0 +1,368 @@
...
+        )
+
+    @property
+    def sandbox_id(self) -> str:
+        return self._sandbox_id
+
+    def __enter__(self) -> Sandbox:
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
```
eabatalov, line 93: How use would understand what behavior this __exit__ method would take (terminate vs client close)?

src/tensorlake/sandbox/sandbox.py
```diff
@@ -0,0 +1,368 @@
...
+
+    # --- Process management ---
+
+    def start_process(
+        self,
+        command: str,
+        args: Optional[List[str]] = None,
+        env: Optional[Dict[str, str]] = None,
+        working_dir: Optional[str] = None,
+        stdin_mode: str = "closed",
```
eabatalov: There are supposed to be enums for all string or any other constants.

src/tensorlake/sandbox/exceptions.py
```diff
@@ -0,0 +1,49 @@
...
+
+class PoolNotFoundError(SandboxError):
+    """Raised when a sandbox pool is not found."""
+
+    def __init__(self, pool_id: str):
+        self.pool_id = pool_id
+        super().__init__(f"Sandbox pool not found: {pool_id}")
+
+
+class PoolInUseError(SandboxError):
```
eabatalov, line 47: SDK interface need to be clearly documented and define a clear boundary. So it's better to list every user accessible propertry separately for each class available to users in SDK interface, ie.

```
@property
def pool_id(self) -> str:
     return self._pool_id
```

src/tensorlake/sandbox/models.py
```diff
@@ -0,0 +1,199 @@
...
+    namespace: str
+    status: SandboxStatus
+    image: Optional[str] = None
+    resources: ContainerResourcesInfo
+    secret_names: List[str] = Field(default_factory=list)
+    timeout_secs: Optional[int] = None
+    entrypoint: Optional[List[str]] = None
+    network: Optional[NetworkConfig] = None
+    pool_id: Optional[str] = None
+    created_at: Optional[int] = None
```
eabatalov: All timestamps in user-visible SDK interfaces need to use native Python date data types so users don't need to parse the timestamps (not even knowing their format so far).

src/tensorlake/sandbox/sandbox.py
```diff
@@ -0,0 +1,368 @@
...
+
+        deadline = time.time() + timeout if timeout else None
+        while True:
+            info = self.get_process(proc.pid)
+            if info.status != ProcessStatus.RUNNING:
+                break
+            if deadline and time.time() > deadline:
+                self.kill_process(proc.pid)
+                raise SandboxError(f"Command timed out after {timeout}s")
+            time.sleep(0.1)
```
eabatalov, line 167: Why this polling interval? UX vs load. Please document.

.github/workflows/tests.yaml
eabatalov: See example how we do this with doc ai https://github.com/tensorlakeai/tensorlake/blob/main/.github/workflows/document_ai.yaml

This helps us to speed up CI by only running CI tests for the changed code paths in PRs.

tests/sandbox/test_lifecycle.py
```diff
@@ -0,0 +1,292 @@
...
+from testing import DataplaneProcessContextManager
+
+from tensorlake.sandbox import (
+    PoolNotFoundError,
+    SandboxClient,
+    SandboxNotFoundError,
+    SandboxStatus,
+)
+
+# ---------------------------------------------------------------------------
```
eabatalov: Please extract all this copy-paste into testing.py.

Also we can use setUpClass to avoid global vars.

.github/workflows/tests.yaml
diptanu: Document AI is not tied to Tensorlake's Compute Engine but sandboxes are so lets keep this here? Indexify Server and dataplane has tests for both sandboxes and allocations.

src/tensorlake/sandbox/client.py
diptanu: Addressed.

diptanu: Addressed

src/tensorlake/sandbox/sandbox.py
diptanu: Added comment

src/tensorlake/sandbox/sandbox.py
diptanu: Commented. 

        If this sandbox was created via ``create_and_connect()``, the
        sandbox is terminated (deleted from the server). Otherwise only
        the HTTP connection is closed and the sandbox keeps running.

src/tensorlake/sandbox/sandbox.py
diptanu: The sandbox object owns the client, as long as its active we don't create a new client. Do you mean why not create a client globally and inject it into various objects?

src/tensorlake/sandbox/client.py
diptanu: Fixed

src/tensorlake/sandbox/client.py
diptanu: Fixed

src/tensorlake/sandbox/client.py
diptanu: Its easier to call functions with one or two attributes in it vs creating a request object and then passing it to these creator functions. Same pattern is followed in other devtools products.

src/tensorlake/sandbox/client.py
diptanu: At the moment we can't because for applications we don't allow users to bring in random images but for sandboxes we do. I think mediium term in a few weeks we can expand the image api and let users bring in their own images.

PR #520, opened 2026-02-12
Relax grpcio dependencies

eabatalov:
Use old grpcio package version during development when we generate .py grpc stubs. This ensures that the .py proto stub files work with any newer version of grpcio package (same major version). See details about backward compatibility of grpc runtime components with generated code at https://protobuf.dev/support/cross-version-runtime-guarantee/. The old grpcio-tool version we're using is 1.60.0 from Dec 7, 2023. See https://pypi.org/project/grpcio/1.60.0/. Should be safe to use as the oldest supported.

Relax version requirements for non-development (real) grpcio package version in pyproject.toml. This version is used by pip install. This allows users to use grpcio version they want (within the same major version).


src/tensorlake/function_executor/proto/function_executor_pb2_grpc.py
```diff
@@ -1,35 +1,11 @@
...
 """Client and server classes corresponding to protobuf-defined services."""
-import warnings
-
 import grpc
 
 from tensorlake.function_executor.proto import (
     function_executor_pb2 as tensorlake_dot_function__executor_dot_proto_dot_function__executor__pb2,
 )
 
-GRPC_GENERATED_VERSION = "1.76.0"
```
eabatalov, line 11: I didn't remove this myself. The old version of grpcio-tools that we're using is not even doing the check. The check was added in 2024.

PR #525, opened 2026-02-13
Image builder split

calavera:


IMAGEBUILDER_V2_V3_COEXISTENCE.md
```diff
@@ -0,0 +1,268 @@
+# ImageBuilder v2/v3 Coexistence Implementation
```
eabatalov, line 1: Looks like this file is not needed?

PR #530, opened 2026-02-16
improve venv detection

### LINKED ISSUE #527: tensorlake deploy fails if there're absolute paths in .gitignore
Error message:

```
✅ All images built successfully
⚙️  Deploying applications...

Applications could not be deployed, please check the error message: Non-relative patterns are unsupported
Traceback (most recent call last):
  File "/Users/eugene/workplace/compute-engine-internal/indexify/tensorlake/src/tensorlake/cli/deploy.py", line 146, in _deploy_applications
    deploy_applications(
  File "/Users/eugene/workplace/compute-engine-internal/indexify/tensorlake/src/tensorlake/applications/remote/deploy.py", line 39, in deploy_applications
    ignored_absolute_paths: Set[str] = ignored_code_paths(applications_dir_path)
  File "/Users/eugene/workplace/compute-engine-internal/indexify/tensorlake/src/tensorlake/applications/remote/code/ignored_code_paths.py", line 40, in ignored_code_paths
    for match in root.glob(pattern):
  File "/Users/eugene/.pyenv/versions/3.10.18/lib/python3.10/pathlib.py", line 1032, in glob
    raise NotImplementedError("Non-relative patterns are unsupported")
NotImplementedError: Non-relative patterns are unsupported
Aborted!
```

This is because Path.glob used by us only support relative paths, see https://docs.python.org/3/library/pathlib.html#pathlib.Path.glob

See example: https://github.com/mmarcon/adk-101/blob/main/.gitignore#L142C1-L143C6

diptanu:
Improve venv detection 

Fixes #527

### DIFF
--- a/src/tensorlake/sandbox/models.py
+++ b/src/tensorlake/sandbox/models.py
@@ -131,6 +131,7 @@ class SandboxInfo(BaseModel):
     entrypoint: list[str] | None = None
     network: NetworkConfig | None = None
     pool_id: str | None = None
+    outcome: str | None = None
     created_at: OptionalTimestamp = None
     terminated_at: OptionalTimestamp = None
 
--- a/tests/sandbox/test_lifecycle.py
+++ b/tests/sandbox/test_lifecycle.py
@@ -636,6 +636,103 @@ class TestPoolDeletion(BaseSandboxTest):
 
 
 # ---------------------------------------------------------------------------
+# TestSandboxTimeout
+# ---------------------------------------------------------------------------
+
+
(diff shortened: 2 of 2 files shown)


src/tensorlake/applications/remote/code/ignored_code_paths.py
```diff
@@ -1,44 +1,140 @@
 import os
+import subprocess
 from pathlib import Path
-from typing import Set
+from typing import Optional, Set
+
+import click
```
calavera: This code can be called from the SDK, outside the CLI. We should not import `click` here.

PR #547, opened 2026-03-04
Preserve FunctionError semantics while surfacing child tracebacks

diptanu:
## Summary
- preserve the existing distinction between `RequestError` and runtime `FunctionError`
- stop swallowing child function runtime exceptions as generic "failed"
- forward child runtime tracebacks through `request_error_output` with an internal marker and reconstruct them into `FunctionError` on the parent side

## Details
- when a child function call raises a non-`RequestError`, FE now uploads traceback text with a private prefix marker
- FE still returns `ALLOCATION_FAILURE_REASON_FUNCTION_ERROR` for these runtime failures
- parent watcher path decodes marked payloads into `create_function_error(..., cause=traceback)`
- intentional `RequestError` paths are unchanged and remain `RequestError`

## Validation
- `python3 -m py_compile src/tensorlake/function_executor/allocation_runner/allocation_runner.py`
- full test suite not run in this environment (missing local deps/tooling)

### DIFF
--- a/src/tensorlake/function_executor/allocation_runner/allocation_runner.py
+++ b/src/tensorlake/function_executor/allocation_runner/allocation_runner.py
@@ -7,6 +7,7 @@ import time
 import weakref
 from collections.abc import Coroutine, Generator
 from dataclasses import dataclass
+from traceback import format_exception
 from typing import Any, Dict, List
 
 import grpc
@@ -121,6 +122,8 @@ from .upload import (
     upload_serialized_objects_to_blob,
 )
 from .value import SerializedValue, Value
+
+_CHILD_FUNCTION_ERROR_TRACE_PREFIX: str = "__tensorlake_function_error_trace__\n"
 
 
 @dataclass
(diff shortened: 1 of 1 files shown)


src/tensorlake/function_executor/allocation_runner/allocation_runner.py
```diff
@@ -1129,6 +1145,40 @@ def _run_allocation(self) -> AllocationResult:
                 uploaded_request_error_blob=uploaded_output_blob,
             )
         except BaseException as e:
+            # For child function calls (created by SDK), forward traceback via
```
eabatalov, line 1148: Don't copy-paste the logic here, reuse logic in `except RequestError as e:` branch.

src/tensorlake/function_executor/allocation_runner/allocation_runner.py
```diff
@@ -1129,6 +1145,40 @@ def _run_allocation(self) -> AllocationResult:
...
+            # request_error_output while preserving FUNCTION_ERROR semantics.
+            if (
```
eabatalov, line 1150: This if looks redundant and obscure.

src/tensorlake/function_executor/allocation_runner/allocation_runner.py
```diff
@@ -764,11 +767,24 @@ def _wait_function_call_future_completion(
...
                         )[0]
                     )
-                    future._set_exception(
-                        RequestError(
-                            message=serialized_request_error.data.decode("utf-8")
-                        )
+                    request_error_message: str = serialized_request_error.data.decode(
+                        "utf-8"
                     )
+                    if request_error_message.startswith(
```
eabatalov, line 773: Update FE proto comments to change and explain new request_error field its semantic.

eabatalov (changes requested):
Not clear what we're achieving with this change. We'll discuss this outside of this PR.

Besides the other comments, the gaps in this PR:

* Not implemented for LocalRunner.
* No tests that validate that backtraces are now present in FunctionError. The existing local runner test needs to get updated for the new UX, see test_original_function_exception_is_pretty_printed_in_local_mode and the same test needs be added for remote mode.

PR #561, opened 2026-03-09
Support for v3 image building implementation behind the version flag

thedebugger:
This PR adds the direct v3 application image-build path end to end. The client now creates v3 application builds at /images/v3/applications, sends the new multipart request shape, and wires the Python -> PyO3 -> Rust layers to the same contract. Compared to v2, v3 build requests also include per-image context_sha256, along with key, name, context_tar_part_name, and function_names.

On the SDK side, the Rust cloud client now exposes the v3 application-build APIs and the Python CloudClient surfaces them cleanly for the builder and CLI layers. The builder-side v3 flow now uses the real application-build create/info/cancel endpoints and streams per-image logs with labeled output and final summary reporting.

A key behavioral fix in this PR is v3 log streaming concurrency. The initial implementation was serializing concurrent log streams at the PyO3 boundary. The fix was to release the Python GIL around the long-lived Rust SSE stream methods, which allows multiple v3 image log streams to run concurrently and interleave correctly, matching the expected behavior.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "tensorlake"
-version = "0.4.16"
+version = "0.4.17"
 description = "Tensorlake SDK for Document Ingestion API and Serverless Applications"
 readme = "README.md"
 authors = [{ name = "Tensorlake Inc.", email = "support@tensorlake.ai" }]
@@ -9,7 +9,6 @@ dependencies = [
     "httpx[http2]>=0.27.2,<1.0",
     "pydantic>=2.0,<3.0",
     "httpx-sse>=0.4.1",
-    "click>=8.1,<9.0",
     "grpcio>=1.40.0,<2.0.0",
     "grpcio-tools>=1.40.0,<2.0.0",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,7 @@ members = ["crates/cloud-sdk", "crates/cli", "crates/rust-cloud-sdk-py"]
 resolver = "3"
 
 [workspace.package]
-version = "0.4.16"
+version = "0.4.17"
 authors = ["Tensorlake Inc. <support@tensorlake.ai>"]
 edition = "2024"
 license = "Apache-2.0"
--- a/crates/rust-cloud-sdk-py/pyproject.toml
+++ b/crates/rust-cloud-sdk-py/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "maturin"
 
 [project]
 name = "tensorlake-rust-cloud-sdk"
-version = "0.4.16"
+version = "0.4.17"
 description = "PyO3 bindings for Tensorlake Rust cloud client"
 requires-python = ">=3.10"
 
--- a/poetry.lock
+++ b/poetry.lock
@@ -114,7 +114,7 @@ version = "8.3.1"
 description = "Composable command line interface toolkit"
 optional = false
 python-versions = ">=3.10"
-groups = ["main", "dev"]
+groups = ["dev"]
 files = [
     {file = "click-8.3.1-py3-none-any.whl", hash = "sha256:981153a64e25f12d547d3426c367a4857371575ee7ad18df2a6183ab0545b2a6"},
     {file = "click-8.3.1.tar.gz", hash = "sha256:12ff4785d337a1bb490bb7e9c2b1ee5da3112e94a8622f26a6c77f5d2fc6842a"},
@@ -129,7 +129,7 @@ version = "0.4.6"
 description = "Cross-platform colored terminal text."
 optional = false
 python-versions = "!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*,!=3.5.*,!=3.6.*,>=2.7"
-groups = ["main", "dev"]
+groups = ["dev"]
 markers = "platform_system == \"Windows\""
 files = [
     {file = "colorama-0.4.6-py2.py3-none-any.whl", hash = "sha256:4f1d9991f5acc0ca119f9d443620b77f9d6b33703e51011c16baf57afb285fc6"},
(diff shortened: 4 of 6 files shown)


tests/test_cloud_client.py
```diff
@@ -0,0 +1,28 @@
...
+from tensorlake.applications.interface.exceptions import InternalError
+from tensorlake import cloud_client
+
+
+class FakeRustClientError(Exception):
+    pass
+
+
+class TestCloudClientErrors(unittest.TestCase):
+    def test_raise_as_tensorlake_error_surfaces_connection_details(self):
```
thedebugger, line 16: add another test case for unclassified error

crates/rust-cloud-sdk-py/src/lib.rs
```diff
@@ -1066,27 +1169,47 @@ fn into_py_error(error: SdkError) -> PyErr {
...
             } else {
-                CloudApiClientError::new_err((
-                    "internal",
-                    Option::<u16>::None,
-                    http_error.to_string(),
-                ))
+                CloudApiClientError::new_err(("internal", Option::<u16>::None, message))
+            }
+        }
+        SdkError::Middleware(middleware_error) => {
```
thedebugger, line 1176: surfacing inner errors (like connect error) which were getting swallowed previously

src/tensorlake/applications/applications.py
```diff
@@ -21,6 +21,25 @@ def filter_applications(
         yield function
 
 
+def functions_for_application(
+    application: Function,
+    functions: Iterator[Function],
+) -> list[Function]:
+    """Returns functions included in the application build request.
+
+    This is intentionally broader than the application's reachable function graph.
```
eabatalov: It's not possible to know which functions are used by an application because this is decided by user code depending on any condition or even a dynamic import. This comment is misleading and very complex.

src/tensorlake/applications/applications.py
```diff
@@ -21,6 +21,25 @@ def filter_applications(
         yield function
 
 
+def functions_for_application(
```
eabatalov, line 24: Application functions can be called from any application and non-application function that can import it. Please remove this functions_for_application filter and just deploy all functions including application functions per application. We even have tests that verify calling applications functions from non-application functions.

src/tensorlake/applications/image.py
```diff
@@ -41,13 +41,18 @@ def image_infos() -> Dict[Image, ImageInformation]:
 def create_image_context_file(img: Image, file_path: str) -> None:
     """Create a tar.gz file containing the Dockerfile and all necessary files for building the image"""
     try:
```
eabatalov, line 60: Why are all these import workaround used here?

src/tensorlake/applications/interface/exceptions.py
```diff
@@ -45,7 +45,14 @@ def message(self) -> str:
         return self.args[0]
 
 
-class RemoteAPIError(TensorlakeError):
+class RemoteError(TensorlakeError):
+    """Raised when communication with a remote Tensorlake service failed."""
```
eabatalov: This is a more generic, not about "communication".

"Raised when communication with a remote Tensorlake service failed" -> "Raised on a failure with remote Tensorlake service".

src/tensorlake/applications/interface/exceptions.py
```diff
@@ -56,6 +63,13 @@ def __str__(self):
         return f"{super().__str__()} (Status Code: {self.status_code})"
 
 
+class RemoteTransportError(RemoteError):
+    """Raised when a remote Tensorlake request failed before receiving a response."""
```
eabatalov: Something is not right here. If a request failed then it's not a transport error, it's a request error. Either exception name is wrong or docsting.

eabatalov: The confusion is coming from the fact that a Request is one of the primary entities in Tensorlake Applications. An application run/invocation is called a request. You can see RequestError exception in this file. So let's clarify the naming to avoid confusion. This is SDK interface, it needs to be clear.

src/tensorlake/applications/applications.py
thedebugger: Since goal was to only build images for an application. Since we can't map functions to application, we would have to build images for all functions which is not ideal. That is what this comment is trying to say. I'll make it  concise

src/tensorlake/applications/interface/exceptions.py
thedebugger: I think my changes are at an incorrect place. Surface area was supposed to be b/w CLI -> rust client -> tensorlake - exception was used to pass information if that call fails. It was not meant to change public APIs. Let me go back on this one.

src/tensorlake/applications/applications.py
thedebugger: Added a concise comment about this and encoded behavior in test case

src/tensorlake/applications/interface/exceptions.py
thedebugger: Reverted new types for now since this was suppose to be internal only

src/tensorlake/applications/applications.py
```diff
@@ -21,6 +21,20 @@ def filter_applications(
...
+    functions: Iterator[Function],
+) -> list[Function]:
+    """Returns all loaded functions for the application build request.
+
+    We cannot reliably determine application membership statically, and one
+    application can call another application's function at runtime. For
+    compatibility with deploy/runtime packaging, each application build includes
+    all loaded Tensorlake functions.
+    """
+    return list(functions)
```
eabatalov, line 35: Looks a bit odd to have such a trivial function but also this gives a fairly good abstraction with explanation. lgtm.

pyproject.toml
```diff
@@ -9,6 +9,7 @@ dependencies = [
     "httpx[http2]>=0.27.2,<1.0",
     "pydantic>=2.0,<3.0",
     "httpx-sse>=0.4.1",
+    "click>=8.1,<9.0",
```
calavera: I don't think we want to add more dependencies to the python project because they become available in people's projects that depend on our sdk. The client_v3.py should not have any print on its own, the Rust code should do the reporting instead.

PR #569, opened 2026-03-11
Fix tensorlake secret command

thedebugger:


src/tensorlake/cli/deploy.py
```diff
@@ -212,7 +212,7 @@ def deploy(
 
     missing = _warning_missing_secrets(auth, list(list_secret_names()))
     if missing:
-        _emit({"type": "missing_secrets", "count": len(missing)})
+        _emit({"type": "missing_secrets", "count": len(missing), "names": missing})
```
thedebugger, line 215: add missing secret names to improve UX

PR #571, opened 2026-03-11
Fix image hash determinism and error propagation

thedebugger:
## Summary - These are good hygiene fixes but very small likelyhood of running into it (caught with claude review)

- **Bug 1**: `hash_directory` iterated directory entries in filesystem order (non-deterministic). Now sorts by path and includes relative file paths in hash for structural correctness.
- **Bug 2**: `render_build_operation` joined HashMap options in random iteration order, producing non-deterministic Dockerfiles. Now sorts options by key.
- **Bug 3**: `hash_directory` called `.unwrap()` on IO errors, panicking the process. Now propagates `io::Result` through `add_build_op_to_hasher` and `image_hash`.
- Removes unused `httpx-sse` Python dependency.

## Test plan
- [x] `cargo test -p tensorlake-cloud-sdk -- models::tests` — 3 new tests covering all 3 bugs
- [ ] Verify `cargo check -p tensorlake-cli` compiles (changed `image_hash` return type)

### DIFF
--- a/crates/cloud-sdk/src/images/mod.rs
+++ b/crates/cloud-sdk/src/images/mod.rs
@@ -181,7 +181,7 @@ impl ImagesClient {
         request
             .image
             .create_context_archive(&mut context_data, &request.sdk_version, None)?;
-        let image_hash = request.image.image_hash(&request.sdk_version);
+        let image_hash = request.image.image_hash(&request.sdk_version)?;
         let form = Form::new()
             .text("graph_name", request.application_name.clone())
             .text("graph_version", request.application_version.clone())
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2450,6 +2450,7 @@ dependencies = [
  "serde_with",
  "sha2",
  "tar",
+ "tempfile",
  "thiserror 2.0.18",
  "tokio",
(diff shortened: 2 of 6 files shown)


crates/cloud-sdk/src/images/mod.rs
```diff
@@ -181,7 +181,7 @@ impl ImagesClient {
         request
             .image
             .create_context_archive(&mut context_data, &request.sdk_version, None)?;
-        let image_hash = request.image.image_hash(&request.sdk_version);
+        let image_hash = request.image.image_hash(&request.sdk_version)?;
```
eabatalov, line 184: Do we really use image hash for anything? I'm not aware where it'd be helpful as image hash computed based on image definition doesn't guarantee the same built image content after built because i.e. newer version of packages get installed with the same command like `pip install numpy`.

thedebugger: yeah we don't use it right now which is why this fix has no/little impact

eabatalov: We might also want to remove it. This is preferred imo than having a non-trivial dead code.

PR #572, opened 2026-03-12
Make v3 image builder the default

thedebugger:
Included in the PR:

* v3 image builder is now the default for deploy
* restored GIL release during v3 image log streaming
* added terminal SSE regression coverage
* fixed the city guide sample to avoid local agents import requirements
* bumped the CLI/package version to 0.4.20

### DIFF
--- a/examples/readme_example/city_guide.py
+++ b/examples/readme_example/city_guide.py
@@ -3,6 +3,9 @@ import subprocess
 import sys
 import tempfile
 
+from agents import Agent, Runner
+from agents.tool import WebSearchTool, function_tool
+
 from tensorlake.applications import Image, application, function, run_local_application
 
 # Define the image with necessary OpenAI dependencies.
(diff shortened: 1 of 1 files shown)


examples/readme_example/city_guide.py
```diff
@@ -156,14 +160,32 @@ def city_guide_app(city: str) -> str:
     2. Get activity based on weather using Web Search.
     3. Create final guide with appropriate units.
     """
+    from agents import Agent, Runner
+    from agents.tool import function_tool
+
+    @function_tool
```
thedebugger: lazy loading function tool calls because user might not have OpenAI SDK installed locally, which causes function_tool load to fail during the deploy

eabatalov: This adds a bunch of wrapper tool functions. This makes the example more verbose and hides the fact that tensorlake functions are fully able to work as @function_tool in the agentic SDK. This showcases our product as worse than as a better thing.

We don't do this in any of our examples afaik.
People before deploying remotely their code would typically be developing it in local mode.

For these reasons I think this change makes things worse.

PR #576, opened 2026-03-12
Add `sbx terminate` and `sbx snapshot rm` commands

bkolobara:
## Summary

This PR adds CLI support for deleting snapshots and terminating sandboxes.

### Changes

- Added `tl sbx snapshot rm <SNAPSHOT_ID>...` to delete one or more snapshots
- Added sandbox termination support in the CLI and exposed it as:
  - `tl sbx terminate <SANDBOX_ID>...`
- Kept `tl sbx rm` as a compatibility alias for `tl sbx stop`

### Why

Server-side behavior differs between sandboxes and snapshots:

- deleting a sandbox does **not** hard-delete it; it **terminates** the sandbox and keeps it visible in sandbox listings as `terminated`
- deleting a snapshot performs a real delete and removes it from snapshot listings

I think that `terminate` is a better user-facing command for sandboxes, while `snapshot rm` is better for snapshots.

### DIFF
--- a/crates/cli/src/commands/sbx/terminate.rs
+++ b/crates/cli/src/commands/sbx/terminate.rs
@@ -16,14 +16,14 @@ pub async fn run(ctx: &CliContext, sandbox_ids: &[String]) -> Result<()> {
             let status = resp.status();
             let body = resp.text().await.unwrap_or_default();
             return Err(CliError::Other(anyhow::anyhow!(
-                "failed to stop sandbox {} (HTTP {}): {}",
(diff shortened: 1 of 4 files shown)


crates/cli/src/commands/sbx/terminate.rs
```diff
@@ -0,0 +1,30 @@
+use crate::auth::context::CliContext;
```
eabatalov, line 1: Suggestion: let's rename all of this to terminate, update all the messages.

PR #595, opened 2026-03-24
Remove resume call on ssh and set sandbox default to 1 CPU and 1GB RAM

bkolobara:


crates/cli/src/commands/sbx/ssh.rs
```diff
@@ -10,8 +8,6 @@ pub async fn run(ctx: &CliContext, sandbox_id: &str, shell: &str) -> Result<()>
         return Err(CliError::usage("ssh requires an interactive terminal"));
     }
 
-    resume_sandbox_if_suspended(ctx, sandbox_id).await?;
```
thedebugger, line 13: do you know why we added this? And, what is going to be the behavior if someone tries to SSH to suspended VM?

bkolobara: This is now handled directly by the server. This was a leftover from the previous implementation where we did the resume from the CLI, but we don't need it anymore.

PR #598, opened 2026-03-26
Use FilesystemOnly snapshots for sandbox images

eabatalov:
## Summary
- Thread `snapshot_content_mode` parameter through Rust cloud SDK, PyO3 bindings, and Python SDK layers
- `create_sandbox_image` now creates `FilesystemOnly` snapshots instead of `Full`
- This allows `tl sbx new --image` to override CPU, memory, and other resource settings (previously rejected by server for Full snapshots)
- Added image ls and describe commands
- Let the server figure out the defaults when creating a new sandbox from a snapshot (phase 1). Once server changes are deploy, we can remove all defaults from client side. Client-side should only pass user overrides
- Make 1024 MB default consistent

This enables the following use case:

1. User runs `tl sbx create-image ./image_file.py` to create a bootable image from the Image object.
2. User runs `tl sbx new --image image_name` to create a new sandbox that boots from the image.

## Test plan
- [x] All Rust crates compile (`cargo check` for cloud-sdk, rust-cloud-sdk-py, cli)
- [x] Python unit tests pass (`tests/cli/test_create_sandbox_image.py`)
- [x ] E2E: `tl sbx create-image` then `tl sbx new --image <name> --cpus 4 --memory 8192` succeeds

### DIFF
--- a/src/tensorlake/sandbox/client.py
+++ b/src/tensorlake/sandbox/client.py
@@ -247,7 +247,7 @@ class SandboxClient:
             image: Container image to use
             cpus: Number of CPUs to allocate. Defaults to 1.0 for non-snapshot
                 creates; omitted for snapshot restores unless explicitly set.
-            memory_mb: Memory in megabytes. Defaults to 2048 for non-snapshot
+            memory_mb: Memory in megabytes. Defaults to 1024 for non-snapshot
                 creates; omitted for snapshot restores unless explicitly set.
             ephemeral_disk_mb: Ephemeral disk space in megabytes. Defaults to
                 1024 for non-snapshot creates; omitted for snapshot restores
@@ -516,7 +516,7 @@ class SandboxClient:
         self,
         image: str | None = None,
         cpus: float = 1.0,
-        memory_mb: int = 2048,
+        memory_mb: int = 1024,
         ephemeral_disk_mb: int = 1024,
         secret_names: list[str] | None = None,
         timeout_secs: int = 0,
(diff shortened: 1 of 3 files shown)


crates/cli/src/commands/sbx/create.rs
```diff
@@ -127,23 +86,79 @@ pub async fn run(
...
+fn build_create_request_body(
+    cpus: Option<f64>,
+    memory: Option<i64>,
+    timeout: Option<i64>,
+    entrypoint: &[String],
+    snapshot_id: Option<&str>,
+) -> serde_json::Value {
+    let mut body = serde_json::json!({});
+
+    if let Some(snapshot_id) = snapshot_id {
```
thedebugger, line 106: if snapshot_id is set, then only pass user overrides and let server figure out right defaults

src/tensorlake/sandbox/client.py
```diff
@@ -243,9 +245,13 @@ def create(
 
         Args:
             image: Container image to use
-            cpus: Number of CPUs to allocate
-            memory_mb: Memory in megabytes
-            ephemeral_disk_mb: Ephemeral disk space in megabytes
+            cpus: Number of CPUs to allocate. Defaults to 1.0 for non-snapshot
+                creates; omitted for snapshot restores unless explicitly set.
+            memory_mb: Memory in megabytes. Defaults to 2048 for non-snapshot
```
thedebugger: Defaults to 1024

PR #606, opened 2026-03-31
Use Filesystem only snapshots for sandbox images

thedebugger:
## Summary
- Thread `snapshot_content_mode` parameter through Rust cloud SDK, PyO3 bindings, and Python SDK layers
- `create_sandbox_image` now creates `FilesystemOnly` snapshots instead of `Full`
- This allows `tl sbx new --image` to override CPU, memory, and other resource settings (previously rejected by server for Full snapshots)
- Added image ls and describe commands
- Let the server figure out the defaults when creating a new sandbox from a snapshot (phase 1). Once server changes are deploy, we can remove all defaults from client side. Client-side should only pass user overrides
- Make 1024 MB default consistent

This enables the following use case:

1. User runs `tl sbx create-image ./image_file.py` to create a bootable image from the Image object.
2. User runs `tl sbx new --image image_name` to create a new sandbox that boots from the image.

## Test plan
- [x] All Rust crates compile (`cargo check` for cloud-sdk, rust-cloud-sdk-py, cli)
- [x] Python unit tests pass (`tests/cli/test_create_sandbox_image.py`)
- [x ] E2E: `tl sbx create-image` then `tl sbx new --image <name> --cpus 4 --memory 8192` succeeds

### DIFF
--- a/crates/cli/src/commands/sbx/image/create.rs
+++ b/crates/cli/src/commands/sbx/image/create.rs
@@ -11,22 +11,22 @@ use crate::python_ast::{self, ImageDef, OpDef};
 
 pub async fn run(
     ctx: &CliContext,
-    application_file_path: &str,
+    image_file_path: &str,
     image_name: Option<&str>,
-    template_name: Option<&str>,
+    registered_name: Option<&str>,
 ) -> Result<()> {
     // 1. Parse the Python file for Image definitions.
-    let abs_path = tokio::fs::canonicalize(application_file_path)
+    let abs_path = tokio::fs::canonicalize(image_file_path)
         .await
-        .map_err(|e| CliError::usage(format!("Cannot read '{}': {}", application_file_path, e)))?;
+        .map_err(|e| CliError::usage(format!("Cannot read '{}': {}", image_file_path, e)))?;
     let app_dir = abs_path.parent().unwrap_or(&abs_path).to_path_buf();
 
-    eprintln!("\u{2699}\u{fe0f}  Loading {}...", application_file_path);
+    eprintln!("\u{2699}\u{fe0f}  Loading {}...", image_file_path);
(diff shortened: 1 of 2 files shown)


crates/cli/src/commands/sbx/image/create.rs
eabatalov: This arg and CLI arg should be named image_file_path with its description updated.

crates/cli/src/commands/sbx/image/create.rs
```diff
@@ -56,20 +56,21 @@ async fn run_build_and_register(
...
+    eprintln!("\u{1f4f8} Snapshot created: {}", snapshot_id);
 
     // 6. Build Dockerfile text.
     let sdk_version = env!("CARGO_PKG_VERSION");
     let image = build_cloud_sdk_image(image_def)?;
     let dockerfile = image.dockerfile_content(sdk_version, None);
 
     // 7. Register template via Platform API.
-    eprintln!("⚙️  Registering template...");
+    eprintln!("\u{2699}\u{fe0f}  Registering template...");
```
eabatalov: Rename all occurences of template to image

eabatalov: We use image term, not template term.

src/tensorlake/sandbox/client.py
```diff
@@ -694,7 +694,7 @@ def create_and_connect(
         self,
         image: str | None = None,
         cpus: float = 1.0,
-        memory_mb: int = 2048,
```
eabatalov, line 697: Seems like you changed default from 2 GB to 1 GB, just fyi.

thedebugger: this one was not getting applied. it was always 1GB. This PR makes it defaults more consistent.

PR #638, opened 2026-04-17
Use routing hints to skip proxy placement lookup on first request

calavera:
## Summary

Client-side implementation of [compute-engine-internal#680](https://github.com/tensorlakeai/compute-engine-internal/pull/680).

When the server is configured with a hint signing key, the sandbox create/claim response now carries a short-lived `routing_hint` HMAC-SHA256 token. This PR threads that token through to the first proxy request as `X-Tensorlake-Route-Hint`, letting the sandbox proxy route directly to the dataplane without an Indexify placement lookup.

- **Rust cloud-sdk** (`crates/cloud-sdk/`): adds `routing_hint` to `CreateSandboxResponse`, adds `with_routing_hint()` to `SandboxProxyClient`, includes `X-Tensorlake-Route-Hint` on all proxy requests when a hint is present
- **Rust Python bindings** (`crates/rust-cloud-sdk-py/`): exposes `routing_hint` param on `CloudSandboxProxyClient`
- **Python SDK** (`src/tensorlake/sandbox/`): parses `routing_hint` from create response, threads it through `connect()` into the proxy client; `create_and_connect()` fast-paths when status is already `running` (skips the redundant poll RTT)
- **TypeScript SDK** (`typescript/src/`): same pattern — `routingHint` on `CreateSandboxResponse`, `SandboxOptions`, and `HttpClientOptions`; `connect()` accepts the hint; `createAndConnect()` has the same fast path

The hint has a 5-second TTL and the proxy falls through silently on any failure, so no client-visible errors on expiry or misconfiguration.

## Test plan

- [ ] Rust: `cargo check -p tensorlake-cloud-sdk -p tensorlake-rust-cloud-sdk-py` passes
- [ ] TypeScript: `tsc --noEmit` passes
- [ ] Integration: create a sandbox, verify first proxy request includes `X-Tensorlake-Route-Hint` header when server is configured with signing key

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,7 @@ members = ["crates/cloud-sdk", "crates/cli", "crates/rust-cloud-sdk-py"]
 resolver = "3"
 
 [workspace.package]
-version = "0.4.46"
+version = "0.4.47"
 authors = ["Tensorlake Inc. <support@tensorlake.ai>"]
 edition = "2024"
 license = "Apache-2.0"
--- a/crates/cloud-sdk/src/sandboxes/mod.rs
+++ b/crates/cloud-sdk/src/sandboxes/mod.rs
@@ -223,6 +223,7 @@ impl SandboxesClient {
 pub struct SandboxProxyClient {
     client: Client,
     host_override: Option<String>,
+    routing_hint: Option<String>,
 }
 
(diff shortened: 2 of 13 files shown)


cooleel (approved):
LGTM, thanks!

PR #675, opened 2026-05-11
Run offline rootfs builder for sandbox images

diptanu:
## Summary
- route sandbox image builds through Platform API's offline rootfs build prepare/complete flow
- start registered-parent diff builds from the Platform API-selected parent image, apply Dockerfile steps there, then run the offline rootfs builder wrapper
- resolve rootfs builder commands to /usr/local/bin when the API returns a bare command, and set PATH so wrapper subprocesses can find indexify-rootfs-snapshot
- wire Python SDK, TypeScript SDK, and Rust CLI build specs through rootfsPath/parentRootfsPath
- fail closed when Platform API prepares an unregistered OCI base as a customer-owned rootfs base until the Docker-export rootfs materializer lands, so the clients do not snapshot tensorlake/rootfs-builder as the user image
- add unit coverage for builder command path resolution, offline rootfs builder flow, and the unmaterialized OCI-base guard

## Supporting PRs
- tensorlakeai/platform-api#502 prepares unregistered OCI bases as customer-owned rootfs base builds and keeps registered parents as diffs
- tensorlakeai/provisioner#767 injects tl-rootfs-build and indexify-rootfs-snapshot into all managed rootfs images so parent images can run the diff builder
- tensorlakeai/compute-engine-internal#880 lets indexify-rootfs-snapshot size /dev/vda as a block device instead of rejecting it as an empty rootfs

## Validation
- cargo fmt --check
- cargo test -p tensorlake-cli requires_oci_base_import_when_base_builder_differs_from_from_image
- poetry run black src/tensorlake/image/sandbox_builder.py tests/image/test_sandbox_builder.py
- PYTHONPATH=src poetry run python -m unittest tests.image.test_sandbox_builder.TestOfflineRootfsBuilder
- npm exec --yes --package=node@24 -- bash -lc 'cd typescript && node --version && npm test -- tests/sandbox-image.test.ts'
- git diff --check

### DIFF
--- a/crates/cli/src/commands/sbx/image/create.rs
+++ b/crates/cli/src/commands/sbx/image/create.rs
@@ -4,11 +4,15 @@ use std::{
     time::Duration,
 };
 
+use serde::Deserialize;
 use serde_json::{Map, Value};
 use shlex::split as shlex_split;
 use tensorlake::{
     Client,
-    sandbox_templates::models::CreateSandboxTemplateRequest,
+    sandbox_templates::models::{
+        CompleteSandboxTemplateBuildRequest, CreateSandboxTemplateRequest,
+        PrepareSandboxTemplateBuildRequest, RootfsNodeKind, SandboxTemplateBuildPrepared,
+    },
     sandboxes::{
         SandboxProxyClient, SandboxesClient,
         models::{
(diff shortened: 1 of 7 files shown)


diptanu:
Closing this version because it still depends on the fragile live-rootfs builder path that we abandoned in tensorlakeai/compute-engine-internal#880. The replacement should build/export into an unmounted ext4 rootfs image in tensorlake/rootfs-builder, then snapshot that regular file through the offline rootfs snapshot tool.

PR #678, opened 2026-05-13
fix(sandbox): add LOCAL_READY snapshot status

sgirones:
## Summary

- Adds `LOCAL_READY = "local_ready"` to `SnapshotStatus` enum so the server's new terminal status doesn't cause a Pydantic validation error
- `snapshot_and_wait` now returns successfully on `LOCAL_READY` alongside `COMPLETED`

## Test plan

- [ ] Verify sandbox stress test passes after this SDK version is published and picked up by indexify-deployment CI


sgirones:
Superseded —  and  are already in main.

PR #689, opened 2026-05-21
RFC, do not merge feat(sandbox-image): support USER in Dockerfile and Image DSL

eabatalov:
## Summary

- Stop rejecting `USER` in the Python and TypeScript sandbox-image Dockerfile parsers; `docker build` already handles USER semantics natively for the rootfs builder.
- Add `Image.user(name, group=None)` to the shared Python/TypeScript `Image` DSL (accepts names or numeric uid/gid; renders as Docker's `USER <user>[:<group>]`).
- Suppress `USER` rendering on the Applications-side Dockerfile (`tensorlake.image.utils.dockerfile_content`) so the trailing root-only `pip install tensorlake` step still works. Emits a one-time warning when an Image used by Applications has `.user()` set. TypeScript has no Applications render path, so only the sandbox flow is touched there.

## Safety / coordination

All changes here are additive and require no dataplane coordination:

- The Rust client parser (`crates/cloud-sdk/src/sandbox_images.rs:1139-1149`) already accepts USER (only validates FROM), so the dataplane request shape is unchanged.
- `docker build` natively handles USER, so build infrastructure needs no changes.
- The `include_user=False` filter on the Applications render path means existing `@function` Dockerfiles are byte-identical.

## Follow-ups (documented here, not in this PR)

The runtime default ProcessUser will use **safety-preserving semantics**: `tl-user` stays the default for sandboxes booted from images whose Dockerfile contains no `USER` directive. Only an explicit `USER` (or `Image.user(...)`) overrides the default.

| Dockerfile | Runtime default ProcessUser |
|---|---|
| no `USER` | `tl-user` (unchanged) |
| `USER appuser` | `appuser` |
| `USER root` (explicit) | `root` (opt-in) |
| `USER 1000:1000` | `1000:1000` |

### Builder: write `/etc/tensorlake/sandbox.json`

After `docker build`, the rootfs-builder (`tl-rootfs-build`, source in a sibling repo) should write the runtime default ProcessUser to `/etc/tensorlake/sandbox.json` **only when the Dockerfile contained an explicit `USER` directive**:

```json
{\"default_process_user\": {\"name\": \"appuser\", \"group\": \"appgroup\"}}
```

Images without `USER` either omit the file or leave the field absent — the daemon falls back to `tl-user` in either case. Concretely, the builder needs to distinguish \"Dockerfile had a USER line\" from \"Config.User defaulted to empty/root because nothing was set\", and only write the field in the first case.

This is a no-op until something reads it. Old daemons booting new images ignore the file; current SDKs still pass explicit \`user=\"tl-user\"\`, so it isn't consulted. Independently safe to release whenever the rootfs-builder change is ready.

### PR B: flip SDK default to image-derived (with tl-user fallback)

Once the daemon knows how to (a) read \`/etc/tensorlake/sandbox.json\` to resolve a default user, (b) fall back to \`tl-user\` when the file/field is absent, and (c) accept a \"use default\" sentinel on the wire:

- \`src/tensorlake/sandbox/sandbox.py:753, 835\` and \`async_sandbox.py:372, 429\`: \`user: ProcessUser = \"tl-user\"\` → \`user: ProcessUser | None = None\`
- TypeScript mirrors in \`typescript/src/sandbox.ts\`
- \`_normalize_process_user\` (\`sandbox.py:713\`): emit the new sentinel when \`user is None\`

**Hard ordering constraint:** that SDK change must ship only after the daemon update is rolled out, or \`start_process\` calls without an explicit \`user=\` will hit older daemons that mis-handle the sentinel. Gate on a daemon-version check if available; otherwise enforce by release order.

## Test plan

- [x] \`poetry run python -m unittest tests.image.test_sandbox_builder\` — all new USER tests pass (one pre-existing macOS \`/private/var\` symlink flake in \`test_image_build_does_not_write_generated_dockerfile_into_context\` is unrelated).
- [x] \`pnpm vitest run tests/sandbox-image.test.ts\` — 20/20 tests pass, including the 3 new USER cases.
- [ ] End-to-end: build a sandbox image with \`Image(...).run('useradd -m appuser').user('appuser').run('id')\` against staging and verify the second \`RUN\` records \`uid=appuser\` in build logs.
- [ ] End-to-end: deploy an \`@function\` with \`image=Image().user('appuser')\` and confirm the resulting Applications Dockerfile contains no \`USER\` line (the warning fires) and the SDK install step still completes.

### DIFF
--- a/src/tensorlake/image/_dockerfile.py
+++ b/src/tensorlake/image/_dockerfile.py
@@ -23,15 +23,20 @@ def render_op_line(op: _ImageBuildOperation) -> str:
     return f"{op.type.name}{options_str} {' '.join(op.args)}"
 
 
-def image_to_dockerfile(image: Image) -> str:
+def image_to_dockerfile(image: Image, *, include_user: bool = True) -> str:
     """Render a plain Dockerfile string from an :class:`Image`.
 
     Mirrors the TypeScript ``dockerfileContent`` exactly.
+
+    When ``include_user`` is False, ``USER`` ops are skipped. Applications
+    image builds use this to keep the trailing SDK install step as root.
     """
     lines: list[str] = []
     if image._base_image:
         lines.append(f"FROM {image._base_image}")
     for op in image._build_operations:
+        if not include_user and op.type == _ImageBuildOperationType.USER:
+            continue
         lines.append(render_op_line(op))
     return "\n".join(lines)
 
--- a/src/tensorlake/image/image.py
+++ b/src/tensorlake/image/image.py
@@ -15,6 +15,7 @@ class _ImageBuildOperationType(Enum):
     ENV = 3
     RUN = 4
     WORKDIR = 5
+    USER = 6
 
 
 @dataclass
@@ -93,6 +94,16 @@ class Image:
             _ImageBuildOperation(
                 type=_ImageBuildOperationType.WORKDIR,
                 args=[directory],
+                options={},
+            )
+        )
+
+    def user(self, name: str | int, group: str | int | None = None) -> "Image":
+        rendered = f"{name}:{group}" if group is not None else f"{name}"
+        return self._add_operation(
+            _ImageBuildOperation(
+                type=_ImageBuildOperationType.USER,
+                args=[rendered],
                 options={},
             )
         )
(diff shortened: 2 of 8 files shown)


eabatalov:
I'm extending the scope of this from beyond USER command.

PR #698, opened 2026-05-26
sandbox-image: complete with dataplane-signed URI

sgirones:
## Summary

Sandbox image builds now get their rootfs upload URLs from the dataplane instead of from platform-api. When `/prepare` returns `snapshotRelPath`, the CLI asks the builder sandbox's proxy to call dataplane `POST /api/v1/blob/sign`; dataplane composes the final object URI from its regional config, signs the multipart upload, and returns that canonical `uri` with the upload spec.

The CLI treats dataplane's signed `uri` as the source of truth. It writes that URI into `spec.snapshotUri` so the in-sandbox builder keeps the existing input contract, keeps the same URI in CLI state, and sends it directly to `/complete`. Builder metadata still supplies computed fields like `snapshot_size_bytes`, `snapshot_format_version`, and `rootfs_disk_bytes`; if metadata includes `snapshot_uri`, the CLI checks that it matches the dataplane URI.

The legacy `/prepare` shape is unchanged. If `snapshotRelPath` is absent, the prepared spec already contains `upload`, and completion continues to use builder metadata with the prepared `snapshotUri` fallback.

## Notes For Review

- `snapshotRelPath` is only used to request dataplane signing. Completion is keyed off the saved signed URI, not the rel path.
- The new upload response must include `uri`; missing `uri` fails before the builder runs.
- The upload block remains mostly opaque JSON in the passthrough spec. The CLI only reads `uri`.
- Diff builds still sign parent manifests as full-URI `SingleGet` requests.

## Test Plan

- [x] `cargo +nightly fmt --check`
- [x] `git diff --check`
- [x] `cargo test -p tensorlake sandbox_images`
- [ ] Manual smoke against legacy platform-api response with embedded `upload`.
- [ ] Manual smoke against new platform-api + sandbox-proxy `sign_blob`.

## Related PRs

- `platform-api`: https://github.com/tensorlakeai/platform-api/pull/530
- `compute-engine-internal`: https://github.com/tensorlakeai/compute-engine-internal/pull/1039
- `tensorlake`: https://github.com/tensorlakeai/tensorlake/pull/698

## Flow

```
CLI -> platform-api /prepare
  <- snapshotRelPath

CLI -> sandbox-proxy /api/v1/blob/sign { rel_path, multipart_put }
  -> dataplane validates namespace, composes final URI, signs upload
  <- { uri, uploadId, partUrls, completeUrl, abortUrl }

CLI writes uri into spec.snapshotUri
CLI runs the in-sandbox builder
CLI completes with the saved dataplane-signed uri
```

For parent snapshot reads, the CLI uses `{ uri, op: SingleGet }` because the parent manifest already has a full stored URI.


thedebugger:
Updated this PR to send the provider-neutral /api/v1/blob/sign request shape (Artifact + PutArtifact for snapshot uploads, Blob + GetBlob for parent downloads). This now depends on the server/dataplane API change in tensorlakeai/compute-engine-internal#1096.

sgirones:
LGTM

thedebugger (approved):

PR #699, opened 2026-05-26
Add hidden --json flag to tl sbx image ls

sgirones:
## Summary
- `tl sbx image ls` now accepts a hidden `--json` flag that prints the image list as a pretty-printed JSON array (the same `Vec<serde_json::Value>` already used to render the table). The default human-readable table output is unchanged.
- Mirrors the existing hidden `--json` flag on `tl sbx image create`.

## Motivation
`indexify-deployment`'s nightly stress-test image build workflow (`tests/build_sandbox_image/build.py`) calls `tl sbx image ls --json` to GC old images. That flag never existed, so cleanup has been silently skipped on every run since PR #460 there (2026-04-14), and stress-test images have been accumulating for ~6 weeks. Example failure from a recent run:

```
--- Cleaning up old stress-test images ---
  Warning: could not list images: error: unexpected argument '--json' found
```

Adding the flag here is the cleanest fix — parsing the `comfy_table` output from the Python side would be fragile.

## Test plan
- [x] `cargo check -p tensorlake-cli` passes
- [ ] Manual: `tl sbx image ls --json` against a real org returns a JSON array parseable by `json.loads`
- [ ] Manual: `tl sbx image ls` (no flag) still renders the table as before
- [ ] Manual: `tl sbx image ls --help` does NOT show `--json` (it's `hide = true`)


calavera (approved):
I would probably not hide these flags, they sound useful 🤷

sgirones:
makes sense. I made both visible

PR #706, opened 2026-05-28
Let rootfs builder interpret Dockerfile instructions

diptanu:
Updates the sandbox image build path on current main so the SDK parser only validates the single-stage FROM requirement and preserves Dockerfile instructions for the rootfs builder/docker build to interpret.\n\nThis replaces the older stale branch contents; main already contains the Rust rootfs-builder routing work and process-user fixes.\n\nValidation:\n- cargo test -p tensorlake sandbox_images::tests -- --nocapture\n- cargo fmt --check\n\nNote: TypeScript build was not run locally in the clean worktree because TS dependencies were not installed there (tsup missing); the TS change is documentation-only.


eabatalov:
rootfs builder already doing this

Issue #528, opened 2026-02-16
Review and fix Sandbox client code


eabatalov:
The PR https://github.com/tensorlakeai/tensorlake/pull/519 was merged with some not addressed comments.
Need to review and fix the gaps.

PR #723, opened 2026-06-06
fix(cloud-sdk): fail closed on indeterminate sandbox process exit

jare1686:
## Summary

`run_streaming_process` (`crates/cloud-sdk/src/sandbox_images.rs`) treats a sandbox process
that exits reporting neither an `exit_code` nor a `signal` as a success — the fallthrough
synthesizes exit code `0`. As a result, a build step with an indeterminate outcome can be
reported as successful and a potentially incomplete rootfs image published.

This change makes the terminal-state handling fail closed. A small `classify_process_exit`
helper handles the `(exit_code, signal)` cases explicitly and returns an error for the
indeterminate `(None, None)` case; existing exit-code precedence and signal handling are
preserved.

## Validation

Unit tests cover a zero exit code, a non-zero exit code, signal termination, and the
indeterminate case.

Related: #528

### DIFF
--- a/crates/cloud-sdk/src/sandbox_images.rs
+++ b/crates/cloud-sdk/src/sandbox_images.rs
@@ -1162,22 +1162,25 @@ async fn run_streaming_process(
         info = proxy.get_process(pid).await?.into_inner();
     }
 
-    let exit_code = if let Some(code) = info.exit_code {
-        code
-    } else if let Some(signal) = info.signal {
-        -signal
-    } else {
-        0
-    };
+    classify_process_exit(command, info.exit_code, info.signal)
+}
 
-    if exit_code != 0 {
-        return Err(SandboxImageBuildError::other(format!(
+fn classify_process_exit(command: &str, exit_code: Option<i64>, signal: Option<i64>) -> Result<()> {
+    match (exit_code, signal) {
+        (Some(0), _) => Ok(()),
+        (Some(code), _) => Err(SandboxImageBuildError::other(format!(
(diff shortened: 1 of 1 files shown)


crates/cloud-sdk/src/sandbox_images.rs
```diff
@@ -1162,22 +1162,25 @@ async fn run_streaming_process(
...
             "Command '{}' failed with exit code {}",
-            command, exit_code
-        )));
+            command, code
+        ))),
+        (None, Some(signal)) => Err(SandboxImageBuildError::other(format!(
+            "Command '{}' was terminated by signal {}",
+            command, signal
+        ))),
+        (None, None) => Err(SandboxImageBuildError::other(format!(
```
thedebugger, line 1179: do you know when this happens?

no-unwrap: Honestly, I haven't been able to reproduce the exact server-side trigger. It happens when a process shows as non-`running` but both `exit_code` and `signal` are missing, and the Python and TypeScript SDKs treat these as optional too. My best guess is a timing gap, a propagation issue, or a dropped exit record, but you'd know more about when the daemon can send this state.

I treated it as an error instead of defaulting to `0`. Since the client can't tell "not reported yet" from "lost," failing a retryable build is safer than publishing a partially built rootfs as successful. Open to other ideas if you have a different approach. If it helps, I can add a test for the indeterminate case (using `0` and `1` as controls), and change the poll budget if there's a known limit for how long a status delay is acceptable.

PR #724, opened 2026-06-06
fix(cloud-sdk): classify malformed invoke success body as decode error

jare1686:
## Summary

`ApplicationsClient::invoke` (`crates/cloud-sdk/src/applications/mod.rs`) decodes the response
as a raw `serde_json::Value` and returns `SdkError::ServerError { status: 200 OK }` when
`request_id` is missing or not a string. That status is contradictory, since `execute_json`
only returns `Ok` on a successful (2xx) response.

This change decodes the success body into a typed (internal) struct, so a malformed body
surfaces as a regular decode error rather than a fabricated server error. It brings `invoke`
in line with the other typed client methods, removes the manual `Value` indexing, and adds no
public API surface or new error variant.

## Validation

Tests cover a present, missing, and non-string `request_id`.

Related: #528

### DIFF
--- a/crates/cloud-sdk/src/applications/mod.rs
+++ b/crates/cloud-sdk/src/applications/mod.rs
@@ -134,15 +134,12 @@ impl ApplicationsClient {
             .header(ACCEPT, "application/json")
             .json(&request.body)
             .build()?;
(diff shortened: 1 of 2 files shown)


thedebugger (approved):

thedebugger:
Thanks for the fix :)

PR #727, opened 2026-06-06
fix(cloud-sdk): reclaim builder sandbox on cancellation

jare1686:
## Summary

`build_sandbox_image` (`crates/cloud-sdk/src/sandbox_images.rs`) creates a remote builder
sandbox and reclaims it with a sequential `delete` after the build's awaited block. If the
future is cancelled — a caller timeout, a losing `select!` branch, or client shutdown — during
the build, that delete never runs and the builder sandbox is leaked until the server-side
timeout reaps it. The keepalive task compounds this: it is aborted only on the success path, so
on cancellation its `JoinHandle` is dropped and detached, leaving a background loop pinging the
orphaned sandbox.

This change ties cleanup to ownership. A small `BuilderSandboxCleanup` guard captures the
runtime handle at creation and reclaims the sandbox via a spawned (detached) delete, so
cancellation before or during cleanup cannot suppress it; the normal path still awaits the
delete and emits the existing warning, and a `404` is treated as success. The keepalive handle
is wrapped so it aborts on drop. No new dependencies and no public API surface.

This is best-effort SDK-side cleanup once the sandbox ID is known; guaranteed reclamation across
process exit or runtime shutdown is a server-side concern (a sandbox lease/TTL or a reaper) and
is out of scope here.

## Validation

Tests cover delete-on-drop, awaited reclaim (no double delete), cancellation during cleanup,
404-as-success, non-404 error propagation, successful delete, and keepalive abort-on-drop.

Related: #528

### DIFF
--- a/crates/cloud-sdk/src/sandbox_images.rs
+++ b/crates/cloud-sdk/src/sandbox_images.rs
@@ -337,6 +337,7 @@ where
         .await?;
     let sandbox_id = created.sandbox_id.clone();
     let routing_hint = created.routing_hint.clone();
+    let cleanup = BuilderSandboxCleanup::arm(sandboxes.clone(), sandbox_id.clone());
 
     let result = async {
         let running_info = wait_for_sandbox_status(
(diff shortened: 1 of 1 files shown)


diptanu:
@jare1686 Thanks for the contribution! Can you please update the PRs so that lint passes and rebase from main?

no-unwrap:
Done! My pleasure.

PR #731, opened 2026-06-09
[codex] Ensure function image executor script is on PATH

calavera:
## Summary

- install the final Tensorlake runtime layer for function images into `/usr/local` so `function-executor` lands at `/usr/local/bin/function-executor`
- fail function image builds unless that exact console script exists and is executable
- keep the generated install step compatible with the sandbox image builder by using `sudo` inside the final `RUN` when the Docker build user is not root, instead of adding a rejected `USER root` instruction
- add SDK/Rust coverage for the generated function image Dockerfile contract
- hide the legacy `tl build-images` subcommand from CLI help while keeping the command parseable
- bump SDK/CLI package versions to `0.5.36`

## Why

Function images run `function-executor` through the Firecracker dataplane as `root`, with `/usr/local/bin` on the guest process `PATH`. A pip user install can place console scripts under `/home/tl-user/.local/bin`, which may be visible during image build but is not a reliable runtime contract for the root-launched function executor. The final Tensorlake install layer now forces a system-prefix install and verifies the runtime script location directly.

## Validation

- `PYTHONPATH=src python3 -m pytest tests/image/test_sandbox_builder.py -q`
- `cargo test -p tensorlake images::models::tests::test_dockerfile_installs_sdk_with_python3_module_pip`
- `cargo test -p tensorlake-cli build_images_command_is_hidden_from_help`
- `cargo run -q -p tensorlake-cli --bin tl -- --version`
- `cargo run -q -p tensorlake-cli --bin tl -- --help | rg -n "build-images|Commands:" || true`
- `git diff --check`

Note: `black` was not available in the local Python 3.14 environment, so Python formatting could not be run locally.

### DIFF
--- a/src/tensorlake/image/utils.py
+++ b/src/tensorlake/image/utils.py
@@ -35,10 +35,18 @@ def dockerfile_content(img: Image, extra_env_vars: List[tuple] | None = None) ->
     # Run Tensorlake install after all user commands so user layers cannot
     # remove or downgrade the runtime. Force reinstall makes pip replay package
     # console scripts even when Tensorlake was already present in the base.
+    # Install into /usr/local so function-executor is visible to the root
+    # process launched by the Firecracker dataplane.
+    install_cmd = (
+        "python3 -m pip install --break-system-packages --force-reinstall "
+        f"--no-cache-dir --prefix=/usr/local tensorlake=={_SDK_VERSION}"
+    )
     dockerfile_lines.append(
-        "RUN python3 -m pip install --break-system-packages "
-        f"--force-reinstall --no-cache-dir tensorlake=={_SDK_VERSION} "
-        "&& command -v function-executor"
+        'RUN if [ "$(id -u)" = "0" ]; then '
+        f"PIP_USER=false {install_cmd}; "
+        "else "
+        f"sudo -E env PIP_USER=false {install_cmd}; "
+        "fi && test -x /usr/local/bin/function-executor"
     )
 
     return "\n".join(dockerfile_lines)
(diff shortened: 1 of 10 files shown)


sgirones (approved):

src/tensorlake/image/utils.py
```diff
@@ -32,10 +32,21 @@ def dockerfile_content(img: Image, extra_env_vars: List[tuple] | None = None) ->
...
+        "python3 -m pip install --break-system-packages --force-reinstall "
+        f"--no-cache-dir --prefix=/usr/local tensorlake=={_SDK_VERSION}"
+    )
     dockerfile_lines.append(
-        f"RUN python3 -m pip install --break-system-packages tensorlake=={_SDK_VERSION}"
+        'RUN if [ "$(id -u)" = "0" ]; then '
+        f"PIP_USER=false {install_cmd}; "
+        "else "
+        f"sudo -E env PIP_USER=false {install_cmd}; "
+        "fi && test -x /usr/local/bin/function-executor"
```
eabatalov, line 49: This looks like a complex workaround script for me, not sure how reliable it is. Can't we just install tensorlake package globally like we did in the past? If we need to run FE as root then we can just run pip install as root? sudo or something.

calavera: the reason why this is more than just `sudo pip install` is because we have a combination of internal and external images as base. Internal images, like `tensorlake/ubuntu-minimal`, will run this as root. However, that's not guaranteed which user is when we're using an external image, like docker's `python/slim`. This is a safer approach to ensure the package is always correctly installed no matter who's installing it.

I've tested creating images with both external and internal bases, and it works correctly in either case, we get the function-executor script where we expect.

eabatalov (approved):

Issue #697, opened 2026-05-25
Support all Dockerfile commands missing for Harbour integration in Sandbox image builds


eabatalov:
Support Dockerfile commands missing for Harbour integration and more.

All Dockerfiles in https://github.com/harbor-framework/terminal-bench-2/tree/main has to work.

Plus need to support all extended filesystem metadata like acls, apparmor, etc.

Also user reported ENV command not working.

eabatalov:
Harbor repo Dockerfiles analysis:

```
Here are the exact counts across all 89 Dockerfiles in harbor-framework/terminal-bench-2:

  ┌─────────────┬───────┐
  │ Instruction │ Count │
  ├─────────────┼───────┤
  │ RUN         │ 171   │
  ├─────────────┼───────┤
  │ COPY        │ 100   │
  ├─────────────┼───────┤
  │ WORKDIR     │ 96    │
  ├─────────────┼───────┤
  │ FROM        │ 90    │
  ├─────────────┼───────┤
  │ ENV         │ 6     │
  ├─────────────┼───────┤
  │ CMD         │ 1     │
  ├─────────────┼───────┤
  │ ARG         │ 1     │
  └─────────────┴───────┘

  Notes for sizing your sandbox image builder:

  - Only 7 distinct instructions are used across the entire corpus.
  - 88 of 89 Dockerfiles are single-stage; only financial-document-processor/environment/Dockerfile is multi-stage (2× FROM), which is why FROM totals 90.
  - Not used anywhere: ENTRYPOINT, EXPOSE, USER, LABEL, ADD, VOLUME, HEALTHCHECK, SHELL, STOPSIGNAL, ONBUILD, MAINTAINER.
  - The long tail (ENV, CMD, ARG) is tiny — supporting just FROM, RUN, COPY, WORKDIR covers ~98% of all instructions (457/465).

  Cleanup: the temporary clone is at /tmp/tb2 if you want to spot-check.
```

There's multi-stage Docker build usage but it only uses public Docker registry images so I think `docker build` should handle it smoothly. The only case where multy-stage docker builds won't work is if an already built Sandbox image is used in FROM command. This will fail at the moment because we don't run a Docker/OCI registry for Sandbox images. Also our Dockerfile FROM command replacement code seems to be a bit strict, so I'll be looking into that too.

eabatalov:
Huzzle only uses WORKDIR command

eabatalov:
[...]

PR #737, opened 2026-06-12
Don't inject tl-user; send process user only when explicitly requested

eabatalov:
## Problem

The sandbox Process-execution APIs (Python SDK, TypeScript SDK, and CLI) hardcoded a default process user of `tl-user` and **always** serialized it into the request. On arbitrary OCI/Docker images (built from external bases like `ubuntu:24.04`, `python:*-slim`) there is no `tl-user`, so every `run` / `start_process` / `tl sbx exec` failed with:

```
HTTP 500 {"error":"user not found: tl-user"}
```

This was found during OCI/Docker runtime-compatibility validation for #697. Confirmed empirically that it's a **client-side** default, not the daemon's: when the `user` field is omitted, the daemon resolves the image's configured user correctly (image `USER`, falling back to `root`) and honors `WORKDIR`. The File API already omitted the field, which is why it was never affected.

## Fix

Send `user` **only when the caller explicitly provides one**; otherwise omit it and let the daemon resolve the image default.

- **Python SDK** (`sandbox.py`, `async_sandbox.py`): `run`/`start_process` default `user=None`; `_normalize_process_user` returns `None` for `None` (so `_build_command_payload` omits it). Docstrings updated.
- **TypeScript SDK** (`typescript/src/sandbox.ts`): `processUserPayload` returns `undefined`; `run`/`startProcess` set `body.user` only when defined; removed `DEFAULT_PROCESS_USER`.
- **CLI** (`crates/cli/src/main.rs`, `exec.rs`): `tl sbx exec` / `tl sbx run` `--user` is now `Option<String>`, omitted by default; added an empty-`--user` guard for parity with the SDKs.

The Rust core (`cloud-sdk`) and napi bindings pass the payload verbatim — no change. PTY/`tl sbx ssh` already send no user.

## Verification

- Unit tests updated to assert **omission by default** and pass: Python 66, Rust CLI 162, TS sandbox 28. TS SDK builds.
- **Managed image (no regression):** with `user` omitted, default `tensorlake/ubuntu-minimal` still resolves to `tl-user` (`uid=1000(tl-user)`).
- **Arbitrary OCI image (the fix):** with `user` omitted, resolves to `root` and honors `WORKDIR` — previously `500 user not found: tl-user`.

## Versioning

Bumped `0.5.39 → 0.5.40` across the CLI and all three SDKs (`bump_version.py` for Python/Rust/CLI + `typescript/package.json`); `Cargo.lock` / `package-lock.json` refreshed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2981,7 +2981,7 @@ dependencies = [
 
 [[package]]
 name = "tensorlake"
-version = "0.5.39"
+version = "0.5.40"
 dependencies = [
  "base64",
  "bollard",
@@ -3024,7 +3024,7 @@ dependencies = [
 
 [[package]]
 name = "tensorlake-cli"
-version = "0.5.39"
+version = "0.5.40"
 dependencies = [
  "anyhow",
  "base64",
@@ -3067,7 +3067,7 @@ dependencies = [
 
 [[package]]
 name = "tensorlake-rust-cloud-sdk-node"
-version = "0.5.39"
+version = "0.5.40"
 dependencies = [
  "napi",
  "napi-build",
@@ -3081,7 +3081,7 @@ dependencies = [
 
 [[package]]
 name = "tensorlake-rust-cloud-sdk-py"
-version = "0.5.39"
+version = "0.5.40"
 dependencies = [
  "futures",
  "pyo3",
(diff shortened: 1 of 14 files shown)


eabatalov:
The CI-failures are pre-existing and not related to this PR. I run Sandbox tests in prod with this PR SDK and all good: https://github.com/tensorlakeai/indexify-deployment/actions/runs/27420861813/job/81045777307

PR #747, opened 2026-06-15
[codex] Add GPU sandbox client support

thedebugger:
## Summary

Adds internal GPU sandbox creation support across the Tensorlake clients:

- CLI: hidden `tl sbx create/new --gpus <count>` and `--gpu-model <model>` flags
- Rust cloud SDK request model: serializes `resources.gpus` as an array of `{ count, model }`
- Python SDK: `gpus=` and `gpu_model=` on sync, async, and `Sandbox.create`
- TypeScript SDK: `gpus` and `gpuModel`

For this first cut, GPU resources are only sent when a GPU count is explicitly provided. The model defaults to `A10`, and non-A10 model requests are rejected when GPU resources are requested.

## Validation

Local checks:

- `cargo test -p tensorlake-cli gpu -- --nocapture`
- `PYTHONPYCACHEPREFIX=/tmp/tensorlake-gpu-pycache poetry run python -m unittest tests.sandbox.test_client_rust_backend`
- `npm -C typescript test -- client.test.ts`

Earlier live dev smoke on this PR:

- CLI created GPU sandbox, ran `nvidia-smi -L`, then terminated it
- Python SDK created GPU sandbox, ran `nvidia-smi -L`, then terminated it
- TypeScript SDK created GPU sandbox, ran `nvidia-smi -L`, then terminated it

All three live sandboxes reported `GPU 0: NVIDIA A10`.


src/tensorlake/sandbox/sandbox.py
```diff
@@ -240,6 +242,9 @@ def create(
             memory_mb: Memory in megabytes.
             disk_mb: Root disk size in megabytes. When omitted, the server
                 uses its default disk size.
+            gpus: Number of GPUs to allocate. Must be provided with
+                ``gpu_model``.
+            gpu_model: GPU model to allocate, such as ``A10``.
```
eabatalov: Would probably be helpful to have a list of exactly supported GPU models so users don't have to guess what to put here. Also presence of gpu_model can enabled --gpus 1 by default.

thedebugger: i was thinking of doing that when we are release this feature. this is for us to test while we publicly release it

thedebugger: let me default to A10 for now so user don't need to pass it

crates/cli/src/main.rs
```diff
@@ -410,6 +424,14 @@ enum SbxCommands {
         #[arg(long = "disk_mb")]
         disk_mb: Option<u64>,
 
+        /// Number of GPUs to request
+        #[arg(long = "gpus", hide = true, value_parser = clap::value_parser!(u32).range(1..))]
+        gpus: Option<u32>,
+
+        /// GPU model to request
+        #[arg(long = "gpu-model", value_enum, hide = true)]
```
calavera: add a `requires` attribute here so we fail if people set `gpu-model` but not `gpus`, example:
```suggestion
        #[arg(long = "gpu-model", value_enum, hide = true, requires = "gpus")]
```

Or a different alternative would be to set `gpus` to 1 if the model is set, but there is no number.

calavera (approved):

crates/cli/src/main.rs
thedebugger: fixed

PR #754, opened 2026-06-16
[codex] Return function executor init errors

calavera:
## Summary
- add `InitializeResponse.error_message` to the function-executor proto
- return the caught initialization exception type and message when user function loading fails
- assert the FE initialize RPC exposes constructor/import errors to callers

## Root Cause
The function executor caught initialization exceptions and returned only `INITIALIZATION_FAILURE_REASON_FUNCTION_ERROR`, dropping the original exception message at the RPC boundary. Downstream services could classify the failure but could not show users what to fix.

## Validation
- `poetry run python -m unittest test_run_allocation.TestRunAllocation.test_function_initialization_raises_error` from `tests/function_executor`
- `python3 -m py_compile tests/function_executor/test_run_allocation.py`
- `git diff --check`

## Notes
This changes the function-executor proto surface, so the SDK package version should be bumped before release.


src/tensorlake/function_executor/proto/function_executor.proto
```diff
@@ -99,6 +99,7 @@ enum InitializationFailureReason {
 message InitializeResponse {
     optional InitializationOutcomeCode outcome_code = 1;
     optional InitializationFailureReason failure_reason = 2;
+    optional string error_message = 3;
```
eabatalov, line 105: Nit, would be nice to at least add a comment that this is a customer visible and human readable error message to establish this contract.

Or otherwise error message depends on failure reason if it's INITIALIZATION_FAILURE_REASON_FUNCTION_ERROR then it's human readable and can be shown to customer.

eabatalov (approved):

src/tensorlake/function_executor/proto/function_executor.proto
calavera: fixed. Thanks!

PR #774, opened 2026-06-26
Provision tl-user in sandbox images built from upstream bases

cooleel:
## Changes

Inject a best-effort `RUN` into generated sandbox Dockerfiles that provision the `tl-user` account when it's missing from the base image.

## Why

Sandbox commands run as `tl-user`. Images built from upstream bases (e.g. `python:3.11-slim`) don't ship that account, so commands failed at runtime with `user not found: tl-user`. Tensorlake's own bases already include it, which is why this only surfaced with custom bases.

## How

`ENSURE_RUNTIME_USER_COMMAND` runs after `FROM`:

- no-op if `tl-user` already exists (Tensorlake bases)
- `useradd` for glibc bases (Debian/Ubuntu/…)
- `adduser -D` fallback for Alpine/BusyBox
- trailing `|| true` so the build never aborts

Kept byte-identical between the Python and TypeScript SDKs.

## Notes

- Bumps Python SDK to 0.5.51 and TypeScript SDK to 0.5.50.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ members = [
 resolver = "3"
 
 [workspace.package]
-version = "0.5.50"
+version = "0.5.51"
 authors = ["Tensorlake Inc. <support@tensorlake.ai>"]
 edition = "2024"
 license = "Apache-2.0"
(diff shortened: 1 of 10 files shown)


cooleel:
It would be best if we fix it from server side as well. https://github.com/tensorlakeai/compute-engine-internal/issues/1261

eabatalov:
This is incorrect, we don't modify user images

PR #778, opened 2026-06-29
Add file system support across the SDK (cloud-sdk, Python, TypeScript, CLI)

Barre:
Adds file system support across the stack — cloud-sdk, Python, TypeScript, and the CLI (tl fs create/ls/rm, plus sbx fs attach/detach/ls). Lets you create file systems and mount them into sandboxes.

Bumps Python/CLI to 0.5.52 and the TS package to 0.5.51.


crates/cli/src/commands/sbx/fs.rs
```diff
@@ -0,0 +1,142 @@
...
+            "failed to {action} (HTTP {}): {}",
+            status,
+            body
+        )));
+    }
+    resp.json().await.map_err(CliError::Http)
+}
+
+/// Render the sandbox's currently-mounted file systems as a table.
+fn print_mounts_table(info: &serde_json::Value) {
```
eabatalov: One thing I see in this PR is that we're using &serde_json::Value instead of a model object. Can we use a SandboxSharedFSMount or SandboxSharedFSMountInfo or some other model object to pass mount information around?

crates/cli/src/commands/fs.rs
```diff
@@ -0,0 +1,116 @@
...
+        .ok_or_else(|| CliError::auth("Project ID is required to manage file systems"))?;
+    Ok((org_id, proj_id))
+}
+
+/// Build a cloud SDK client scoped to the current org/project.
+///
+/// Mirrors `commands::sbx::image::scoped_cloud_client`: scope headers are only
+/// attached for personal-access-token auth, since project API keys are already
+/// implicitly scoped.
+fn scoped_cloud_client(ctx: &CliContext) -> Result<Client> {
```
eabatalov: This seems like a very generic helper, why is it in FS specific file?

crates/cli/src/commands/fs.rs
```diff
@@ -0,0 +1,116 @@
+//! `tl fs` — manage the project-scoped ZeroFS file-system registry.
```
eabatalov: I'd suggest to not mention ZeroFS anywhere because implementation doesn't matter at SDK level.

And rename every FS -> SharedFS or something like that because there's Sandbox local ephemeral FS too, just to disambiguate entities used in source code

src/tensorlake/sandbox/file_system.py
```diff
@@ -0,0 +1,167 @@
...
+def _debug_enabled() -> bool:
+    return os.environ.get("TENSORLAKE_DEBUG", "").lower() in {
+        "1",
+        "true",
+        "yes",
+        "on",
+    }
+
+
+def _build_context_from_env() -> Context:
```
eabatalov: This is a copy-paste of generic helpers. Please remove it and dedup.

eabatalov (approved):

eabatalov (approved):

PR #795, opened 2026-07-03
tl fs: versioned filesystem workspaces on artifact storage

diptanu:
Client half of [artifact_storage#24](https://github.com/tensorlakeai/artifact_storage/issues/24). `tl fs` speaks artifact-storage workspaces end to end — mounting is kernel-native on both platforms: **FUSE on Linux, FSKit on macOS** (no kernel extension, no macFUSE, no sudo), and no materialized-directory fallback by design.

## Product surface

```bash
tl fs create --name my-fs          # an artifact-storage repo, seeded so it mounts immediately
tl fs mount my-fs ./work           # kernel mount: lazy reads, local overlay for writes
# ...work in the mount...
tl fs snapshot ./work -m "wip"     # the overlay IS the dirty set → chunk upload → ref advances
tl fs promote ./work main          # squash by default; --full-history opts into the chain
tl fs restore ./work <snapshot>    # differential overlay refill from any snapshot
tl fs status / diff / pin / unmount / workspaces
```

Registry (`create/ls/rm`) routes to artifact-storage repos instead of the platform file-systems API; `tl sbx fs` untouched. Windows builds keep the portable commands and fail mounts immediately with a clear unsupported error.

## Architecture

One Rust brain, thin per-platform kernel bindings:

- **`crates/gsvc-mount` vendored** from artifact_storage (read-only mount core: native-API client, immutable commit-keyed caches, ref-following with open-handle pinning, counted inode table), including the credential-rotation API from artifact_storage#34 — a live mount can't rebuild its client, so the daemon re-mints and swaps the token in place.
- **`OverlayFs`** (VFS-agnostic, kernel-free-testable): upper dir + whiteout-marker tree over the core. Copy-up on first write; merged readdir; whiteouts hide whole subtrees; lookups always re-resolve the lower backing so a snapshot's ref advance reaches new lookups while kernel-held inos keep their commit; lower reads absorb post-snapshot `IndexNotReady` with a bounded settle-poll instead of surfacing EAGAIN.
- **Per-mount daemon**: mount session, ref watcher, 20-minute lease heartbeat, credential rotation, and a line-JSON control socket (`ping`/`refresh`/`clear_upper`/`shutdown`). `tl fs mount` spawns it detached; failed startup deletes the orphan workspace.
- **Linux**: `fuser` glue in-process (pure-Rust `/dev/fuse`), compiled unconditionally.
- **macOS**: production **FSKit** module (`platform/macos/tlfs`, Swift `FSUnaryFileSystem`) — the sandboxed appex can't read daemon state or hold credentials, so it proxies a small VFS wire protocol to the daemon over localhost TCP (`tlfs://127.0.0.1:<port>/<secret>`, per-mount getentropy secret, HELLO-gated). Wire protocol + server live in `vfsserver.rs`; packaging/registration mechanics are documented in `platform/macos/tlfs/README.md` and the validated `fskit-hello` prototype.
- **Snapshot = enumerate the overlay** (`upper/` + `wh/`): no scanning, no manifests, no mtime heuristics. After publish: follow the ref, clear the overlay — the merged view is unchanged, now lower-backed. Restore reads everything from the server first (does no harm on failure), then refills the overlay differentially; the workspace ref (history) is untouched.

## Kernel cache coherence (macOS)

Restore/snapshot mutate the overlay behind the kernel's back and FSKit has no invalidation-push API, so coherence is engineered from what the kernel honors (measured on macOS 26.5 lifs): real content mtimes on the wire (upper = disk mtime, lower = first-seen time of the pinned commit), directory verifiers derived from the dir's mtime, real `itemID`s on every packed entry, and a bounded `converge_kernel_view` pass after restore that re-opens changed paths (close-to-open revalidation), purges cached pages via `msync(MS_INVALIDATE)`, and breaks pinned negative dentries with an `EEXIST` probe. Result: restore returns in ~0.1s with open/read and listings fully coherent; details and residual caveats in the module README.

## Validation

- **Overlay integration test** (ignored-by-default, runs against a local `gsvc-server`): merged reads, copy-up, whiteouts, merged readdir, rmdir empty-check, rename, and the full snapshot round-trip through the workspace ref.
- **Production end-to-end** against api.tensorlake.ai with real PAT→mint auth: create → FSKit mount (1.6s) → kernel writes → snapshot → immediately coherent listings → cross-commit restores (~1.2s, both directions) → squash promote → fresh-mount read-back → clean unmounts. Prod surfaced the 425/index-latency windows now handled by restore's poll-out and the overlay's settle-poll.
- **Cross-platform**: `cargo check` clean for `aarch64-apple-darwin`, `x86_64-unknown-linux-gnu`, and `x86_64-pc-windows-gnu` (cfg-equivalent to the shipped msvc target); SDK crates unaffected; 196 unit tests pass.
- **Remaining:**
  - Kernel-mount smoke on Linux (this workstation has no FUSE runtime); suggest a CI job running the overlay test plus a real mount/snapshot/promote loop before merge.
  - `tl fs rm` is denied in prod — minted git credentials can't delete repos (fail-closed); needs a server-side scope decision or routing through the platform API.
  - Suggested: a `cargo check --target x86_64-pc-windows-msvc` CI job so mount-side changes can't silently break the Windows CLI.


diptanu:
Pushed a performance follow-up to this branch: minted-credential caching under ~/.config/tensorlake (PAT-store convention), single-read snapshot hashing (blob oid computed in the CDC pass; new files skip the pre-hash), parallel per-file chunking + 3-in-flight dedup uploads, a symlink content-correctness fix (link text was uploading as the target's bytes), and a 425-retry on promote-right-after-snapshot. Re-smoked end-to-end against a local server: touch-no-change stays clean, 24MB binary byte-identical through promote+clone, readlink-verified symlinks, dedup observed.

diptanu:
Since the description was last refreshed, this branch gained four more arcs (all validated live against a local server, most also against prod):

- **Async commit client** (`tl git push` + `tl fs snapshot`): every commit submits as a durable server job — 201 inline within the grace window, 202 + poll past it, with live server-side progress (`committing: read_back 61440/94210 chunks...`), a printed job id on detach, and `tl git commit-status --repo <r> <job-id>` for out-of-band checks. Strict job-shape decode (no legacy tolerance); the client still sends the opt-in header until every environment runs the job-only server.
- **Presigned-PUT staging on push**: dedup chunks are zstd-framed into client-assembled packs and PUT straight to S3; service sees only JSON control requests. Falls back to chunk frames for non-presigning backends.
- **Durable workspaces**: no leases, no pin — a workspace lives until `tl fs workspace rm`. `unmount` keeps it by default and prints the reattach command; `tl fs mount <fs> --workspace <id>` resumes at the last snapshot; `workspace ls/rm` group replaces top-level `workspaces`/`pin`.
- **Mount modes** (issue #24 trio complete): `--shared-ro` (workspace read session following the branch; all writes EROFS) and `--shared-rw` (snapshots auto-publish to the branch, server-ordered). Fixed a staleness bug this surfaced that also affected writable mounts: path-level ops now re-bind their lower when the followed ref advances (generation-checked), so followed branches actually move; open handles keep their commit.

Server counterparts are all on artifact_storage main (#38–#40) plus the small rescue PR #46.

diptanu:
Pushed `bb15fb2`: client adoption of artifact-storage #57 (tensorlakeai/artifact_storage#58).

- Every mostly-missing file now uploads whole under a file token (8 MiB floor removed) — staged server-side at upload, zero commit read-back; tokened uploads run 4 files in flight and identical content shares one token.
- Chunks referenced by dedup-path files always upload via the dedup path (staging servers don't register small tokened files' chunks).
- New `PushSource::KnownOid`: commit repeat content by blob oid, no bytes in the request; gated on the session's `oid_files` capability so older servers (which would silently write an empty file) fail fast instead.
- Validated live against a server built from artifact_storage PR #58: new `push_paths_roundtrip_against_local_server` SDK integration test + the overlay workspace suite with the vendored mount core; full workspace tests green.

diptanu:
Pushed the batch-endpoint adoption: small files now upload via `PUT .../ingest/sessions/{id}/files` (16 MiB groups, 4 in flight, per-file oid verification), gated on the session's `batch_files` capability with the per-file path as fallback. Pairs with artifact_storage PR #58's second push; server A/B for 4k×24KiB files: upload wall 11.2s → 463ms, commit 875ms → 331ms.

diptanu:
Pushed `4ffa86f`: small files now hash whole (no CDC), skip `missing` negotiation, and commit token-only (`{path, file_token, oid}` — no chunk lists) against servers advertising `token_commits`. Pairs with artifact_storage PR #58 third push (batched claim reads + parallel deflate, token-only commit support, stored-block fallback).

diptanu:
Pushed auto-retry for idempotent ingest requests: transient failures (transport, 5xx/429) back off and retry across session create / negotiation / staging / presigned PUTs / untokened frames / batch uploads / commit submit (idempotency-key reattach) / job polls. Large tokened offset uploads stay single-shot by design (a lost success would conflict on replay). Paused-clock unit tests + live validation.

diptanu:
Pushed `c340589`: gsvc-codec is no longer vendored into this tree — it follows the gsvc-mount arrangement now (resolution-only placeholder + ephemeral vendor). `just build-cli-full` vendors both private crates and builds `--features mount,git-clone`; CI action renamed to `vendor-private-crates` and publish workflows vendor on every target. The cloud SDK dropped its codec dependency entirely (blob-oid hashing is a local helper), so push/ingest and all bindings build from the public tree standalone; only `tl git clone`'s fast-clone engine is feature-gated. artifact_storage AGENTS.md updated with the new sync duties (placeholder manifest on dep changes, companion PR only for wire-format changes).

PR #800, opened 2026-07-05
feat(documentai): support BytesIO uploads

### LINKED ISSUE #274: Support file upload with bytes.IO
### Description

Users should not be forced to write a file to the local filesystem to upload it to the Files API. This issue is to extend the upload method to accept a `bytes.IO` parameter and use it as the multipart form data into the Tensorlake Upload API.

tlysanhuo:
## Summary
Fixes #274.

- accept in-memory `io.BytesIO` buffers in DocumentAI `upload` / `upload_async`
- preserve the current buffer position by reading with `getvalue()`
- use the `BytesIO.name` basename when present, otherwise `upload.bin`
- reuse the existing Rust upload response handling through a shared helper
- add fake-Rust tests for named and unnamed `BytesIO` uploads

## Testing
- `python -m py_compile src\tensorlake\documentai\files.py src\tensorlake\documentai\_files.py tests\document_ai\test_rust_backend.py`
- local smoke script for named/unnamed `BytesIO` upload behavior
- `PYTHONPATH=src python -m unittest tests.document_ai.test_rust_backend` (does not run successfully in this checkout because the package is not installed, so top-level import fails with `PackageNotFoundError: No package metadata was found for tensorlake` before test collection)

### DIFF
--- a/src/tensorlake/documentai/_files.py
+++ b/src/tensorlake/documentai/_files.py
@@ -4,6 +4,7 @@ File-management helpers (upload, list, delete).
 
 from __future__ import annotations
 
+from io import BytesIO
 from pathlib import Path
 from typing import Optional, Union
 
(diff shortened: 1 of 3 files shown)


diptanu:
We have sunsetted document ai

PR #818, opened 2026-07-09
tl fs mount --auto-commit-interval-secs: event-driven async snapshots with an append fast path

diptanu:
## What

`tl fs mount <target> <path> --auto-commit-interval-secs <N>` makes the mount daemon seal local changes into workspace snapshot commits every N seconds, asynchronously. `tl fs snapshot` stays the on-demand seal-and-clear. Read-only mounts reject the flag; `tl fs status` and the mount output surface it.

## How it stays cheap

**Event-driven dirty index (no scanning).** Every `OverlayFs` mutation records `path → (generation, kind, min written offset)` inline — the daemon *is* the write path, so dirtiness is never re-discovered by walking. An idle tick is one atomic load. The index rebuilds from disk at daemon start and after `tl fs restore` (new `reindex` control op).

**Incremental seals.** The workspace ref advances with each snapshot, so each commit carries only paths touched since the last sealed generation. Unchanged dirty files are served by the lower and never re-hashed or re-sent. The overlay is deliberately not cleared: clear-after-push is only safe with writes quiesced (racing writes would drop; open write handles would orphan) — the upper keeps shadowing byte-identical sealed content.

**Append fast path — O(appended bytes).** New `PushSource::StablePrefix` in the SDK reuses the previous seal's CDC chunk list for the untouched prefix; FastCDC restarts its rolling state at every boundary, so resumed chunking equals a full re-chunk exactly (unit-tested). Dedup-path files never send a client blob oid (identity comes from server read-back), so the prefix is never re-read at all. The upload pass now also seeks past chunks the server already has instead of re-reading them — a win for every mostly-deduped push. The daemon feeds this from a per-path chunk cache returned via `PushOptions::collect_file_chunks`.

## Also fixed along the way

- A resurrection race: a path deleted right after being sealed (before the lower advanced) wrote no whiteout and silently reappeared. The daemon now polls the ref immediately after each push and tombstones vanished-but-recently-sealed paths.
- The daemon's heartbeat held a static git credential that went stale ~1h into a mount's life; heartbeats and auto-commits now share the rotation task's refreshed `(user, token)` slot.

## Measured (local gsvc-server, macOS FSKit, debug build)

| Scenario | Before | After |
|---|---|---|
| Idle tick | tree walk + stat per dirty file | one atomic load |
| 1-byte change, 500 MB dirty | 22.3 s | ~tick-instant |
| Append 1 MB to 500 MB file | full re-hash (~27 s) | seals within a 2 s tick |
| In-mount read+write+ls during a push | — | 22–40 ms (idle baseline 33 ms) |

Correctness verified live: 5 append seals with adversarial sizes (517 B … 3 MB) reconstruct byte-identical content on a fresh mount (`cmp` over 68 MB); deletes propagate; `tl fs snapshot`/`promote`/`restore` interplay intact. 202 tests pass with `--features mount`, 535 across both crates default; new unit tests cover `resolve_seal` routing, the tombstone arm, stable-prefix selection, and CDC resume equivalence.

Companion PR in artifact_storage fixes a pre-existing mount bug this work surfaced (cold sequential reads of large files EIO when the server's rate limiter trips — the vendored `FsClient` had no retry handling); link to follow.

Note for release: version bump ships separately per convention (`python .github/scripts/bump_version.py`) — the `tensorlake` crate gains public SDK surface (`PushSource::StablePrefix`, `PushOptions::collect_file_chunks`, `PushReport::file_chunks`).

### DIFF
--- a/crates/cli/src/commands/fs.rs
+++ b/crates/cli/src/commands/fs.rs
@@ -832,6 +832,7 @@ pub async fn mount(
     path: &Path,
     mode: WritePolicy,
     shared_rw: bool,
+    auto_commit_interval_secs: Option<u64>,
     foreground: bool,
 ) -> Result<()> {
     // Bail before creating a workspace or spawning the daemon-wait loop.
@@ -851,6 +852,11 @@ pub async fn mount(
     if shared_rw && mode == WritePolicy::Ro {
         return Err(CliError::usage(
             "--shared-rw publishes every snapshot; it cannot be combined with --mode ro",
+        ));
+    }
+    if auto_commit_interval_secs.is_some() && mode == WritePolicy::Ro {
+        return Err(CliError::usage(
+            "--auto-commit-interval-secs seals local writes; a read-only mount has none",
         ));
     }
     if shared_rw && base.is_none() {
@@ -1008,6 +1014,15 @@ pub async fn mount(
         }
     };
 
+    // An attach can resolve read-only implicitly (the workspace is live-mounted elsewhere);
+    // nothing server-side was created on that path, so erroring here leaks nothing.
+    if auto_commit_interval_secs.is_some() && read_only {
+        return Err(CliError::usage(
+            "--auto-commit-interval-secs needs a writable mount; this workspace is already \
+             mounted elsewhere and attached read-only (pass --mode rw to take writes)",
+        ));
+    }
+
     let mountpoint = canonical_mountpoint(path)?;
     let state_dir = alloc_state_dir(&ws.id);
     let (owner_uid, owner_gid) = mount_owner();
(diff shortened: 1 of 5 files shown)


diptanu:
Companion fix for the pre-existing cold-read EIO this work surfaced: tensorlakeai/artifact_storage#78 (mount FsClient had no retry handling; the server's admission rate limiter killed sustained streams).

diptanu:
Code review complete (10 confirmed findings across both PRs) and all fixes are now on this branch (52cbf59). Highlights: an overlay epoch counter makes the sealer's caches coherent with out-of-band clears (fixes the stale-chunk-cache chimera and the false-delete race with manual snapshots), directory renames now record per-child events and resolve_seal walks directory upserts (a renamed dir's files were silently dropped from snapshots), the resurrection guard is confirmation-based instead of a 2-deep ring, and generation assignment moved under the dirty lock (a racing write could previously go permanently unsealed). Verified live: renamed-dir children survive a fresh mount; mid-file edit + manual snapshot + appends seal byte-identical. PushSource/PushReport are now #[non_exhaustive] for the published crate.

PR #823, opened 2026-07-10
tl fs: rename committed directories instantly; seal reconciles by oid

diptanu:
## Summary

Implements v1 of tensorlakeai/artifact_storage#88. `rename(2)` of a committed (lower-backed) directory previously failed (`ENOTSUP` after #822; a bare `EIO` before) — breaking `git mv`, refactoring scripts, `shutil.move`, and every other tool that reorganizes directories on a mount. Content addressing makes a directory move the cheapest operation in the system (the subtree oid never changes), so the gap was purely mount-side.

**Design: record at syscall time, reconcile at seal time.**

### Overlay redirect (syscall side)

- A pending-rename table (`destination → true-lower source`) persisted to `redirects.json` in the state dir. `rename(2)` of a committed directory records an entry, whiteouts the source, and moves local state (child copy-ups, deletion markers) under the new name — **metadata only, nothing materializes, unbounded I/O never enters the syscall**.
- Sources are stored in true-lower coordinates (composed at insert), so chains (`mv a b; mv b c`) stay one hop; nested pending renames ride ancestor moves.
- Reads resolve through the remap: lower-binding walks (with a redirect generation on cached bindings), lookup/readdir of destination roots, whiteout shielding, refresh-delta translation into destination inos. Removing or overwriting a destination drops its entry.

### Seal reconciliation (publish side)

- Each pending rename expands to per-file **`PushSource::KnownOid` upserts** — zero bytes read or uploaded; the content is already server-side. Upper copy-ups shadow their lower file; whiteouts inside the moved tree stay deletions. The source delete rides the existing whiteout.
- Wired into both sealers: `tl fs snapshot` (new `expand_redirects` control op; the daemon must be running when renames are pending — publishing the source delete without the destination would lose the subtree) and the daemon's auto-commit loop.
- Entries are consumed only once the lower serves the sealed commit; a reaper heals the crash/lag window by detecting entries whose source vanished and destination materialized (runs each auto-commit tick and on `refresh`).

### UX

- `tl fs status` shows pending renames as `R src -> dst` (and `pending_renames` in `--json`), suppressing the redundant `D src` line.
- `tl fs sync` refuses while renames are pending; `promote` counts them in its dirty warning.

## Follow-up (v2, artifact_storage)

A tree-reference commit op would publish a directory move as **one** tree edit instead of O(files) by-oid rows, and gives jj-style merges a true move signal — tracked in tensorlakeai/artifact_storage#88.

## Testing

- 5 new pure unit tests (remap longest-prefix, chain composition, overwrite drop, whiteout shielding both ways) + the 4 existing inode-rename tests still green; 22 fs-module tests total.
- **Both live-server integration tests pass** against a local `gsvc-server` (open auth), including the new `overlay_renames_committed_directory_and_seals_by_reference`: instant rename → redirect-served reads (content, exec bits, nested dirs) → edits + deletes inside the moved tree → chained re-move → by-oid seal + source delete → ref advance → reap → byte-identical lower-served destination.
- Default no-mount build compiles (placeholder arrangement untouched).

Stacked on #822 (base branch `fix-mount-errnos-and-logging`); rebase onto main after it lands.

### DIFF
--- a/crates/cli/src/commands/fs.rs
+++ b/crates/cli/src/commands/fs.rs
@@ -2109,6 +2109,57 @@ fn enumerate_overlay(state_dir: &Path, mount_root: &Path) -> Result<(OverlayUpse
     Ok((upserts, deletes))
 }
 
+/// Pending committed-directory renames recorded by the mount daemon (`redirects.json` in the
+/// state dir, destination -> true-lower source), sorted by destination. Empty when the file
+/// is absent or unreadable — the daemon owns the authoritative copy.
+fn pending_renames(state_dir: &Path) -> Vec<(String, String)> {
+    let Ok(raw) = std::fs::read(state_dir.join("redirects.json")) else {
+        return Vec::new();
+    };
+    let Ok(map) = serde_json::from_slice::<HashMap<String, String>>(&raw) else {
+        return Vec::new();
+    };
+    let mut entries: Vec<(String, String)> = map.into_iter().collect();
+    entries.sort();
+    entries
+}
+
+/// Expand pending renames through the daemon into by-oid push files, merged behind the
+/// regular dirty set (an upper copy-up under a renamed tree shadows its lower file). The
+/// daemon must be running: publishing the rename's source delete without its destination
+/// upserts would lose the subtree, so a failed expansion fails the seal.
+async fn redirect_push_files(
+    state_dir: &Path,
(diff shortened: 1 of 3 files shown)


diptanu:
Folding into #822 per review preference for a single PR — the branch is now part of that PR's head.

PR #827, opened 2026-07-10
Review fixes for the detached-job poll path: single poll site, long-poll, encoded params, call-site regression test

diptanu:
Applies all client-side findings from the #826 code review.

## Changes

**Single poll site (reuse).** The detached-job poll loop and the pub `commit_job_status` were token-for-token the same request; both now go through a private `commit_job_get`, the only place the `commits/jobs/{id}` route is built. The load-bearing comment (submit route ≠ poll route; the workspace-scoped jobs route does not exist) lives there. The single-use route helpers from #826 are gone.

**`wait_generation` long-poll (efficiency).** The server's `commits/jobs/{id}` route already supports `?wait_generation=&timeout_ms=` (clamped 100–55000 ms) and returns a `generation` counter. The client now:
- deserializes `generation` (new optional `CommitJobStatus` field, `serde(default)` — absent on older servers),
- seeds it from the 202 body so the first poll already long-polls (25s hold, under the 30s request timeout),
- falls back to the previous 500ms→2s backoff for servers without generations or polls that come back unheld (elapsed-time guard so a server that ignores the param is never hammered).

One held request per state transition instead of ~T/2 short polls per job.

**Atomic submit-route/branch decision (simplification).** Restored the single `match` on `workspace_snapshot` deciding `(commit_suffix, branch)` together — the #826 split into two independently-derived expressions is undone.

**Percent-encoded path params (hardening, pre-existing).** `ws_id` and `job_id` now go through `encode_path_segment` like `project`/`repo` always did, closing the query-smuggling hole on the pub `commit_job_status` API.

**Call-site-honest regression test (test-coverage).** The old test asserted helper string outputs and had a tautological `assert_ne!` — reverting the buggy call-site line would have kept it green. The new `detached_snapshot_push_polls_commit_job_route` drives the real `push_files` flow against a local canned-response HTTP server (tokio `TcpListener`, no new dev-deps): known-oid file → session → snapshot submit → 202 detach → poll; it asserts the submit hit `workspaces/{id}/snapshots` and **every** poll hit `commits/jobs/{id}`, with the mock 404ing all other routes exactly like the real server.

## Wire compatibility

`generation` is a new **optional** response field the server already emits — no request shape changed, so no companion artifact_storage PR is required for this. (The remaining review finding — the 202 carrying no `Location`/poll URL — is server-side and lands separately in artifact_storage.)

Verified: `cargo test -p tensorlake --lib` — 135 passed, 1 ignored (live-server test); `cargo check -p tensorlake-cli` clean.

### DIFF
--- a/crates/cloud-sdk/src/artifact_storage/ingest.rs
+++ b/crates/cloud-sdk/src/artifact_storage/ingest.rs
@@ -284,6 +284,10 @@ pub struct CommitJobError {
 pub struct CommitJobStatus {
     pub job_id: String,
     pub state: String,
+    /// Server-side change counter: bumps on every state-machine transition, enabling
+    /// `?wait_generation=` long polls. Absent on older servers.
+    #[serde(default)]
+    pub generation: Option<u64>,
     #[serde(default)]
     pub phase: Option<String>,
     #[serde(default)]
@@ -305,6 +309,11 @@ pub struct CommitJobStatus {
 /// zstd level for staged chunk frames — the server stores frames verbatim, so this matches its
 /// own chunk compression default (speed/ratio balance; any valid zstd frame decodes).
 const STAGED_ZSTD_LEVEL: i32 = 3;
+
+/// Server-side hold for `?wait_generation=` job polls. Kept under the 30s request timeout so a
+/// full-window hold returns normally instead of tripping the client timeout (the server clamps
+/// to 55s; 25s also stays inside typical LB idle limits).
+const JOB_LONG_POLL_TIMEOUT_MS: u64 = 25_000;
 
 #[derive(Serialize)]
 struct CommitFileWire {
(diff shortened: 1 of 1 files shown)


diptanu:
Round-2 review fixes pushed (26d9e92):

- **Long-poll pacing**: an advanced poll now sleeps a 1s floor before re-holding — the executor bumps the job generation on every throttled read-back tick (up to 2/s), so the previous shape re-polled at ~2 Hz during read-back, quadrupling request rate for the phase that dominates large commits. Terminal states still observed with zero added latency.
- **Dropped the 5s dwell heuristic + carried `held` flag**: it timed the whole retry wrapper (client backoff masqueraded as server hold) and defended against a server that never existed (generation-in-body and `wait_generation` shipped in the same server commit). Loop is now sleep-at-bottom.
- **Timeout derives from the hold**: `JOB_LONG_POLL_TIMEOUT_MS + 10s` per attempt, so a full-window hold plus slow delivery can't trip the timeout and fail a push whose job is succeeding (a retried push mints a new idempotency key, so that failure wasn't cleanly resumable).
- **`poll_url` wire mirror**: `CommitJobStatus` now declares the 202's `poll_url` field (inert, `serde(default)`) per the artifact_storage wire-duplication rule — consumption deferred, `commit_job_get` builds the identical route.
- **Branch-route coverage restored**: the mock-server test also drives an ordinary branch push asserting `POST .../commits`.

`cargo test -p tensorlake --lib`: 135 passed. Companion server PR: tensorlakeai/artifact_storage#89.

PR #830, opened 2026-07-10
test: hoist hand-rolled HTTP mock-server framing into shared test support

diptanu:
## Summary

Four places hand-rolled the same minimal HTTP/1.1 mock-server framing (CRLFCRLF header scan + content-length body read + canned JSON response writer):

- `crates/cloud-sdk/tests/sandboxes.rs`
- `crates/cloud-sdk/tests/sandbox_templates.rs`
- `crates/cloud-sdk/tests/images.rs`
- the `detached_snapshot_push_polls_commit_job_route` unit test in `crates/cloud-sdk/src/artifact_storage/ingest.rs` (the only copy with keep-alive connection reuse)

This PR single-sources them in `src/test_support.rs`, compiled twice on purpose from one file: as `#[cfg(test)] crate::test_support` for in-crate unit tests, and re-included via `#[path]` as `common::http_mock` for the integration test binaries. Net −125 lines of test code.

The helper API preserves both connection disciplines:
- `read_http_request` / `write_empty_response` / `write_json_response` — `Connection: close`, so pooled clients open a fresh connection per request (the integration tests accept one connection per expected request and would hang otherwise)
- `read_http_request_keep_alive` / `write_json_response_keep_alive` — carries leftover bytes across requests on one socket and sends no close header, which the ingest detached-job poll test depends on

## Why not wiremock

Evaluated and rejected: these tests assert on the **raw bytes** the client puts on the wire — header casing (`\r\nhost: sandbox-host.test\r\n`), exact framing (`ends_with(b"\r\n\r\nhello")`), and raw multipart body content. wiremock hands you a parsed `Request`, so those wire-level assertions (the point of several of these tests) would be lost.

Intentional, test-invisible behavior notes: the shared reader is content-length-aware everywhere (sandbox_templates' old copy stopped at CRLFCRLF, fine for its bodyless GET/DELETE requests), and the shared 204 writer includes `Content-Length: 0`, which one old inline response omitted.

Test-only change; no runtime code is affected, so no version bump.

## Test plan

- `cargo test -p tensorlake --lib` — 135 passed, including the detached-snapshot regression test
- `cargo test -p tensorlake --test sandboxes --test sandbox_templates --test images` — 5 passed (live-API tests ignored as usual)
- `cargo test -p tensorlake --no-run` — all targets compile with zero warnings

### DIFF
--- a/crates/cloud-sdk/src/artifact_storage/ingest.rs
+++ b/crates/cloud-sdk/src/artifact_storage/ingest.rs
@@ -1688,7 +1688,9 @@ impl ArtifactStorageClient {
         // push can fail against a job that is succeeding.
         let (suffix, timeout) = match wait_generation {
             Some(g) => (
-                format!("commits/jobs/{id}?wait_generation={g}&timeout_ms={JOB_LONG_POLL_TIMEOUT_MS}"),
+                format!(
+                    "commits/jobs/{id}?wait_generation={g}&timeout_ms={JOB_LONG_POLL_TIMEOUT_MS}"
+                ),
                 std::time::Duration::from_millis(JOB_LONG_POLL_TIMEOUT_MS + 10_000),
             ),
             None => (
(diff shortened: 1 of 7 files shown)


diptanu:
Closing as unnecessary test-only abstraction. The four local helpers are small and stable; this adds no coverage or runtime value and saves little code overall. We can revisit shared HTTP test support if the pattern expands materially.

PR #833, opened 2026-07-11
tl fs mount: one-round-trip fast path, credential reuse, phase timings

diptanu:
## Why

`time tl fs mount nomad nomad101` → 5.3s on a sandbox. The daemon.log showed the FUSE attach takes 3ms; the rest was a strictly serial chain of HTTPS round trips across two processes, about half of it redundant:

1. CLI: mint (or cached) → `list_repos` (existence probe) → `ref_status` (seed pre-check) → `create_workspace`
2. Daemon: **a second unconditional platform mint** → `ref_status` + serve-probe of the commit the create response had already named → FUSE attach
3. CLI: readiness poll on a flat 250ms grid

## What

- **One-round-trip fast path**: the CLI attempts `create_workspace` outright and classifies the server's answer for the slow paths — 404 "not found" → try the target as a workspace id (attach), 400 "has no commits" / "does not resolve to a commit" → seed the unborn default branch and retry (writable mounts only; same rules and error messages as before). `list_repos` and the `ref_status` pre-check leave the common path entirely.
- **Credential reuse**: the daemon reads the config-dir credential cache the CLI just wrote (repo scope, then `*`) before minting; the rotation task still re-mints before expiry. Kills the second platform-ingress mint.
- **`MountState.start_oid`** → `MountOptions::start_oid` (new in gsvc-mount): the daemon adopts the create/attach response's commit with zero round trips of its own; the follow poll reconciles a stale start within one interval. Exception: a writable attach of a shared-rw workspace follows the target branch (a ref the response says nothing about) and still resolves itself.
- **Readiness poll ramps 15ms→250ms** instead of a flat 250ms grid.
- **Phase timings**, info level: the CLI logs `phase=session|workspace|daemon … "mount timing"` (+`total_ms`) to stderr; the daemon logs credential source, `mount: core ready`, warmup, and `startup_ms` on the serving line into daemon.log.

Net: daemon startup goes from 3+ serial network calls (incl. a platform mint) to zero before the kernel attach; the CLI chain drops from 4 round trips to ~1–2 (create, plus mint only when the cache is cold).

## Ordering

Requires gsvc-mount with `MountOptions::start_oid`: **merge tensorlakeai/artifact_storage#95 first** — the full CLI build vendors gsvc-mount from that repo's main, so this PR's `mount`-feature build fails until it lands.

## Validation

- `cargo test -p tensorlake-cli --features mount,git-clone` (against artifact_storage#95's gsvc-mount, swapped in via the build-cli-full mechanism): 226 passed
- Default-features build + tests (placeholder crates): 199 passed
- clippy introduces no new warnings; `cargo fmt` clean
- Server-side contract test pinning the 404/400 statuses + messages the classifier reads lands in artifact_storage#95

### DIFF
--- a/crates/cli/src/commands/fs.rs
+++ b/crates/cli/src/commands/fs.rs
@@ -1182,6 +1182,32 @@ fn unzip_app(archive: &Path, staging: &Path) -> Result<PathBuf> {
         .ok_or_else(|| CliError::usage(format!("no .app found inside {}", archive.display())))
 }
 
+/// The two workspace-create failures `tl fs mount` can recover from, read off the server's
+/// answer. Mount tries the create outright (its one required round trip) instead of
+/// pre-flighting with list-repos and ref-status calls; these are the answers that pick the
+/// recovery path. Anything else propagates as-is.
+enum CreateRecovery {
+    /// 404: no repo by that name — a bare target may be a workspace id to attach.
+    RepoMissing,
+    /// 400 on base resolution — an unborn default branch is seedable; other branches are not.
+    BaseUnresolved,
+}
+
+fn create_recovery(e: &tensorlake::error::SdkError) -> Option<CreateRecovery> {
+    let tensorlake::error::SdkError::ServerError { status, message } = e else {
+        return None;
+    };
+    match status.as_u16() {
+        404 if message.contains("not found") => Some(CreateRecovery::RepoMissing),
+        400 if message.contains("does not resolve to a commit")
+            || message.contains("has no commits") =>
+        {
+            Some(CreateRecovery::BaseUnresolved)
+        }
+        _ => None,
+    }
+}
+
 /// A workspace needs a base commit, but a file system fresh out of `tl git create` has an
 /// unborn default branch. Seed it with an empty initial commit so the first mount just works.
 async fn ensure_seeded(session: &FsSession, default_branch: &str, repo: &str) -> Result<()> {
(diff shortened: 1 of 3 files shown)


diptanu:
Adversarial code review of this PR + its gsvc-mount companion surfaced several real hazards; fixes pushed in 2f832fd:

- **Credential hygiene** (the daemon previously adopted any cached token whose expiry was >120s out):
  - adopt cached tokens only with runway past the rotation margin (+5min), else mint fresh — kills the near-expiry + failed-re-mint ~18-minute dead-mount window
  - freshest entry across scopes wins — a stale repo-scoped token from an earlier credential-helper use no longer shadows the fresh `"*"` token the mounting CLI just wrote
  - daemon mints write through to the on-disk cache (`mint_and_cache`), so a cache-cold daemon restart no longer leaves the next CLI command paying the mint again
  - failed rotations retry every ~60s instead of the 30-minute parse-fallback sleep
  - a cached token rejected at mount init (revoked before expiry — auth-epoch rotation is a real server mechanism) purges the cache, re-mints once, and retries; heartbeat auth failures nudge the rotation task instead of waiting out the schedule
  - the env override reuses `git_credential_from_env` instead of a fourth inline copy
- **`start_oid` semantics hardened in the companion** (artifact_storage#95): the daemon still gets a fast start, but ref existence + commit servability are validated concurrently at mount — deleted-workspace phantom mounts and "Mounted"-then-EAGAIN windows are gone.
- **Quiet stderr**: the `mount timing` phase lines are debug level now (`--log-level debug` shows them); the flag's help text says it governs the CLI's own diagnostics too.
- The 404 classifier requires the server's `repo … not found` wording (pinned by the contract e2e in #95), so generic proxy 404 bodies surface raw.
- Cleanups: `ensure_seeded` drops its redundant `ref_status` pre-check, one `create_workspace` call site for attempt + post-seed retry, `follow_ref`/`start_oid` derived in one structural branch.

Known remaining (pre-existing, not introduced here): the git-credential cache is keyed by api_url|project|scope with no identity component, so an identity switch on the same machine can reuse the previous user's unexpired token — was true for all `tl fs`/`tl git` commands before this PR; worth a separate fix keying the cache by auth identity.

PR #834, opened 2026-07-11
tl fs snapshot: keep the overlay after sealing; --clear is the destructive opt-in

diptanu:
## Problem

The manual `tl fs snapshot` destroyed local data two confirmed ways (found during the large-folder snapshot investigation, confirmed by two independent review rounds):

1. **Deterministic**: enumeration honors ignore rules (`target/` is skipped and never pushed), but the post-push `clear_upper` removed the **whole** upper tree — every manual snapshot deleted ignored files written through the mount. Unreadable entries silently skipped by the walker met the same fate.
2. **Race**: a write landing between enumeration and the clear was deleted without entering any snapshot. `clear_upper` deliberately preserves open descriptors, so a racing writer keeps writing an unlinked inode while the dirty index records a path with no backing bytes.

It was also the dominant cost of large snapshots: of a 23.62s 496-file snapshot, 20.64s was the post-clear convergence (one cold HTTP stat per sealed path, sequential).

## Change

- Sealing now **keeps the overlay**, matching the daemon auto-commit path ("the overlay is kept — only `tl fs snapshot` seals-and-clears"): the upper keeps serving the sealed paths, whose bytes the lower now holds identically, so nothing kernel-visible changes and the clear+revalidate phase doesn't run.
- `tl fs snapshot --clear` keeps the old behavior as an explicit, documented-destructive opt-in — it is what empties the dirty set so `tl fs sync` can proceed. Its cost is printed as its own timing bucket (`refresh 1.2s (poll 0.2s, clear+revalidate 1.0s)`).
- `sync`'s dirty-mount refusal now points at `snapshot --clear` with an honest warning; `promote`'s spurious "not in any snapshot" note is reworded.

Reviewed alternatives (rejected as unsafe): clearing only enumerated paths still deletes a newer incarnation of an enumerated path; an overlay-epoch check can't work because ordinary writes advance `write_gen`, not `epoch`.

## Known temporary regressions (until per-path seal tracking)

- Repeat snapshots re-enumerate and re-hash retained upper content (content dedup prevents re-upload; the commit applies the same upserts).
- `status`/`promote` over-report dirt after a snapshot.

## Tests

- `snapshot_seal_keeps_overlay_by_default` — regression repro for both loss modes against a fake daemon control socket: ignored path survives, a raced mid-push write survives **and enumerates into the next snapshot**.
- `snapshot_seal_clear_opt_in_sends_clear_upper` — the opt-in still routes the clear through the daemon.
- CLI parse tests for `--clear`. Full `cargo test -p tensorlake-cli --features mount`: 223 passed.

### DIFF
--- a/crates/cli/src/commands/fs.rs
+++ b/crates/cli/src/commands/fs.rs
@@ -42,6 +42,9 @@ pub mod fusefs;
 pub mod local;
 #[cfg(unix)]
 pub mod overlay;
+// Plain-directory workspace snapshots: `tl fs init` binds a directory to a workspace with no
+// mount at all; snapshot/status dispatch here when the path is a binding rather than a mount.
+pub mod plaindir;
 // How the kernel reaches the overlay on macOS: the FSKit extension proxies this wire protocol
 // over localhost TCP (Linux talks to the overlay in-process through the FUSE glue). macOS-only —
 // on Linux the whole module would be dead code.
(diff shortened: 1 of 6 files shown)


diptanu:
Review round applied in c73f8f2: manual snapshot now seals through the daemon's sealer (one watermark/tombstone path — fixes the auto-commit double-seal, the repeat-snapshot full re-push/no-op commits, and the raced-unlink resurrection), and restore/unmount refuse to drop a non-empty overlay without the new `--discard-local` flag. Stale --clear-era guidance texts updated. 224 tests green.

diptanu:
Second review round applied in 9af93e1: `--clear` now executes inside the seal op under the sealer lock and the CLI revalidates the daemon's full `cleared` list (raw upper+wh walk, not the seal delta — fixes the macOS stale-attr gap and the empty-list clean case); a clean seal with --clear prints what it dropped instead of 'workspace is clean'; `overlay_has_local_state` fails closed on IO errors (unreadable subtree now blocks the destructive path instead of hiding state); the seal protocol streams progress events with a 60s idle timeout (hung daemon is now detectable, long seals show progress) and uses shared serde request/reply structs with a structured `unknown_op` code (prose match kept only as legacy fallback). 227 tests green.

PR #835, opened 2026-07-11
tl fs init: plain-directory workspace snapshots (no mount in the write path)

diptanu:
Stacked on #834. Implements Phase 1 v1 of the reviewed capture/serve decoupling plan (empty-base, single-writer workspaces).

## Why

Snapshotting required a FUSE mount even when every byte is created locally against an empty base — the mount taxes each write and its post-snapshot coherence phase dominated wall time (20.64s of a 23.62s 496-file snapshot). Capture needs only `(path, bytes) → commit`.

## What

- **`tl fs init [path] [--file-system <repo>]`** — binds a plain directory to a freshly created workspace in a **typed binding registry** (separate from the mount registry; overlap with mounts refused both ways). v1 requires a verified-empty base: the head's root tree is checked empty via one tree-page call, and the workspace is deleted again on failure.
- **Two-layer stat index** per binding: server baseline `{oid, mode}` at `indexed_head` + optional `clean_fingerprint` (mode/size/mtime/ctime/dev/ino; entries stamped inside the scan window stay conservatively dirty). Unchanged files are never re-read. An oid binds to a fingerprint only when proven stable across the content read (stat → hash → re-stat); a mid-push mutation keeps the *committed* oid as baseline with `clean_fingerprint = None` — never a stale-clean entry, never a dropped one.
- **Crash safety**: a write-ahead journal (caller-owned idempotency key, CAS base `expected_head`, candidate per-path oids + fingerprints) lands durably before `push_files`; the commit submits with `expect_oid = indexed_head` (preflight head check first, so a stale index fails before hashing gigabytes). Recovery: head unchanged → drop journal, retry fresh; crash between index install and journal removal → drop journal; head advanced → **fail closed** today (adoption verification is implemented + tested behind `AdoptionEvidence` but the SDK lacks a commit-parent API to prove it — seam documented).
- **cloud-sdk**: `PushOptions.idempotency_key` (caller-persistable) and `PushOptions.on_prepared` (per-path prepared oids surfaced after hashing, before any mutation); `BlobOidHasher` exported. Client-only; no wire changes.
- **Strict scanner**: reuses the mount's ignore semantics (built-ins, `.tlignore`, nested `.gitignore`); ignored ≠ missing (an ignore-rule change is never a server delete); unreadable dirs/entries/ignore-files abort rather than enumerate as mass deletion; non-UTF-8 names and fifos/sockets/devices error by name; symlink targets keep raw bytes.
- `status`/`unbind` for bindings; `sync`/`restore`/`promote`/`snapshot --clear` fail with explicit v1 errors. A no-op snapshot verifies the head and does not create a commit (it also silently re-earns cleanliness proofs for rehash-verified files).

## Known v1 limits (per plan)

- Empty-base new workspaces only; no rebind/materialize for existing content (Phase 1.5).
- Crash-after-publish recovery is a hard error until a commit-parent API lands.
- No lease daemon: every binding command heartbeats; an idle binding's workspace lease can expire.

## Tests

14 new in-module tempdir tests (index atomicity/fail-closed, fingerprint + racy-window rules, ignored-vs-missing-vs-unreadable, non-UTF-8/special-file rejection, symlink raw bytes, journal round-trip, recovery decision table, adoption) + parse tests; one `#[ignore]`d documented manual e2e (server on 127.0.0.1:8080, same precedent as overlay tests — not yet run live). `cargo test -p tensorlake-cli --features mount`: 237 passed. Default no-feature build passes.

### DIFF
--- a/crates/cli/src/commands/fs.rs
+++ b/crates/cli/src/commands/fs.rs
@@ -42,6 +42,9 @@ pub mod fusefs;
 pub mod local;
 #[cfg(unix)]
 pub mod overlay;
+// Plain-directory workspace snapshots: `tl fs init` binds a directory to a workspace with no
+// mount at all; snapshot/status dispatch here when the path is a binding rather than a mount.
+pub mod plaindir;
 // How the kernel reaches the overlay on macOS: the FSKit extension proxies this wire protocol
 // over localhost TCP (Linux talks to the overlay in-process through the FUSE glue). macOS-only —
 // on Linux the whole module would be dead code.
@@ -218,11 +221,20 @@ pub async fn ls(ctx: &CliContext, file_system: Option<&str>, output_json: bool) 
     };
     let rows = all_workspaces(&session, &fs_names).await?;
     let mounts = live_mounts();
+    let bound = plaindir::bound_workspaces();
     let mounted_at = |id: &str| {
         mounts
             .iter()
             .find(|(_, s)| s.workspace_id == id)
             .map(|(m, _)| m.clone())
+            // Plain-directory bindings are attachments too; without this they would be
+            // invisible everywhere except path-addressed commands.
+            .or_else(|| {
+                bound
+                    .iter()
+                    .find(|(ws, _)| ws == id)
+                    .map(|(_, root)| format!("{root} (bound)"))
+            })
     };
     if output_json {
         let out: Vec<serde_json::Value> = rows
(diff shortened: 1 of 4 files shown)


diptanu:
Review round applied in aed89b2 (rebased on #834's update): corrupt bindings.json no longer bricks mount commands (lenient policy at mount dispatch, strict for binding ops, one warning); registry mutations are flocked with unique temp names; crash recovery journals the server job id and polls it to terminal — the Fresh-recovery wedge is gone and AdoptionEvidence::Verified is now wired in production (crash-after-publish recovers); dir-replaced-by-symlink occlusion emits subtree deletes; `tl fs rm` and mount-attach single-writer defaults are binding-aware; pre-hash is now 8-wide; unbind walks ancestors; `ls --json` emits a machine-readable kind field. 243 tests green.

diptanu:
Second review round applied in e8369d0 (rebased): recovery now distinguishes job-record-gone (404/expired — the server expires job records after 6h) and falls back to the head-comparison table, so an aged journal can never wedge a binding (head==expected → Fresh; advanced → fail-closed naming the journal path as the escape hatch); transient poll errors retry, cadence ramps 500ms→5s to a 10-min deadline, and 'still running' keeps the journal resumable; Adopt now carries its head (no empty-string sentinel). unbind removes the state dir before the registry entry, dangling entries self-heal, and orphaned state dirs are reapable — `tl fs rm`'s advice is always followable; init re-checks overlap inside the locked registry mutation and cleans up on abort. Registry-corruption handling is coherent: mount-arm errors carry the corrupt-registry note (visible to --json), corrupt binding.json fails `tl fs rm` closed, warn-once actually warns once. Job id persists via a tiny sidecar written off the progress callback (no more mid-push full-journal fsyncs). 254 tests green.

PR #838, opened 2026-07-11
tl fs ls: one workspace-fleet request instead of one listing per file system

diptanu:
## Why

`tl fs ls` fanned out one HTTP request per file system in the project (after a `list_repos` call), each hitting artifact-storage's per-repo workspace listing. With many file systems that's `1 + ceil(N/8)` WAN round-trip waves per invocation; `tl fs rm` and mount's workspace-attach recovery paid the same fan-out to resolve an id.

## What

One paginated call to the project-scope fleet endpoint (`GET /project/{p}/workspaces?principal=self`, server half in [artifact_storage#101](https://github.com/tensorlakeai/artifact_storage/pull/101)):

- **cloud-sdk**: `workspace_fleet` method + `WorkspaceFleetPage`/`WorkspaceFleetItem` wire types, paginated via `next_after`. Decode is pinned by unit tests here; the server's encode is pinned by a matching key-set contract test in artifact_storage's fleet e2e — change either together.
- **Auth**: the listing session always mints the project-wide credential now — repo-scoped mints carry no `project:read`, so the fleet route would 403. A named file system narrows server-side via `?repo=` instead.
- **Old-server defense**: servers that predate `principal=self` ignore it; the CLI decodes the credential's `sub` (unverified, filter-only) and drops other principals' rows client-side. Opaque dev tokens skip the filter, matching open-mode semantics.
- **`resolve_workspace`**: finds the id project-wide via the fleet (`q=` substring, prefix-filtered client-side), then fetches the authoritative record through the per-repo principal-checked API — resolution never widens what mutations may touch.
- **Output**: JSON keeps its shape (`workspace` rebuilt from the fleet row; workspaces are durable, so the lease fields are constants) and gains `status` / `snapshot_count` / `mounted_on`. The table's Mounted column now shows `<host> (remote)` when the workspace is mounted on another machine.

## Sequencing

artifact_storage#101 should deploy before this releases — until then the client-side principal filter keeps the listing correct, and the fleet endpoint itself has been in prod since issue #56.

## Tests

- SDK: `fleet_page_decodes_server_shape`, `fleet_item_tolerates_omitted_optionals`
- CLI: `token_subject_reads_jwt_sub_and_rejects_opaque_tokens`
- Full suites green: `cargo test -p tensorlake --lib` (137), `cargo test -p tensorlake-cli --bin tl` (199 default / 260 with `--features mount,git-clone` and vendored gsvc crates); clippy + fmt clean on touched crates.

### DIFF
--- a/crates/cli/src/commands/fs.rs
+++ b/crates/cli/src/commands/fs.rs
@@ -28,7 +28,7 @@ use tensorlake::artifact_storage::ingest::{PushFile, PushOptions, PushSource};
 use tensorlake::artifact_storage::models::GitCredential;
 use tensorlake::artifact_storage::workspaces::{
     CreateWorkspaceRequest, PromoteOutcome, PromoteWorkspaceRequest, SyncWorkspaceRequest,
-    TreeEntry, WorkspaceInfo,
+    TreeEntry, WorkspaceFleetItem, WorkspaceFleetQuery, WorkspaceInfo,
 };
 
 use crate::auth::context::CliContext;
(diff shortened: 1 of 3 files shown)


diptanu:
Code-review follow-ups pushed (4f47fa0):

- **Honest lease fields**: fleet items now decode the server's `lease_secs`/`lease_due_ms` (added in artifact_storage#101's follow-up); `ls` derives `pinned` from the real lease row instead of hardcoding constants. One named `fleet_item_to_info` conversion, unit-tested for both durable and legacy-leased rows.
- **Repo-scoped credentials keep `tl fs ls <fs>`**: a 403 from the project-scope fleet falls back to the per-repo, principal-bound listing (liveness columns simply absent).
- **`resolve_workspace`** maps a get_workspace 404 (deleted between fleet hit and fetch) to the friendly "no workspace matches" instead of a raw server error.
- **Pagination lives in the SDK now** (`workspace_fleet_all`, same convention as the other list helpers); the principal filter is a plain match; `token_subject` documents the static-token gap that only server-first deployment closes.

PR #839, opened 2026-07-12
tl fs: one definition of dirty — sealed index, daemon dirty/trim ops, truthful status, non-destructive sync

diptanu:
## Problem (live repro, prod sandbox 2026-07-11)

#834 shipped a documented temporary regression, and it bit immediately: after a snapshot, `tl fs status` reported **every sealed-but-retained overlay file as `M`** — a slatedb workspace showed *"496 change(s)"* forever while `tl fs snapshot` correctly answered *"Nothing to snapshot: workspace is clean."* Three commands held three different notions of dirty:

| Command | Source of truth | After a seal |
|---|---|---|
| `snapshot` | daemon sealer (dirty index + watermark) | correct |
| `status` / `promote` / `diff` | raw walk of `upper/` | **over-reports, forever** |
| `sync` | raw walk of `upper/` | refuses; only exit was destructive `--clear` |

Worse, the `--clear` guidance was a data-loss footgun: in the repro workspace it would have deleted a nested `.git` (~3,950 ignored-only files living nowhere but the overlay).

## Change

**One definition of dirty.** New `dirty` control op = the dry-run twin of seal resolution (same dirty index, same ignore rules, same dir-event expansion, no side effects — verified by a test that the tombstone arm does NOT write the whiteout). `status`/`promote`/`sync`/`diff` all consume it via one helper; the raw walk survives only as a labeled daemon-down fallback. It reads a lock-cheap mirror of sealer state, so `status` answers instantly mid-push. `status` now shows `retained:` and `ignored:` as their own categories (JSON: `dirty_exact`, `retained`, `ignored`).

**Seal records survive restarts.** The sealer persists `sealed.json` — per-path size/mtime/git-mode captured at *resolve* time (a racing write always mismatches the record; residual exposure is the classic racy-git same-size-same-ns window), deletes as inert-whiteout markers, plus the sealed commit. Startup and `reindex` reconcile the pessimistic rebuilt dirty index against it, with a generation ceiling so live kernel writes can't be absolved. `clear_upper`/restore/epoch changes reset the index.

**Sync stops needing `--clear`.** New `trim` control op drops retained overlay state in place (lower serves identical bytes; kernel gets invalidations). Dirty and ignored files are never touched; paths pinned by open write handles are reported, not deleted. `sync` pre-flight: refuse while dirty → fix is a plain `tl fs snapshot`; then trim; refuse by name on held-open files. Ignored files now *survive* sync. Trim safety comes from a mutation fence — a `tokio::sync::RwLock` held shared by every mutating overlay entry point and exclusively by trim — closing the copy-up/unlink race (a writer could otherwise be left feeding an unlinked inode no seal can read).

**Retained upper files are deliberately kept outside sync** — they're the local byte cache; background-trimming them would force the next edit of a sealed file to re-read just-pushed bytes off the server. Post-seal auto-trim covers inert whiteout tombstones only. `snapshot --clear` is demoted to the documented disk-reclaim escape hatch.

## Tests

- `resolve_dirty` == `resolve_seal` publication set; dir-event expansion; vanished-recently-sealed reports the delete **without** writing the whiteout.
- Sealed index round-trip + corrupt-file degradation to pessimistic-empty; survivor matching by exact stat identity; resolve-time stat capture.
- `local_changes` prefers the daemon view (retained/ignored accounting) and falls back labeled without a daemon.
- Server-backed overlay e2e (local gsvc-server): trim drops retained content in place (merged view byte-identical, lower-served), skips dirty paths, pins open write handles; existing overlay e2e passes under the new fence.
- Full suite: `cargo test -p tensorlake-cli --features mount` 262 passed; default features 199 passed.

Old daemons: a CLI talking to a pre-this-PR daemon gets the labeled fallback in `status`/`promote`/`diff`; `sync` asks for a remount (it needs the new ops).

### DIFF
--- a/crates/cli/src/commands/fs.rs
+++ b/crates/cli/src/commands/fs.rs
@@ -5,15 +5,20 @@
 //! artifact-storage repo backing them (managed with `tl git`); a mounted workspace (private
 //! leased ref) is served by a FUSE daemon — reads stream lazily from the server through the
 //! vendored `gsvc-mount` core's immutable caches, writes land in a local overlay.
-//! **The overlay is the dirty set**: the daemon's sealer resolves its dirty index into
-//! incremental snapshot commits on the workspace ref (auto-commit ticks it; `tl fs snapshot`
-//! runs one cycle through the `seal` control op), and the mount's lower layer follows the ref
-//! to the new snapshot. The overlay is **kept** after sealing —
-//! the upper keeps serving the byte-identical sealed content; `snapshot --clear` is the
-//! explicit, destructive opt-in that drops it (required before `sync`). `promote`
-//! CAS-advances a real branch (squash by default); `restore` refills the overlay from any
-//! snapshot. FUSE is the only mount path — Linux builds
-//! carry it unconditionally, macOS requires macFUSE and the `macfuse` build feature.
+//! **The daemon's sealer owns the definition of dirty**: the overlay's dirty index resolves
+//! into incremental snapshot commits on the workspace ref (auto-commit ticks it; `tl fs
+//! snapshot` runs one cycle through the `seal` control op), and the mount's lower layer
+//! follows the ref to the new snapshot. Every dirt-consulting command (`status`, `promote`,
+//! `sync`, `diff`) reads the same view through the `dirty` control op, so none of them can
+//! disagree with what a snapshot would seal. The overlay is **kept** after sealing — the
+//! upper keeps serving the byte-identical sealed content as a local byte cache, accounted
+//! for by the sealed index (`sealed.json`, which also survives daemon restarts) and
+//! reported by `status` as `retained`. `sync` asks the daemon to `trim` retained content
+//! (safe: it is all in workspace history; ignored files survive); `snapshot --clear`
+//! remains the explicit, destructive opt-in that drops the WHOLE overlay, ignored files
+//! included. `promote` CAS-advances a real branch (squash by default); `restore` refills
+//! the overlay from any snapshot. FUSE is the only mount path — Linux builds carry it
+//! unconditionally, macOS requires macFUSE and the `macfuse` build feature.
 
 use std::collections::HashMap;
 use std::path::{Component, Path, PathBuf};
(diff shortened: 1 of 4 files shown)


diptanu:
## Adversarial review round 1 (8 finder angles + verification) — fixes pushed as 830b38b + b90d422

**Fixed:**
1. **copy_up records dirt** (the blocker): an `O_RDWR` open with no write, or a `touch` (setattr with neither size nor mode), materialized an upper copy that no dirty index, seal, or trim ever learned about — invisible to `status` and silently shadowing every later sync of that path. Pre-PR the raw-walk sync refusal caught this accidentally; the new flow made it consequential. copy_up now records the Upsert (free to seal — dedup commits by reference). e2e extended to pin it.
2. **Post-push epoch re-check in `seal_once`**: a `clear_upper`/restore racing the push window could resurrect a `sealed.json` describing the dropped world right after the op reset it; a stat-coincident future file could then absolve against dead content.
3. **Unlink failures report as `held_open`**: sync now refuses by name instead of proceeding over a still-shadowing copy when a trim unlink fails (EACCES etc.).
4. **Post-seal tombstone trim uses `try_write`**: hygiene never parks the seal — and, behind the write-preferring fence, all new mutating ops mount-wide — behind one slow in-flight copy-up.
5. **`trim_all` gates tombstones on the lower serving the sealed commit** (the gate `seal_once`'s own hygiene already had): removing a whiteout while the ref lags transiently resurrects the deleted file mid-sync.
6. **Tombstone trim skips redirect-covered paths** (a whiteout can be load-bearing over remapped lower content inside a renamed tree).
7. **`local_changes` requires the dirty payload field** — a bare `{"ok":true}` ack can no longer parse as an exactly-clean answer.

**Deferred (known, not regressions of this PR):**
- `unmount`/`restore` still gate on raw overlay state, so a retained-only overlay (the post-snapshot default since #834) always demands `--discard-local`, and the gate's suggested remedy (snapshot first) can never satisfy it. Fix wants the sealed-index accounting this PR added — good follow-up.
- Unfenced read-only `open()` can see a transient ENOENT if a sync trim unlinks between its upper-stat and open (lower serves identical bytes; retry-once would close it). Same exposure class as pre-existing `clear_upper`.

**Refuted during verification** (for the record): seal replies can't idle-timeout behind the fence (15s keepalive arm), read-only mounts never reach the trim op (early return), macOS vfsserver doesn't bypass the fence (it routes through OverlayFs), and `prune_dirty` can't eat reindex-rebuilt entries (the mutation clock is monotonic).

Suites after fixes: 262 (mount) + 199 (default) pass; overlay e2e 3/3 against a local server. One unrelated transient during e2e: a by-oid commit hit "referenced oid was relocated or removed concurrently" once right after a server restart — same family as the non-reproducing push flake in artifact_storage BENCHMARK.md notes.

PR #840, opened 2026-07-12
tl fs status: report the sealer's dirty set

diptanu:
## Summary

- add a typed dirty daemon control op backed by the same dirty watermark and path classification used by seal
- make fs status and promote consume that authoritative view, so retained sealed upper files no longer appear as unsealed changes
- use a visibly approximate raw-overlay fallback only when the daemon socket is absent/refused; live old, busy, or malformed daemons fail closed
- serialize restore clear/reindex bookkeeping with seal inspection and preserve raced-delete resurrection guards until pending dirt is resolved

## Scope

This is Phase 1 only. It does not add the persisted sealed index, retained/ignored accounting, overlay trimming, or sync changes. After a daemon restart, the current in-memory design can still conservatively re-report retained upper files until the next seal; persistence is Phase 2.

## Validation

- cargo fmt --all -- --check
- just test-rust
- just build-cli-full --tests
- mount-enabled tl test harness: 266 passed, 4 ignored
- focused coverage for retained upper content, daemon-down fallback, live-old-daemon fail-closed behavior, strict wire parsing, rename deduplication, and side-effect-free raced-delete inspection

Workspace-wide Clippy remains blocked by existing warnings in untouched cloud-sdk/CLI files; the changed mount code introduced no additional Clippy diagnostics.

## Follow-up verification

Before merge, run the live mounted-workspace sequence with the built Linux CLI: write -> status M -> snapshot -> status clean while upper remains -> touch -> status M.

### DIFF
--- a/crates/cli/src/commands/fs.rs
+++ b/crates/cli/src/commands/fs.rs
@@ -5,12 +5,13 @@
 //! artifact-storage repo backing them (managed with `tl git`); a mounted workspace (private
 //! leased ref) is served by a FUSE daemon — reads stream lazily from the server through the
 //! vendored `gsvc-mount` core's immutable caches, writes land in a local overlay.
-//! **The overlay is the dirty set**: the daemon's sealer resolves its dirty index into
-//! incremental snapshot commits on the workspace ref (auto-commit ticks it; `tl fs snapshot`
-//! runs one cycle through the `seal` control op), and the mount's lower layer follows the ref
-//! to the new snapshot. The overlay is **kept** after sealing —
-//! the upper keeps serving the byte-identical sealed content; `snapshot --clear` is the
-//! explicit, destructive opt-in that drops it (required before `sync`). `promote`
+//! The daemon's dirty index is the snapshot source of truth: its sealer resolves that index
+//! into incremental snapshot commits on the workspace ref (auto-commit ticks it; `tl fs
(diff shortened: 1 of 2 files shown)


diptanu:
Closing as superseded by #839, which ships this same phase-1 slice (daemon `dirty` op, status/promote consuming it, labeled daemon-down fallback) plus the parts this PR scoped out: the persisted sealed index (`sealed.json`, so restarts don't re-report), retained/ignored accounting in status, the `trim` op, the non-destructive `sync` flow, and the review-round fixes (notably: copy_up now records dirt, closing an untracked-upper-copy hole neither phase-1 scope would have caught).

The one structural idea here that was better than #839's epoch-check defense has been ported over as tensorlakeai/tensorlake@a66137c: `clear_upper`/`reindex` now route through the sealer (state-lock serialized against in-flight seals), with a `reindex_pending` guard so the dirty view, seals, and trim fail closed while a restore refills the overlay — previously that window read as false-clean.

Two deliberate divergences from this draft, for the record: #839 keeps the labeled approximate raw-walk fallback for a live-but-old daemon in `status` (friendlier than failing closed; `sync` still fails closed), and keeps the flat upserts/deletes/renames wire shape with a payload-field presence check instead of the typed `{path, kind}` entries.

PR #841, opened 2026-07-12
tl fs unmount/restore: retained sealed content no longer demands --discard-local

diptanu:
Follow-up to #839 (flagged in its review round).

## Problem

Since #834, sealing keeps the overlay — and #839 made that retained state accounted-for — but `unmount` and `restore` still gate on the **raw** upper/wh trees. A retained-only overlay is the default state after every snapshot, so every unmount of a workspace that ever wrote a file refused without `--discard-local`, and the gate's own suggested remedy ("`tl fs snapshot` first") could never satisfy it. The flow trained users to reach reflexively for the flag that also deletes genuinely-unsealed ignored files.

## Change

The gate becomes a classification (`overlay_losable_state`), built on #839's machinery (the daemon `dirty` op + the sealed index):

- **Passes**: retained sealed content (dirty-clean and sealed-index-vouched, tombstones included) — loss-free to drop; bare directories (git cannot represent an empty tree, so no snapshot could ever cover one — gating on them would recreate the unsatisfiable gate).
- **Refuses, named in the error**: unsealed changes (pending renames included), ignored local-only files, and any overlay entry *neither* the dirty view *nor* the sealed index vouches for.
- **Fail-closed preserved**: a populated overlay with no daemon to classify it (down, or pre-#839) errors naming the bypass flag; the classification walk is strict — an unevaluable ignore rule or unreadable tree aborts instead of guessing. An empty overlay passes without a daemon, so unmounting a never-written dead mount doesn't demand a remount first.

Error texts now separate the remedies: `tl fs snapshot` for unsealed changes; `--discard-local` only for what a snapshot can never cover.

## Tests

5 new gate tests: retained-only (+ sealed tombstone + bare dir) passes; unsealed refuses; ignored-only refuses; uncovered-entry refuses; daemon-down fails closed unless empty. Full suites: 269 (mount feature) + 199 (default).

### DIFF
--- a/crates/cli/src/commands/fs.rs
+++ b/crates/cli/src/commands/fs.rs
@@ -2275,6 +2275,13 @@ pub async fn unmount(
              {mountpoint}"
         )));
     }
+    // The gate's answer ages while the volume stays writable through teardown; fingerprint
+    // now, re-check at the point of no return (state-dir deletion below).
+    let local_baseline = if discard_local {
+        None
+    } else {
+        Some(overlay_fingerprint(&state_dir)?)
+    };
     let pid = daemon::daemon_pid(&state_dir);
     // The daemon replies to `shutdown` only once the kernel released the volume, so this call
     // covers the slow phase (FSKit teardown on macOS takes seconds); spin so it doesn't look
@@ -2352,6 +2359,20 @@ pub async fn unmount(
                 &state.workspace_id,
             )
             .await?;
+    }
+    // Point of no return. Writes that landed after the gate's answer (through the mount
+    // while it was still attached) would be silently destroyed here — the daemon is gone,
+    // but the fingerprint needs no daemon.
+    if let Some(baseline) = local_baseline
+        && overlay_fingerprint(&state_dir)? != baseline
+    {
+        return Err(CliError::usage(format!(
+            "local changes landed while unmounting; the volume is detached but the local \
+             state is kept at {}. Remount to recover and seal them (`tl fs mount \
+             {mountpoint}`, then `tl fs snapshot {mountpoint}`), or re-run \
+             `tl fs unmount --discard-local {mountpoint}` to drop them.",
+            state_dir.display()
+        )));
     }
     std::fs::remove_dir_all(&state_dir)?;
     registry_remove(&mountpoint)?;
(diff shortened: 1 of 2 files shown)


diptanu:
## Adversarial review round 1 (8 finder angles + verification) — fixes pushed as aad020d

**Fixed:**
1. **Stat-verified vouching** (the blocker): the gate checked `sealed.upserts.contains_key(path)` only — the one race direction that *passed* losable data. A write landing right after the dirty reply, or out-of-band into `state_dir/upper`, now stat-mismatches its seal record and refuses.
2. **Point-of-no-return re-checks**: restore had minutes of remote-tree-walk + download between the gate and `clear_upper` on a live writable mount — and unlike the old always-refusing gate, the new one passes the common retained-only case, making that window routinely reachable. Both commands now capture a raw overlay fingerprint when the gate passes and re-verify it at the destructive step (restore: before `clear_upper`; unmount: before state-dir deletion, daemon already dead — the fingerprint needs none). Any change aborts with state kept.
3. **Offline classification**: sealed.json + stat identity is the same mechanism daemon startup reconciliation trusts, so a dead daemon no longer forces a full remount just to unmount a retained-only workspace (the reboot case). The daemon's dirty view remains an additional refusal source when it answers. Side effect: the gate's ~4 tree walks and double ignore evaluation collapse into one strict pass with ignored-directory pruning.
4. **Coverage outranks ignore rules**: a file sealed before an ignore rule matched it is retained, not "never part of any snapshot" with an unsatisfiable snapshot remedy.
5. **Ignored whiteouts refuse** like ignored files — silently dropping one resurrected the deleted committed file.
6. Smaller: ignore-eval errors carry the `--discard-local` guidance; "entrie(s)" typo; non-unix stub errors honestly; stale `overlay_has_local_state` doc rewritten (it invited re-wiring destructive commands to the raw check); shared rel-path/unreadable helpers (those strings are the join keys against the sealed index — drift = permanently-unsatisfiable gate); sealed.json save timing documented as a cross-process contract.

**Refuted during verification:** bare dirs don't trip the dead-daemon path (`any_entry` recurses past them); sealed.json is never written before push acknowledgment, and a sealed-commit-vs-dirty-commit equality check would break shared-rw refs; rename tuple plumbing traced clean end-to-end.

Suites after fixes: 274 (mount) + 199 (default); 9 gate tests + fingerprint change-detection.

PR #842, opened 2026-07-12
tl fs / tl git split: filesystems are the fs surface, mounts join git

diptanu:
The CLI half of the fs/git decoupling. `tl fs` becomes the filesystem product (the create → copy → save mental model users arrive with); branch-vocabulary mounts move under `tl git`. One engine — workspaces, the mount daemon, plaindir bindings — under both surfaces; they differ in vocabulary and defaults only. **No back-compat layer: the old forms are gone, not deprecated** (second commit).

**Depends on [artifact_storage#102](https://github.com/tensorlakeai/artifact_storage/pull/102)** (repo kinds + genesis snapshots) being merged and deployed: `tl fs create` sends `kind: "filesystem"`, `tl fs ls` filters with `?kind=`, and mounting a fresh filesystem needs the server-minted genesis commit.

## tl fs (filesystem vocabulary — "workspace" never surfaces)

| Command | Behavior |
|---|---|
| `create <name>` | New filesystem, born mountable (genesis save server-side) |
| `ls` | Filesystems + where each is attached locally; `ls <fs>` shows its sessions |
| `mount <name> <path> [--ro]` | Publish-on-save session (shared-rw onto the default branch, materialize), autosave always on for writable mounts, detached local session auto-resumes. A filesystem name and `--ro` are the whole surface: `repo:ref` targets and repositories error with a pointer to `tl git mount` |
| `push <dir> <name> [-m]` | Upload a folder as one save, no mount; re-push reuses the binding's stat index so only changes upload |
| `history [target] [-n]` | Save timeline (operation log; 403 degrades with clear guidance) |
| `rm <name> [-f]` | Deletes the filesystem after a confirm |
| `unmount [--discard] [--delete]` | `--discard` overrides the unsaved-work gate; hidden `--delete` also drops the session |
| `snapshot -m / status / restore [--discard] / setup` | Core verbs; clean snapshot = quiet no-op |

Removed outright (was: deprecated aliases): `promote`, `sync`, `diff`, `rm <session-id>`, `mount <repo>:<ref>`, `--mode`, `--shared-rw`, `--auto-commit-interval-secs`.

Third commit (review round): `init`/`unbind` folded away — `push` is the one way a plain directory enters a filesystem (binds on first use) and `unmount` detaches any attachment (mount or pushed directory); fs `snapshot --clear` removed (hatch stays hidden on `tl git snapshot`); sessions view says publishing/private instead of workspace vocabulary.

## tl git gains the mount family (git vocabulary, git defaults)

`mount <repo>[:<ref>] <path>` (`--ro`, `--publish`, `--workspace <id>`), `unmount` (`--delete`), `snapshot -m`, `sync` (flagless; conflicts materialize as diff3 markers), `promote [path] <branch>` (`--merge` only), `status`. Explicit checkpointing — no autosave, no publish unless asked, and **no `--rw` override**: a workspace attached elsewhere mounts read-only until the other attachment is released. `tl git info` loses its `status` alias to the mount status command (`url` alias stays).

Every surviving flag is either a guard override (`--discard`, `--delete`, `-f`) or the answer to an error (`--merge`, `--workspace`, `--publish`, `--check`); `-m` and `--ro` are the only preference flags.

## Mechanics

- Engine: `fs::mount()` takes an explicit `shared_target` publish branch (replaces the `shared_rw` flag); `plaindir::init` split into a reusable `bind` core so `push` gets a publish-on-save binding; the dead workspace-level `rm()` and `MountWriteMode` are deleted.
- SDK: `create_repo_of_kind` / `list_repos_of_kind`, `kind` on the `Repo` model (serde-defaulted for pre-kind servers, omitted from create when unset).
- Auto-resume: the mounts registry is scanned for a dead-daemon session of the same filesystem; newest wins; read-only mounts never resume a writer session.

## Validation

- Live e2e against a local artifact_storage server built at the #102 commit: new ignored `filesystem_kind_contract_roundtrips` (kind create → listing partition → genesis-backed publish-on-save workspace) plus the full ignored overlay suite — all green, re-run after the back-compat removal.
- 279 unit tests pass; the parse tests now assert the legacy forms **fail** to parse.
- `cargo check` clean for the plain build and `--features mount,git-clone --all-targets`; clippy adds no new warnings.
- gsvc-mount/gsvc-codec placeholders untouched (vendored sources swapped in only for local builds, restored before committing).

## Follow-ups

- `history` reads the admin-gated operation log; a product commits-log endpoint would remove the project-admin requirement.
- Release notes for the removed commands/flags.

### DIFF
--- a/crates/cli/src/commands/fs.rs
+++ b/crates/cli/src/commands/fs.rs
@@ -1,10 +1,15 @@
-//! `tl fs` — versioned filesystem workspaces on artifact storage, mounted over FUSE.
+//! `tl fs` — filesystems on artifact storage: create, mount, save, restore.
 //!
-//! Product model (artifact_storage issue #24): the *workspace* is the unit `tl fs` manages —
-//! `mount` creates or attaches one, `ls` lists them, `rm` deletes them. A *file system* is the
-//! artifact-storage repo backing them (managed with `tl git`); a mounted workspace (private
-//! leased ref) is served by a FUSE daemon — reads stream lazily from the server through the
-//! vendored `gsvc-mount` core's immutable caches, writes land in a local overlay.
+//! Product model (the fs/git split): the *filesystem* is the unit `tl fs` manages — `create`
+//! makes one (a `kind=filesystem` repo, born with a genesis empty save), `ls` lists them,
+//! `rm` deletes them, `mount` attaches one, `push` uploads a folder into one. The vocabulary
+//! is drives and saves: every mount is a publish-on-save session (shared-rw onto the default
+//! branch, autosave on), sessions survive crashes and auto-resume on remount, and the words
+//! branch/commit/workspace never surface. `tl git mount` is the branch-vocabulary sibling on
+//! the same engine (explicit checkpointing, promote/sync verbs, publish only with --publish);
+//! both are served by the same workspace machinery below and the same FUSE daemon — reads
+//! stream lazily from the server through the vendored `gsvc-mount` core's immutable caches,
+//! writes land in a local overlay.
 //! **The daemon's sealer owns the definition of dirty**: the overlay's dirty index resolves
 //! into incremental snapshot commits on the workspace ref (auto-commit ticks it; `tl fs
 //! snapshot` runs one cycle through the `seal` control op), and the mount's lower layer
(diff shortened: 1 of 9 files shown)


diptanu:
Adversarial self-review (same harness as artifact_storage#102: 8 finder angles → verified candidates) surfaced 13 findings; all fixed in `a010405`. Highlights:

**Correctness**
- **`tl fs rm` was no longer fail-closed** — the old workspace-level rm refused when the target was live-mounted or bound; the filesystem-level rewrite lost those guards, so `rm` (or `rm -f`) would delete the repo out from under a running FUSE daemon. Restored: live mounts and tracked directories of the filesystem block deletion with the exact `tl fs unmount` command.
- **Uninvokable remediation text**: the #841 unsaved-work gates still told users to run `--discard-local`, which this PR renamed — the gate's own escape hatch didn't parse. Same for promote's conflict recovery naming the deleted `tl fs sync`/`tl fs promote`. All remediation strings now name commands that exist.
- **Second-writer bypass**: `tl fs mount` forced `WritePolicy::Rw`, so a resumed session attached writable elsewhere became a silent concurrent writer; now `Auto`, preserving the read-only downgrade.
- **`history`**: worked only on mounts (errored on pushed directories — the workflow it advertises), showed failed operations as saves, and applied `-n` before sorting. Now binding-aware, committed-only, newest-first.
- **Silent no-publish push**: `tl fs push` into a pre-split `tl fs init` binding snapshotted to a private workspace and reported success while the filesystem never saw the save. `Binding` records `publish` (serde-defaulted false for legacy) and push refuses with a re-bind hint.
- **Compile break**: the cloud-sdk's own test target still had a 5-arg `create_repo_with_credential` caller (`cargo test -p tensorlake` failed; the CLI's `--all-targets` check doesn't cover it).

**Vocabulary & structure**
- The shared mount engine now takes the surface as a parameter: `tl fs mount` output says filesystem/session/save in every line; `tl git mount` keeps workspace/snapshot/branch. The module doc's "the words never surface" promise was being violated by the most common command.
- Auto-resume runs before any server round trip (resuming costs zero listings); `create`/`ls`/`rm` use FsSession's cached credential instead of minting per call (rm minted twice); one `resolve_filesystem` helper; `dialoguer::Confirm` for rm; one registry scan behind both the live and detached views; kind strings as SDK consts + `Repo::is_filesystem()`; shared credential-purge helper across both surfaces.
- New e2e assertion pins `history`'s save definition to the server's op vocabulary (genesis = committed `push`), so a server rename breaks the contract test instead of history silently going empty.

Deferred with rationale: server-side limit for the operations log (needs an endpoint param), the newest-mtime resume tie-break (documented; all fs sessions publish to one target so the stakes are low), and the ro-path `name:branch` string round-trip through the engine.

279 unit tests green; kind-contract + overlay e2e green against a local artifact_storage server at the #102 review-fix commit. Placeholders untouched.

Issue #103, opened 2025-04-09
[SDK] image builder V2 integration


miguelhrocha:
## Description

As we are rolling out the new iteration of the image builder, we want to provide access to it through our SDK operation `tensorlake deploy`.

## Acceptance criteria

- Users should be able to use the V2 version of the builder by execution: `tensorlake deploy --builder-v2`
- Logs should be streamed

PR #843, opened 2026-07-12
fs/git split: fixes from live e2e against the deployed server

diptanu:
The complete `tl fs` / `tl git` client stack for issue #103, consolidated after the stacked PRs (#844, #845) merged into this branch: live-e2e fixes, one-RTT attach, collision visibility, event-driven autosave, and three rounds of review fixes — each re-validated on a real Linux sandbox (mounts against prod for the attach path; an in-guest artifact_storage#108 server for reconcile flows).

**Requires artifact_storage#108 merged + deployed** (this branch vendors `RefreshDelta.changed` from it; the full-feature CI check goes green when it lands, and the reconcile-base semantics need it server-side).

Highlights on top of the earlier stack:

- **Headless sandbox attach**: `TENSORLAKE_GIT_TOKEN` is authentication for the fs surface (project derived from the JWT `iss` claim) — mounts in ~150ms where they used to hang forever polling for browser approval.
- **Seals declare their base** (captured at the batch's first dirty write): overwriting another session's file is a clean write, concurrent writes collide visibly, resolving a collision converges instead of nesting markers.
- **Retained shadows converge**: the overlay records each seal's blob oids and evicts shadows the branch moved past (own-echo keeps the byte cache), from the delta's raw changed rows with directory-boundary expansion; sealed-delete whiteouts retire when the branch re-adds the path (both re-seeded from the persisted seal index across restarts); divergence-unknown fallbacks evict rather than serve stale bytes. Candidates come only from seal records — dirty and ignored upper content is never touched.
- **Crash-resume preserves work under scoped tokens**: resume and `--workspace` attach through the per-repo endpoint with the known repo AND reopen exactly the selected state dir (dead registrations no longer hold dirs or push resumes onto fresh suffixed ones); Linux lazily detaches dead FUSE attachments. Verified: `kill -9`, re-run the same mount with only the repo-scoped token → same state dir, unsealed write intact.
- **Genuinely event-driven autosave**: write hooks stamp mutation clocks; the idle tick is two atomic loads, no dirty-set resolution, no mtime stats.
- **Unmount gates fixed**: OS-junk allowlist applies before gitignore classification (`.Spotlight-V100` etc. no longer block unmounts); ignored dirs count as one subtree instead of a recursive walk.

Deferred (noted for follow-up): server-side history pagination/filtering — `tl fs history` still filters client-side.

### DIFF
--- a/crates/cli/src/auth/context.rs
+++ b/crates/cli/src/auth/context.rs
@@ -231,3 +231,18 @@ impl CliContext {
             .and_then(|r| r.project_id.clone())
     }
 }
+
+/// The project a pre-provisioned git credential is scoped to, read from the JWT `iss` claim
+/// (artifact-storage mints carry `iss: "project_..."`). `None` for opaque dev tokens and
+/// malformed payloads — callers fall back to configured scope.
+pub(crate) fn project_from_git_token() -> Option<String> {
+    use base64::Engine;
+    let token = std::env::var("TENSORLAKE_GIT_TOKEN").ok()?;
+    let payload = token.split('.').nth(1)?;
+    let bytes = base64::engine::general_purpose::URL_SAFE_NO_PAD
+        .decode(payload)
+        .ok()?;
+    let claims: serde_json::Value = serde_json::from_slice(&bytes).ok()?;
+    let iss = claims.get("iss")?.as_str()?;
+    iss.starts_with("project_").then(|| iss.to_string())
+}
--- a/crates/cli/src/auth/guard.rs
+++ b/crates/cli/src/auth/guard.rs
@@ -44,6 +44,24 @@ pub async fn ensure_auth(ctx: &mut CliContext) -> Result<()> {
 /// Ensure authentication and org/project are available.
 /// Triggers login and/or init flows as needed.
 pub async fn ensure_auth_and_project(ctx: &mut CliContext) -> Result<()> {
+    // Sandbox attach path (issue #103): a pre-provisioned repo-scoped git credential IS the
+    // authentication for the fs surface, and its JWT carries the project claim. Never start
+    // an interactive login/init flow here — the guest is headless, and the browser-approval
+    // poll would hang forever.
+    if std::env::var("TENSORLAKE_GIT_TOKEN").is_ok() {
+        if ctx.effective_project_id().is_none()
+            && let Some(project) = crate::auth::context::project_from_git_token()
+        {
+            ctx.project_id = Some(project);
+        }
+        if ctx.effective_project_id().is_some() {
+            return Ok(());
+        }
+        return Err(CliError::auth(
+            "TENSORLAKE_GIT_TOKEN is set but carries no project claim; \
+             re-mint it with `tl fs token <filesystem>` or configure a project",
+        ));
+    }
     if !ctx.has_authentication() {
         eprintln!("It seems like you're not logged in. Let's log you in...\n");
         let login_result = match run_login_flow(ctx, true).await {
--- a/crates/cli/src/commands/fs.rs
+++ b/crates/cli/src/commands/fs.rs
@@ -647,7 +647,9 @@ pub async fn history(
             }
         }
     };
-    let session = FsSession::open(ctx, Some(&fs_name)).await?;
+    // Project-wide session: the operation log is project-scoped, and repo-scoped mints
+    // deliberately omit that scope (a repo credential would 403 here).
+    let session = FsSession::open(ctx, None).await?;
     let (user, token) = session.creds();
     let ops = session
         .client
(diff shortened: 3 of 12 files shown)


diptanu:
Proactive hardening round before the next review pass — adversarial finder fan-out over this PR's session delta; eight confirmed defects fixed across 5be5bcc (round-4 items) and c95c3a4:

- **State dirs are claimed atomically** (exclusive flock on `daemon.pid` holding the CLI's pid until the daemon overwrites it): the aliveness probe left a multi-second window in which two mounts of one workspace could share a single overlay — and the registry retain would delete the first mount's binding.
- The autosave clocks can no longer be zeroed under a racing committed-dir rename (the sealer's close and the rename's stamp now serialize under the redirects lock): a rename-only batch previously could sleep un-sealed forever.
- Divergence trims never `remove_file` an upper directory (EISDIR retry loop through `divergent_pending`), and whiteout tombstones require delta evidence the branch carries the path — on divergence-unknown fallbacks a sealed delete may simply not have applied yet, and retiring its marker resurfaced files the user deleted.
- mkdir's whiteout probe fails safe (only definite NotFound clears a load-bearing marker); the seal heal re-verifies after writing its marker (branch-side file→dir racing the probe); probe expectations are emitted only for delta-covered paths (wrong `present: true` hammered absent paths and churned the overlay via `probe_negative_dentry`); `resolve_dirty` mirrors the seal's replacement arms so status reports the boundary delete the next seal publishes.

Validated on a Linux sandbox against a local merge of artifact_storage #108 + #109: collision/resolve, the heal race, and crash-resume (unsealed write intact, one state dir per session) all green.

diptanu:
Applied the dirty-index suggestion in the latest commit: healed boundaries are recorded as dirty dir upserts alongside their whiteouts, so they ride every subsequent delta natively — retries probe nothing and re-derive nothing, and the dir-upsert arm publishes the delete plus child enumeration (which the raw forced-delete injection couldn't provide for record-less children). The freshly-healed path is still injected into its own seal by hand (its record generation is past the in-flight delta's watermark), and the ancestor prefix scan stays as the in-memory discovery path for the pruned-record interleaving. Integration test updated to pin the record contract; suites green against the combined #108 server (which also re-confirmed the folded oid-recheck: the pre-fold binary reproduces the relocation failure, the combined one is 6/6 clean).

diptanu:
All three client findings fixed.

**Finding 1 — trim handles upper directories.** The candidate pass now checks the upper's kind: directories drop via `remove_dir`, child-first (candidates sort deepest-first, so a sealed replacement's children fall before the directory does); a non-empty directory (dirty/held-open/ignored content) reports `held_open` honestly instead of EISDIR-looping. The live-boundary whiteout check moved INTO the fenced trim as a caller-selected mode (`TrimBoundaries::KeepLive` on the divergence path), evaluated AFTER the candidate pass — so a directory the same trim unwinds releases its whiteout in one pass, while the post-seal hygiene callers (which only pass tombstones once the lower serves the sealed commit) retire inert markers regardless of upper kind, exactly as before. New integration test `divergence_unwinds_sealed_file_to_dir_replacement` runs your scenario end to end: sealed file→dir replacement, branch swaps the dir back to a file, one divergence pass drops the directory AND its whiteout, no requeue, lower file resurfaces.

**Finding 3 — registry mutations are serialized and atomic.** `registry_add`/`registry_remove` now take a blocking flock on `mounts.lock` around load-modify-save, and the save publishes via the existing `write_atomic` (unique temp + rename + parent fsync — the same helper the binding registry already used, now shared instead of duplicated). Readers stay lock-free: rename atomicity means they see a complete document or the previous one, never the torn read that degraded to an empty registry.

**Finding 5 — a healed boundary seals once.** The sealer now re-snapshots its delta after a non-empty heal: the second snapshot's watermark covers the heal record, so the boundary rides THIS seal natively (dir-upsert arm publishes the whiteout-delete plus children — the manual `delta.deletes` injection is gone entirely) and success-pruning consumes the record. The integration test asserts the exact contract: the heal record's generation is past the pre-heal watermark, and pruning to the re-snapshotted watermark leaves the dirty index clean.

Validation: 279 unit tests green; full overlay ignored suite (now 6 tests) green twice against a local server built from #108's head; `cargo check` clean on the paired full-feature build.

diptanu:
Follow-up commit 0963146: `tl fs status` was paying two sequential server round trips (get_workspace + an explicit heartbeat, each on a fresh TLS connection) for three facts that are immutable for the session's lifetime. The create/attach response now caches created_at_secs and shared_target in the mount state, making the human status path fully local (measured 10ms; works with the server unreachable). The status-side heartbeat is removed outright — the daemon owns the lease, and status stamping it for a dead daemon would falsely read as attached-elsewhere. `--json` keeps fetching the live record by contract; pre-cache state files fall back to the fetch.

diptanu:
Follow-up commit 48c2aab, from the question "can status hang during a background snapshot?": a healthy seal never blocks status (ping reads core state; the dirty view reads the sealer's published mirror, not the state lock held across seal_once) — but a WEDGED daemon hung it two ways: the one-shot control round trip had no reply deadline, and the fallback overlay walk read ignore rules through the mountpoint, i.e. kernel calls served by the stuck process. Inspection ops now carry a 10s deadline (mutating ops like trim still wait — they legitimately queue behind an in-flight seal), and a deadlined ping/dirty makes status skip all mountpoint reads and report "daemon: not responding — possibly wedged" + "local: unknown" instead of a false clean. Verified by SIGSTOPping a live mount's daemon: status returns in 11s with the wedged report and recovers after SIGCONT; healthy-path status unchanged (10ms).

PR #107, opened 2025-04-15
Add Function SDK interfaces for function timeouts, retries, resources

eabatalov:
Add Function SDK interfaces for function timeouts, retries, resources

PR #844, opened 2026-07-13
Attach in one round trip: server-owned filesystem semantics (#103 step 2)

diptanu:
Step 2 of [artifact_storage#103](https://github.com/tensorlakeai/artifact_storage/issues/103), stacked on #843. Consumes [artifact_storage#104](https://github.com/tensorlakeai/artifact_storage/pull/104) — **do not release before #104 is deployed** (older servers silently ignore `surface`, which would drop publish-on-save).

## What changes

- **`tl fs mount` is one server round trip again.** The workspace create carries `surface: "filesystem"` (+ `read_only` for `--ro`) and the server fills the publish target from the repo's stored default branch and rejects kind mismatches with a typed error naming `tl git mount`. The pre-flight listing is gone, which kills two prod findings structurally: the ~30s create→mount staleness window (nothing cached is consulted) and the project-scope requirement (every attach-path call is repo-scoped now).
- **Read-only mounts follow `HEAD` itself** — the daemon polls `ref-status?ref=HEAD`; no branch-name resolution anywhere on the client.
- **`tl fs push`** loses its listing and its unconditional seed the same way; first push works under a repo-scoped token.
- **`tl fs rm`** resolves via the authoritative `GET /repos/{r}/meta` point-read (read-your-writes; typed 404/kind answers).
- **`tl fs token <fs>`** mints the narrow credential and prints the sandbox attach recipe — with FUSE available in every target sandbox, "attach from any provider" is literally: `export TENSORLAKE_GIT_TOKEN=…; tl fs mount <fs> <path>`.
- Engine trusts the **create response** for publish state (follow ref + summary text come from the server-filled `shared_target`).
- SDK: `CreateWorkspaceRequest.{surface, read_only}` (omitted when unset), `RepoMetaInfo` + `repo_meta_with_credential`.

## Validation (local server built from #104)

- create → **immediate** mount, same second: writable (publish-on-save, autosave) and `--ro` (HEAD follow); a save from the writable session converged into the ro view.
- Ignored e2e suites green (kind contract incl. surface-filled target + meta point-read + kind-mismatch message; overlay suite).
- 279 unit tests; plain + full-feature builds clean; placeholders untouched.

One pre-existing quirk noticed while testing (not addressed here): the detached mount daemon re-execs without inheriting `--api-url`/`--api-key` flags (env forms propagate fine) — only visible when pointing a mount at a non-config server, e.g. local testing.

### DIFF
--- a/crates/cli/src/auth/context.rs
+++ b/crates/cli/src/auth/context.rs
@@ -231,3 +231,18 @@ impl CliContext {
             .and_then(|r| r.project_id.clone())
     }
 }
+
+/// The project a pre-provisioned git credential is scoped to, read from the JWT `iss` claim
+/// (artifact-storage mints carry `iss: "project_..."`). `None` for opaque dev tokens and
+/// malformed payloads — callers fall back to configured scope.
+pub(crate) fn project_from_git_token() -> Option<String> {
+    use base64::Engine;
+    let token = std::env::var("TENSORLAKE_GIT_TOKEN").ok()?;
+    let payload = token.split('.').nth(1)?;
+    let bytes = base64::engine::general_purpose::URL_SAFE_NO_PAD
+        .decode(payload)
+        .ok()?;
+    let claims: serde_json::Value = serde_json::from_slice(&bytes).ok()?;
+    let iss = claims.get("iss")?.as_str()?;
+    iss.starts_with("project_").then(|| iss.to_string())
+}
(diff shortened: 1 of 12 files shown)


diptanu:
Dependency update: artifact_storage#104 was consolidated (with #105/#106) into artifact_storage#107 — same content; this PR now requires #107 deployed before release.

PR #845, opened 2026-07-13
tl fs: collision visibility and event-driven autosave

diptanu:
Finishes the CLI side of artifact_storage issue #103 (steps 4b + 5). Stacked on #844 (one-RTT attach); requires artifact_storage#106 vendored into gsvc-mount and deployed for the collision surface (degrades gracefully — older servers just never report conflicts, and `conflicted` is serde-defaulted).

## Collision visibility (step 4b)

- Mount daemon tracks unresolved collisions on the followed branch (refresh-delta-fed, cleared when a later save changes the path) and serves them via a new `conflicts` control op.
- `tl fs status` gains an "unresolved collisions" section: `add_add mm.txt (save 372ad1e1877c)` — both saves were kept, markers are in the file, edit and save to resolve. JSON output carries `unresolved_collisions`.
- `tl fs history` badges conflicted saves: `reconcile (collision)`, with an explainer line.
- SDK: `Operation.conflicted` (serde-defaulted).

## Event-driven autosave (step 5)

- The fixed-interval sealer tick becomes: poll every 2s, seal when dirty AND quiet (all dirty upper-file mtimes ≥ 10s old) or dirty ≥ max-latency bound (the old interval). Settled writes publish in ~10s instead of "interval + drift".
- Dirty sets > 256 files skip the stat pass (max-latency bound only).
- Structured `daemon.log` lines: `autosave: sealing reason={quiet|max-latency} dirty_for_s=.. files=.. renames=..` and `autosave: sealed snapshot <oid> seal_ms=.. push_ms=..`.

## Validation (live, local #104+#106 server, two real FSKit mounts)

- Concurrent conflicting saves of the same new file → both mounts' `tl fs status` show the collision with diff3 markers in the file; `tl fs history` badges the reconcile; a resolving save clears status immediately.
- `autosave: sealing reason=quiet dirty_for_s=10 files=1` → `sealed ... seal_ms=324 push_ms=306` — ~10.3s write-to-published.
- Fleet check: 8 concurrent API sessions saving on one filesystem for 45s → 296 saves (6.35/s aggregate, perfectly fair per-session), 0 errors, repo-listing canary p50 <1ms / max 2ms.

Merge order: artifact_storage#104 → deploy → #843 → #844 → this (with artifact_storage#105/#106 deployed before release).

### DIFF
--- a/crates/cli/src/auth/context.rs
+++ b/crates/cli/src/auth/context.rs
@@ -231,3 +231,18 @@ impl CliContext {
             .and_then(|r| r.project_id.clone())
     }
 }
+
+/// The project a pre-provisioned git credential is scoped to, read from the JWT `iss` claim
+/// (artifact-storage mints carry `iss: "project_..."`). `None` for opaque dev tokens and
+/// malformed payloads — callers fall back to configured scope.
+pub(crate) fn project_from_git_token() -> Option<String> {
+    use base64::Engine;
+    let token = std::env::var("TENSORLAKE_GIT_TOKEN").ok()?;
+    let payload = token.split('.').nth(1)?;
+    let bytes = base64::engine::general_purpose::URL_SAFE_NO_PAD
+        .decode(payload)
+        .ok()?;
+    let claims: serde_json::Value = serde_json::from_slice(&bytes).ok()?;
+    let iss = claims.get("iss")?.as_str()?;
+    iss.starts_with("project_").then(|| iss.to_string())
+}
--- a/crates/cli/src/auth/guard.rs
+++ b/crates/cli/src/auth/guard.rs
@@ -44,6 +44,24 @@ pub async fn ensure_auth(ctx: &mut CliContext) -> Result<()> {
 /// Ensure authentication and org/project are available.
 /// Triggers login and/or init flows as needed.
 pub async fn ensure_auth_and_project(ctx: &mut CliContext) -> Result<()> {
+    // Sandbox attach path (issue #103): a pre-provisioned repo-scoped git credential IS the
+    // authentication for the fs surface, and its JWT carries the project claim. Never start
+    // an interactive login/init flow here — the guest is headless, and the browser-approval
+    // poll would hang forever.
+    if std::env::var("TENSORLAKE_GIT_TOKEN").is_ok() {
+        if ctx.effective_project_id().is_none()
+            && let Some(project) = crate::auth::context::project_from_git_token()
+        {
+            ctx.project_id = Some(project);
+        }
+        if ctx.effective_project_id().is_some() {
+            return Ok(());
+        }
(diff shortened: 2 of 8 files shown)


diptanu:
Server-side dependency consolidated: artifact_storage#104/#105/#106 are now a single PR, artifact_storage#107. This PR needs #107 vendored/deployed (same content, one PR).

diptanu:
Review addressed in 0a164d0 — four live defects fixed and re-validated on a Linux sandbox, plus the two strongest performance items. Two findings were already fixed before this review round (in d41a06f): directory-boundary expansion and Option-typed `changed` with fail-safe-on-None — the review snapshot appears to predate that push; both also have live Linux validation (`rm -r` evicting the other mount's retained child).

**Fixed now:**
- **Scoped-token resume (P1)**: resume and `tl git mount --workspace` attach through the per-repo principal-checked endpoint with the known repo — never name resolution. Testing this surfaced a second, worse bug: a crashed mount's surviving registry entry pushed the resume onto a fresh suffixed state dir, **stranding the unsealed overlay**. Registrations now hold their dir only while their daemon lives, and Linux lazily detaches dead FUSE attachments pre-mount. Verified end-to-end in a sandbox: `kill -9` the daemon, re-run the same mount with only `TENSORLAKE_GIT_TOKEN` → 147ms resume, same state dir, the un-autosaved write intact.
- **OS-junk ordering (P1)**: the allowlist applies before gitignore classification, so `.Spotlight-V100`/`.fseventsd`/`.Trashes` no longer classify as uncovered and block unmounts. Ignored dirs count as one subtree (no more walking `node_modules` at unmount).
- **Whiteout re-adds (P1)**: a sealed delete's whiteout retires when the branch re-adds the path, with proper invalidation. Verified live: w1 deletes, w2 re-adds, w1's view converges to the new content.
- **Dir-row trim wart**: eviction candidates come only from seal records (always files), so removed-directory rows are never `remove_file`'d.

**Performance:**
- **Autosave is genuinely event-driven**: write hooks stamp first/last-mutation clocks; the idle tick is a pair of atomic loads — no dirty-set resolution, no mtime stats, per tick or otherwise. The hook clock is also more *accurate* (deletes/renames have no mtime; same-second writes hide in stat granularity). `MTIME_SCAN_CAP` is gone.
- **Divergence intersects the sealed-record map in memory first**: rows untouched by any seal — the bulk of a branch-jump delta — cost zero syscalls.

**Deferred with a note in the commit**: server-side history pagination/filters (needs an artifact_storage ops-listing change; follow-up PR). **P1 build failure**: sequencing — this PR vendors `RefreshDelta.changed` from artifact_storage#108; CI goes green when #108 merges, which is first in the merge order regardless (the reconcile-base semantics need it deployed).

PR #847, opened 2026-07-14
tl fs status: local fast path, wedged-daemon deadlines, honest hints

diptanu:
Recovers four commits stranded by the #843 squash (the branch was silently recreated by pushes that landed minutes after the merge — same trap as before, restacked clean onto main):

1. **`tl fs status` is a local command** — the create/attach response caches `created_at_secs` and `shared_target` in the mount state; the human status path makes zero network calls (measured 10ms, works with the server unreachable). The status-side heartbeat is removed outright: the daemon owns the lease, and a status stamp for a dead daemon would read as attached-elsewhere. `--json` keeps fetching the live record.

2. **Status degrades on a wedged daemon instead of hanging** — inspection ops (`ping`/`dirty`/`conflicts`) carry a 10s reply deadline; a deadlined ping (with a live pid) skips every mountpoint read — the fallback overlay walk's ignore-rule loads go through the stuck daemon's own FUSE path and block unpreemptably — and reports `daemon: not responding … possibly wedged` + `local: unknown`, never a false clean. Verified by SIGSTOPping a live mount's daemon: 11s bounded report, full recovery on SIGCONT. Mutating ops (`trim`/`seal`) keep waiting; they legitimately queue behind an in-flight seal.

3. **The retained hint names a command that exists** — the dumb-simple audit cut `--clear` from `tl fs snapshot` but status (and a seal-refusal error) still recommended it; they now name `tl git snapshot --clear`, which is path-addressed and works on fs-surface mounts.

4. **Test gitignore-only snapshot exclusions** (@diptanu's commit, carried over) — pins that snapshot enumeration has no implicit exclusion list.

Validation: 282 unit tests green on the restacked branch with the paired full-feature build; live checks as described per commit.

### DIFF
--- a/crates/cli/src/commands/fs.rs
+++ b/crates/cli/src/commands/fs.rs
@@ -19,9 +19,9 @@
 //! upper keeps serving the byte-identical sealed content as a local byte cache, accounted
 //! for by the sealed index (`sealed.json`, which also survives daemon restarts) and
 //! reported by `status` as `retained`. `sync` asks the daemon to `trim` retained content
-//! (safe: it is all in workspace history; ignored files survive); `snapshot --clear`
-//! remains the explicit, destructive opt-in that drops the WHOLE overlay, ignored files
-//! included. `promote` CAS-advances a real branch (squash by default); `restore` refills
+//! (safe: it is all in workspace history; ignored files survive); `tl git snapshot
+//! --clear` (the fs surface lost the flag in the dumb-simple audit) remains the explicit,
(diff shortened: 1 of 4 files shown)


diptanu:
Superseded by #848 and its private-engine companion tensorlakeai/artifact_storage#115. The useful status behavior has been carried forward there: human status is local/offline, daemon inspection is bounded, stalled dirty-state queries report unknown rather than clean, and `--json` retains the live platform contract.

PR #848, opened 2026-07-15
Use the native filesystem snapshot engine end to end

### LINKED ISSUE #851: Unify tl fs snapshotting around a durable local mutation journal and generation state engine
## Summary

Replace the current collection of filesystem-local snapshot state machines with one durable, transactional local state engine used by every `tl fs` write workflow.

Today the native mounted-filesystem path has the right hot behavior: mutations feed an in-memory dirty-path index, a background preparer reads/hashes/compresses/uploads only dirty content, and `tl fs snapshot` publishes an already-prepared root without walking the filesystem. However, crash recovery and adjacent workflows still use different mechanisms:

- the mount daemon's in-memory `path -> dirty generation` map
- the on-disk `upper/` and `wh/` trees as restart reconstruction authority
- `sealed.json`
- `native-prepared.json`
- `native-seal-request.json`
- persisted directory-rename redirects
- the separate plain-directory stat index and snapshot-attempt journal
- the separate full-directory `tl fs push` scanner/publisher

This is correct but unnecessarily complex. It also means a daemon restart normally rebuilds dirty state by walking the local overlay, and fixes to snapshot recovery, publication idempotency, or path semantics must be made in several code paths.

Build one embedded local database with a durable WAL and ordered keyspace—an LSM tree is a plausible implementation, but the required abstraction is a transactional, crash-safe state store rather than a specific database. All filesystem mutations, generation preparation, snapshot requests, publication recovery, sealed-state tracking, and local cleanup should flow through it.

For managed mounts, the durable journal becomes the normal source of truth after restart. Overlay walking becomes an explicit repair/migration fallback, not part of the ordinary snapshot or recovery path.

This advances the client side of tensorlakeai/artifact_storage#110 and should replace the transitional state machinery in #848 after the native filesystem vertical is stable. Server behavior remains rooted in tensorlakeai/artifact_storage#111; any wire or idempotency changes should be made atomically across the client and server and reflected in `ARCHITECTURE.md`.

This engine has not reached production. There is no server-side data migration, dual-read/dual-write period, compatibility reader, or conversion job in scope. Server metadata and wire schemas may be changed directly with the companion client, and disposable development server state may be deleted/recreated.

## Goals

1. **No forgotten mutations after an acknowledged filesystem operation.** Once a mutating call succeeds to the application, recovery must conservatively know every affected path or namespace operation.
2. **No normal full walk for mounted incremental snapshots or daemon restart.** Work must be proportional to live dirty/prepared journal entries and affected Merkle ancestors.
3. **One generation state machine.** Manual snapshot, autosave, restore/refill, mounted deltas, tracked directories, and cold import should share preparation, publication, retry, and recovery code.
4. **Snapshot publication remains metadata-only when preparation has caught up.** It must not read local file content, hash, compress, upload, walk directories, or rebuild a Merkle tree on the request path.
5. **Writes continue during preparation and publication.** Freezing generation `N` immediately opens generation `N+1`.
6. **Crash-safe response-loss recovery.** Every network-visible operation has a durably recorded idempotency key and an unambiguous adopt/retry/fail-closed decision.
7. **Bounded local state.** Retired generations and temporary upload artifacts are reclaimed; database size tracks active dirty/prepared work rather than snapshot history.
8. **Fail closed on corruption or an unprovable baseline.** Losing an optimization may cause rereads; it must never cause missed content, false deletion, or an incorrect clean result.
9. **No environment-variable configuration.** Database location and durability policy come from explicit mount/client configuration and existing state-directory layout.

## Non-goals and physical limits

- This does not make arbitrary application file data durable if the application never flushes it. The journal guarantees that a path is not forgotten; file-byte durability still follows the host filesystem's `write`/`fsync` semantics.
- An arbitrary directory passed to one-shot `tl fs push` has no prior namespace knowledge and must be discovered at least once. The initial walk is unavoidable; the goal is for that walk to populate the common generation engine rather than invoke a separate snapshot implementation.
- An unmanaged plain directory can change while Tensorlake is not running. A watcher gap, overflow, or unsupported platform requires a bounded reconciliation/full repair walk. Strict no-walk crash recovery is guaranteed only where all writes pass through the managed mount layer or another authoritative OS change journal.
- The journal does not need to retain every byte-range write forever. Multiple writes to one path in the same open generation may coalesce into one correctness record. Namespace operations whose ordering matters must remain ordered.
- This is not a server-side distributed filesystem journal. It is local preparation state for the native immutable snapshot engine.

## Definition of crash safety

For a managed mount, after a mutating operation is acknowledged:

- the durable state store contains enough information to classify every affected path as potentially changed in a specific generation
- a crash may leave a conservative false-positive dirty entry if the filesystem mutation did not complete
- a crash must not leave a completed filesystem mutation with no durable journal coverage
- recovery resolves false positives by inspecting only journaled paths and explicit namespace-operation participants
- recovery never assumes the tree is clean because a journal file is missing, truncated, or corrupt

This is a write-ahead rule. The correctness record must become durable before the corresponding overlay mutation becomes visible/successful to the caller.

## Why one durable database

The current JSON files are individually atomic, but they do not provide one transaction across:

- advancing the active dirty generation
- recording path/rename intent
- freezing a watermark
- storing a prepared root and its source identities
- persisting a snapshot request/idempotency key
- marking a publication as observed
- installing the new sealed baseline
- retiring the old generation and its temporary files

A transactional state store lets those transitions have one authoritative commit point. It also gives efficient ordered scans over `(generation, path)`, atomic compare-and-update behavior, checksummed pages/WAL records, controlled compaction, and schema-versioned migration.

The store is metadata-only. File and segment bytes remain in the overlay, bounded temporary segment files, and blob storage.

## Storage-engine selection

Start with a short ADR/spike instead of selecting RocksDB by reflex. Compare at least:

- a lightweight embedded LSM implementation
- SQLite in WAL mode
- a pure-Rust transactional B-tree/embedded database
- a small purpose-built append WAL plus immutable checkpoint/compaction files

Selection criteria:

- atomic multi-key transactions
- explicit durability/fsync boundary
- ordered prefix/range scans
- safe single-process locking and clear behavior after unclean shutdown
- corruption detection and fail-closed opening
- bounded compaction stalls
- predictable performance for millions of path keys and small values
- macOS and Linux support in official builds
- minimal native toolchain/dependency burden
- backup/migration/export support for supportability
- no background thread or memory behavior that violates small sandbox limits

The rest of the implementation must depend on a small `LocalSnapshotStore` interface so the chosen engine is replaceable and unit-testable.

## State directory and ownership

Use the existing mount/binding state directory as the ownership boundary, for example:

```text
~/.local/share/tensorlake/mounts/<mount-id>/
  state.json
  upper/
  wh/
  redirects.json          # legacy during migration only
  snapshot-state/         # embedded DB/WAL/checkpoints
  staging/                # bounded local segment/temp artifacts owned by DB rows
```

Requirements:

- one exclusive process lock per state store
- store identity includes project, repo/filesystem, workspace/session, and a random local store UUID
- opening the store against a different remote identity fails closed
- credentials, API keys, bearer tokens, and presigned URLs are never persisted
- persisted server session IDs and operation IDs are allowed, but retries must remint credentials and upload targets
- state-store deletion remains protected by the existing unmount/discard data-loss checks

## Proposed schema

Exact encoding is implementation-defined, but it must be versioned and preserve these logical tables.

### Store metadata

```text
meta/schema_version
meta/store_uuid
meta/project_id
meta/filesystem_id
meta/workspace_id
meta/active_generation
meta/last_retired_generation
meta/current_base_snapshot
meta/overlay_epoch
meta/migration_state
meta/repair_required
```

### Generation record

```text
Generation {
  id
  state
  base_snapshot_id
  created_at
  frozen_at?
  max_sequence
  request_ids[]
  prepared_candidate?
  failure?
}
```

Generation states:

```text
Open
Frozen
Preparing
Prepared
PublishRequested
Publishing
Published
Retiring
Retired
RepairRequired
```

Transitions must be monotonic except an explicitly documented retry transition that preserves the same generation and idempotency key.

### Ordered mutation intents

```text
(generation, sequence) -> MutationIntent

MutationIntent =
  UpsertPath { path, reason }
  DeletePath { path }
  Rename { source, destination, source_kind }
  MetadataChange { path }
  RestorePath { path }
  RestoreSubtree { path }
```

Ordered operations are required for rename chains, delete/recreate, file-to-directory replacements, and other namespace sequences where a final unordered path set is insufficient to reconstruct intent or retain the O(1) committed-directory rename optimization.

### Coalesced dirty-path index

```text
(generation, path) -> DirtyPath {
  last_sequence
  effective_kind
  first_write_offset?
  requires_full_read
  expected_overlay_identity?
}
```

This is the efficient preparation view. Repeated writes to a path in one generation update/coalesce this record transactionally with the ordered intent when needed.

`first_write_offset` and stable-prefix facts are optimization hints. If they cannot be proven after recovery, set `requires_full_read=true`; correctness must never depend on them.

### Prepared generation

```text
PreparedGeneration {
  generation
  base_snapshot_id
  root_id
  exact_change_set
  source_path_identities
  sealed_index_effects
  uploaded_content_refs
  preparation_operation_id
  prepared_at
  validation_version
}
```

Store immutable content references and metadata, not source credentials or file bytes. The record must contain enough information to validate that a recovered candidate still matches the relevant overlay state before it is publishable.

### Snapshot request/publication

```text
SnapshotRequest {
  request_id
  minimum_generation
  selected_generation?
  message
  clear_policy
  publish_operation_id
  server_snapshot_id?
  state
  last_error?
}
```

States:

```text
WaitingForPreparation
Ready
Submitting
Submitted
AdvancingPointer
Completed
Conflict
FailedPermanent
```

The request and its `publish_operation_id` must be durable before the first network call that could publish a snapshot.

### Sealed path baseline

```text
path -> SealedPath {
  snapshot_id
  local_identity
  content_ref?
  kind
}
```

This replaces `sealed.json` and provides the restart-safe distinction between retained sealed cache content and unsaved changes. Missing optimization fields cause rereads; a missing correctness baseline fails closed.

### Temporary artifact ownership

```text
artifact_id -> {
  generation
  path
  kind
  bytes
  lifecycle_state
}
```

Every local segment/temp file must have an owning row before creation or be discoverable as an unowned orphan. Startup and background cleanup can then remove artifacts belonging to retired/aborted generations without scanning unrelated state.

### First content mutation in a generation

For a path not yet dirty in the active generation:

```text
1. Begin local DB transaction.
2. Allocate sequence number.
3. Insert/update ordered intent and dirty-path row.
4. Commit durably to WAL.
5. Perform copy-up/create/write/truncate/chmod in the overlay.
6. Return success to the caller.
```

If the process dies after step 4 but before step 5, recovery sees a false-positive path and resolves it against the overlay. If step 5 completed, the mutation cannot be forgotten.

Subsequent content writes to the same path in the same open generation require no additional durability transaction unless they change a correctness-relevant classification. The lowest write offset may be updated lazily; losing it only disables stable-prefix reuse.

### Writable handles and group commit

Avoid one `fsync` per `write(2)`:

- mark a path durably before returning/activating its first writable handle in a generation
- all later writes through that handle are already covered
- serialize first-dirty records through one journal writer
- group concurrent first-dirty records into a short WAL group commit while each caller waits for the shared durability barrier
- when generation `N` freezes, the first subsequent mutation of that path records it again under `N+1`

Measure the group-commit window rather than hiding it behind an environment variable. If configurable, use explicit config with safe defaults.

### Namespace operations

Create, unlink, mkdir, rmdir, symlink, rename, and file/directory replacement must write their full intent before changing overlay namespace state.

Renames need special treatment:

- record source, destination, source kind, and sequence durably
- apply the local rename/redirect/whiteout operations idempotently
- recovery determines whether the operation was unapplied, partially applied, or complete from the overlay and intent
- committed lower-directory renames remain reference operations; recovery must not materialize or recursively read the remote subtree
- chains such as `a -> b -> c`, exchanges/replacements, and a destination reused after freeze must preserve ordering

If the local overlay cannot prove one valid interpretation after a partial multi-file transition, mark the generation `RepairRequired` rather than guessing.

### Open and freeze

- Exactly one generation is open for new mutations.
- Autosave quiet-time, max latency, or explicit snapshot requests atomically change `Open(N)` to `Frozen(N)` and create `Open(N+1)` in one local transaction.
- The frozen generation captures its exact base snapshot and maximum mutation sequence.
- New writes never modify frozen generation records.
- Multiple snapshot requests may coalesce onto a prepared generation only under an explicit rule. A named/manual request must never silently disappear behind an autosave without its attribution/message policy being defined.

### Prepare

The background preparer consumes only the frozen generation:

1. Resolve its ordered intents into an exact final change set.
2. Inspect only affected overlay paths and explicit namespace participants.
3. Apply `.gitignore` policy consistently.
4. Read changed regular files once unless stable-prefix reuse is proven.
5. Hash, compress, segment, and upload with bounded parallelism.
6. Reuse unchanged file content and directory roots.
7. Rebuild only affected Merkle ancestors.
8. Stage and validate metadata/recipes/root through the native server APIs.
9. Persist `PreparedGeneration` transactionally.
10. Mark the generation `Prepared` only after all immutable server-side content needed for later publication is durable and independently validated.

Preparation may run concurrently with writes into later generations.

### Publish

Publication consumes one exact prepared generation:

1. Persist the snapshot request, selected generation, message, and publish idempotency key.
2. Submit the prepared root and exact change set.
3. Wait for/observe verification state.
4. Advance the workspace/head with expected-base CAS.
5. Persist the returned snapshot ID and actual result.
6. Install sealed-path effects and new local base in one local transaction.
7. Mark the generation published/retiring.
8. Retire its dirty/intents/prepared rows and owned temporary artifacts only after the server result is adoptably durable.

The request path must never fall back to synchronous file reads or a full walk. If preparation is behind, return a structured Pending result immediately and let the durable request resume in background.

### Safe `--clear`

Tie clearing to the published generation rather than “delete the whole overlay later.” After publication:

- remove only upper/whiteout entries whose identities still match the prepared generation's sealed effects
- never remove a path touched in a later generation
- retain ignored/untracked local content unless the user explicitly selected a destructive discard policy
- record cleanup progress durably so a crash resumes it idempotently

This should eliminate the current need for the user to retry `snapshot --clear` merely because preparation was still running.

## Recovery state machine

Startup opens and recovers the database before attaching the kernel mount.

### Open generation

- Resume with the existing durable dirty/intents rows.
- Inspect only those paths when preparation requires actual state.
- Do not walk the entire upper/whiteout tree.

### Frozen or Preparing

- Re-run preparation idempotently using the same generation/base.
- Reuse already uploaded immutable content.
- Expired upload sessions/URLs are reminted; credentials are never loaded from disk.

### Prepared

- Validate the persisted candidate against the generation's durable intent and stored source identities.
- It remains ready for an existing/new snapshot request without reopening unchanged files.

### Snapshot request waiting for preparation

- Wake the preparer immediately.
- Publish the first qualifying prepared generation under the documented coalescing rule.

### Submitting/Submitted/AdvancingPointer with response loss

Use the persisted publish operation ID, expected base, root, and candidate snapshot ID to choose exactly one outcome:

- server did not accept it: retry with the same idempotency key
- immutable snapshot exists but pointer did not advance: retry/adopt the pointer CAS according to server contract
- pointer already equals the requested snapshot: adopt success
- pointer advanced elsewhere: preserve the immutable losing snapshot and report a recoverable conflict
- evidence is ambiguous: fail closed and retain all local/server state for inspection

### Published but not retired

- Confirm the server snapshot/workspace position.
- Atomically install the sealed baseline if missing.
- Resume generation/artifact cleanup.

### Corrupt or missing database

- Never interpret this as clean.
- Refuse writable mount/snapshot by default and preserve `upper/`, `wh/`, and staging data.
- Provide an explicit repair/import command that walks the local overlay and rebuilds a conservative all-dirty generation.
- Emit a support bundle/export path before destructive recovery.

### Managed `tl fs mount`

This is the strongest mode:

- all mutations pass through journal-aware VFS entry points
- normal snapshot and restart require no full overlay walk
- autosave/manual snapshot share one freeze/prepare/publish pipeline
- status reads generation state and resolves only journaled paths

### Restore/refill

Restore must use a journal-aware bulk-apply API:

- create a new generation or explicit overlay epoch
- record subtree/path effects before applying them
- perform refill
- mark the bulk operation complete transactionally
- no separate `reindex` half-state in the normal path

Legacy/out-of-band refills may invoke repair import, but that is not the steady-state design.

### Tracked plain directories

Replace the separate snapshot-attempt journal and stat-index publisher with the common database and generation engine.

There are two supported correctness modes:

1. A long-lived tracker/watch service records changes into the common journal.
2. Without an authoritative watcher history, snapshot performs a strict reconciliation walk to discover changes, writes the discovered generation, and then uses the common preparer/publisher/recovery state machine.

Watcher overflow, downtime, volume replacement, or cursor loss must set `needs_reconciliation`; it must never silently claim the directory is clean. Platform-specific durable change journals may avoid the walk where available, but correctness cannot depend on best-effort inotify-style delivery.

### One-shot `tl fs push`

A new arbitrary directory still needs one initial strict walk. That path should:

1. create an ephemeral/local snapshot store
2. import the discovered namespace as generation 1
3. invoke the common preparer
4. invoke the common publisher
5. retire/remove the ephemeral store only after an adoptable server result

Repeated/tracked use should retain the store and transition into the plain-directory mode rather than maintain another bespoke cold-push state machine.

### Autosave and manual snapshot

They differ only in the request trigger and attribution/message. They must use identical generation, preparation, publication, and recovery machinery.

## Consolidation/removal plan

After migration, remove or reduce these independent authorities:

- in-memory-only `OverlayFs::dirty` as the primary source of truth; an in-memory cache may mirror the durable store
- startup `rebuild_dirty_index()` as the normal recovery path
- `sealed.json`
- `native-prepared.json`
- `native-seal-request.json`
- standalone redirect persistence if rename intents can own it completely
- the plain-directory `index.json`/`journal.json` publication state machine
- duplicate native full-directory versus mounted-delta publish orchestration

Retain one explicit repair walker and one initial cold-import walker, both of which emit ordinary generations into the same engine.

## Concurrency and locking

- one process owns a local state store at a time through an OS file lock
- one journal writer serializes durable first-dirty/group commits
- VFS reads do not need the database writer lock
- freeze is one short local transaction and does not wait for hashing/upload
- preparation workers operate on immutable frozen generations
- later generations may prepare while an earlier one is publishing, subject to bounded CPU/memory/upload concurrency
- publication ordering is explicit; do not advance generation `N+1` past unresolved `N` unless the product semantics define squashing/coalescing and preserve attribution
- compaction never blocks filesystem mutation for an unbounded interval

## Database compaction and lifecycle

- retire mutation/path rows after publication is durably adopted and sealed baseline installation commits
- retain failed/conflicted generation records until explicitly resolved or retention policy expires
- reclaim staging files through owned artifact rows, not filename heuristics
- run compaction/checkpointing in bounded background work during idle periods
- expose database/WAL size and oldest unretired generation
- define a hard fail-closed behavior for ENOSPC during journal commit: the mutating syscall must fail rather than proceed unjournaled
- define safe behavior for ENOSPC during preparation: keep the frozen generation retryable while later writes remain journaled, subject to an explicit local-space backpressure threshold

### Versioning

- every store carries a schema version
- migrations are transactional or use an explicit copy/build/swap sequence
- unknown newer versions fail with a minimum-client-version message
- persisted enums/records must support additive evolution without reinterpreting existing rows

### Existing local development state

No server-side migration is required. Existing development filesystems and server metadata are disposable and may be deleted and recreated.

Local mount state can still contain unsnapshotted user bytes under `upper/` and `wh/`; do not silently discard those files merely because the server is pre-production. On first open after the local-format change, either:

1. import the legacy local state conservatively into one all-dirty generation, or
2. refuse writable operation with an explicit command to discard or repair the local state.

There is no requirement to preserve legacy prepared candidates, server upload sessions, snapshot IDs, or development server history. Prefer a small one-time local safety importer over a general compatibility subsystem. After the new store is durable, remove the legacy local metadata authorities.

### Rollback

Document whether rollback to an older client is supported. If not, fail cleanly before an old binary can mutate a migrated state directory; use an explicit local format marker/minimum version.

## Repair and supportability

Add a command such as:

```text
tl fs doctor <mount-or-binding>
tl fs doctor <mount-or-binding> --repair-journal
```

Read-only doctor output should include:

- database/schema/store identity
- active/frozen/prepared/publishing generations
- dirty path and ordered intent counts
- pending snapshot requests and operation IDs
- staging artifact ownership/bytes
- last adopted server snapshot
- repair/migration state
- integrity-check result

Repair must:

- preserve/export the original database
- walk only local overlay/binding state needed to rebuild a conservative generation
- never issue server mutations unless separately requested
- clearly state that optimization facts may be discarded and content reread

## Correctness invariants

1. Every acknowledged managed-mount mutation is covered by a durable active-generation record.
2. Frozen generation records are immutable.
3. A path modified after freeze appears in a later generation even if it also appeared in the frozen generation.
4. A prepared root is bound to one exact generation, base snapshot, change-set digest, and source identity set.
5. A snapshot request publishes only a qualifying prepared generation.
6. No snapshot request path reads local content or invokes the repair/cold walker.
7. A published snapshot is never forgotten locally because cleanup ran before result adoption.
8. A generation is never retired while it is the sole evidence for a potentially published operation.
9. Corrupt/missing local state never reads as clean.
10. Losing cache/optimization metadata can increase work but cannot omit a change.
11. Server CAS conflict preserves the prepared immutable snapshot and local recovery metadata.
12. Local cleanup never deletes bytes belonging to a later generation.
13. Directory rename recovery never requires reading/materializing an unchanged committed subtree.
14. No persisted row contains credentials or reusable signed URLs.

## Validation strategy

Use the repository's official full/private CLI build and test targets. Add one authoritative journal/state-engine target and include it in the appropriate review target rather than relying on ad hoc `cargo test` invocations.

### Deterministic model/property tests

Generate randomized operation sequences and compare:

- the materialized overlay view
- the journal-resolved change set
- the resulting native snapshot materialization

Operations must include:

- create/write/append/truncate
- chmod and metadata-only edits
- symlink creation/replacement
- unlink/rmdir
- empty directories
- file-to-directory and directory-to-file replacement
- rename chains and destination replacement
- lower-backed O(1) directory renames
- delete/recreate before and after freeze
- concurrent writes racing freeze and preparation

### Crash-injection matrix

Inject process death/faults at least:

- before WAL commit
- after WAL commit, before filesystem mutation
- after filesystem mutation, before syscall reply
- during rename/whiteout/redirect application
- before and after generation freeze
- during file read/hash/compress
- after segment upload, before prepared-row commit
- after prepared-row commit
- after snapshot-request/idempotency commit
- after server submit, before response
- after immutable snapshot creation, before pointer advance
- after pointer advance, before local adoption
- during sealed-baseline installation
- during artifact retirement
- during DB compaction/checkpoint
- during legacy migration

After every restart, assert one of:

- the intended state is recoverable and eventually publishes exactly once
- a conservative retry/reread occurs
- the client fails closed with all evidence retained

Never accept silent missing changes or false deletes.

### Storage faults

- WAL truncation/corruption
- checksum failure
- ENOSPC on journal commit
- ENOSPC during preparation/staging
- permission loss/read-only local volume
- state-directory deletion/identity mismatch
- second daemon attempting to open the store
- compaction crash and orphan staging files

### Lifecycle tests

- daemon restart with 1, 10K, and 1M retained upper paths but only a small dirty set
- restart with open, frozen, prepared, submitting, published/unretired generations
- restore/refill crash recovery
- unmount/remount/resume from another process
- safe `snapshot --clear` with writes racing publication
- autosave and manual snapshot coalescing/attribution
- conflict preservation and retry
- watcher overflow/downtime for plain-directory mode
- one-shot cold import through the common engine
- legacy-state migration and unsupported rollback

### End-to-end verification

For every completed snapshot:

- materialized bytes and supported metadata match the source view frozen for that generation
- unchanged content and directory references are reused
- snapshot history/operation attribution is correct
- a response-lost retry produces no duplicate snapshot/head generation
- a losing CAS snapshot remains recoverable

## Performance acceptance criteria

Benchmark on local macOS and representative Linux sandbox storage. Capture machine profile, source commit, DB engine/config, warm/cold state, and raw artifacts.

Required workloads:

1. 100K and 1M-path mounted trees with no changes.
2. One changed file in a 1M-path retained overlay.
3. 10K changed files produced by a build workload.
4. Repeated writes to one file through one handle.
5. Create/unlink storms of small files.
6. Large committed-directory rename.
7. Daemon crash/restart with large retained state and tiny dirt.
8. Cold `tl fs push` and subsequent tracked-directory incremental snapshot.
9. Preparation and publication while new writes continue.

Acceptance properties:

- normal managed-mount startup/recovery performs no recursive upper/whiteout walk
- startup cost is proportional to database recovery plus live generation rows, not total retained overlay paths
- an idle snapshot performs no filesystem walk, content read, hashing, compression, or upload
- a prepared snapshot request performs no local content work and is dominated by metadata/network publication
- an unprepared snapshot request returns Pending promptly without moving content work onto the control request
- first-dirty journaling adds one durable record per path per generation, not one per `write(2)`
- repeated writes to an already-dirty path avoid durable journal writes until generation rollover
- journal group commit keeps build/create workload regression within a measured and agreed bound; publish both throughput and p95/p99 syscall latency rather than hiding the tradeoff
- restart after crash with one dirty path in a million-path overlay does not stat/walk the million retained paths
- local DB/WAL and staging growth is bounded and returns near steady state after generations retire
- cold import retains the existing parallel scan/hash/compress/upload throughput; routing through the common engine must not reintroduce phase-then-phase behavior

## Observability

Emit structured client logs/telemetry for:

- store schema/engine and recovery mode
- journal commits, group size, bytes, p50/p95/p99 latency
- active/frozen/prepared/unretired generation counts
- dirty paths and ordered namespace intents per generation
- freeze, prepare, upload, validate, publish, adopt, and retire timings
- prepared-hit versus reprepare counts
- recovery decision and response-loss adoption
- repair/reconciliation walks and why they were required
- database, WAL, checkpoint, and staging bytes
- compaction duration/stall time
- ENOSPC/corruption/identity-mismatch failures
- stable-prefix reuse lost due to conservative recovery

Continue reporting snapshot operation IDs and transport/byte metrics through the existing client operation telemetry so support can correlate the local generation with server operations.

### 1. ADR and state-store abstraction

- choose the embedded engine
- define durability/group-commit behavior
- implement `LocalSnapshotStore` and an in-memory model backend
- define versioned keys/values and integrity/open behavior

### 2. Mount mutation write-ahead integration

- route every VFS mutator through durable first-dirty/namespace intent APIs
- implement generation-aware writable handles
- preserve O(1) lower-directory rename semantics
- add fail-closed ENOSPC/error mapping to filesystem calls

### 3. Unified generation preparer

- consume frozen generation rows
- reuse current native delta preparation/upload/composition code
- persist prepared candidates and artifact ownership transactionally
- remove synchronous preparation fallback

### 4. Durable snapshot request and publisher

- persist idempotency keys before network effects
- implement Pending/resume/adopt/conflict state machine
- make autosave/manual snapshot use the same request machinery
- implement generation-safe asynchronous `--clear`

### 5. Restart/recovery and compaction

- recover every generation/publication state without a normal overlay walk
- add bounded artifact cleanup and DB compaction
- add doctor/export/repair tooling

### 6. Restore and plain-directory convergence

- route restore/refill through journal-aware bulk mutations
- replace the plain-directory publication journal/index path with the common engine
- add watcher-gap/reconciliation semantics

### 7. Cold push convergence

- populate an ephemeral generation from the unavoidable initial walk
- reuse the unified preparer/publisher/recovery code
- preserve the pipelined cold-upload path

### 8. Migration, documentation, and rollout

- either conservatively import or explicitly discard existing local mount/binding state; no server-state migration
- remove legacy JSON authorities after validated cutover
- update CLI help and operational repair guidance
- update `ARCHITECTURE.md` in the companion artifact-storage change
- run full private CLI validation and required artifact-storage review/benchmark targets for any server/architecture companion

## Rollout

1. Land the state-store abstraction and shadow-write journal behind an explicit versioned local-format capability.
2. In test/canary builds, compare shadow journal resolution with the existing dirty-index/sealed-file answer and fail loudly on divergence.
3. Make the durable store authoritative for newly created mounts while retaining explicit repair-walk fallback.
4. Migrate existing mount state with one-time conservative import.
5. Converge restore/plain-directory/cold-push paths.
6. Remove legacy JSON/in-memory authorities only after crash and performance campaigns pass.
7. Bump the CLI version and provide a clear minimum-version/local-state-format error for rollback incompatibility.

Do not maintain two writable authorities indefinitely. Shadow mode is validation-only and must have a defined removal milestone.

## Done when

- all managed-mount mutations are durably journaled before acknowledgement
- daemon restart normally recovers without walking `upper/` or `wh/`
- manual snapshot and autosave freeze/publish prepared generations through one state machine
- snapshot publication performs no local content work
- response-lost and crash recovery are unambiguous and idempotent
- restore/refill, tracked directories, and cold push reuse the same generation preparer/publisher
- legacy `sealed.json`, native prepared/request JSON files, and separate plain-directory publication journal are removed as authorities
- explicit repair is the only path that reconstructs a managed mount by walking the overlay
- corruption and ENOSPC fail closed without losing local/server evidence
- state remains bounded through retirement/compaction
- property, crash-injection, lifecycle, migration, and performance gates pass
- companion architecture documentation and any required server wire changes are merged

diptanu:
## Summary

- keep the public repository limited to CLI parsing, host authentication, SDK wire types, and thin adapters
- supply the filesystem and repository-mount implementation through the private `gsvc-fs-client` crate in official local and CI builds
- preserve the completed `tl fs` and `tl git mount` product workflows, including subtree mounts, explicit repository landing operations, recovery UX, and local human status
- keep human `tl fs status` available without platform access while retaining the live-server contract for `--json`
- page repository, operation, workspace, and fleet listings, and reject empty or repeated continuation cursors instead of allowing an SDK call to loop forever
- remove the public implementation sources and internal architecture document
- include current `main`

## Private companion order

The final private client/server source is split across tensorlakeai/artifact_storage#114 and tensorlakeai/artifact_storage#115. #115 is stacked on #114 and the exact combined head has been validated against this PR.

Merge order:

1. artifact_storage#114
2. artifact_storage#115
3. this PR, after rerunning the full-feature CI job against the updated artifact_storage default branch

## Validation

- `make test-fs-client CLIENT_REPO=/Users/diptanuc/Projects/tensorlake-native-fs-client` against the final combined private head
  - 155 private filesystem-client tests ran; 7 local-server tests ignored
  - full Tensorlake Rust workspace tests and doctests passed
  - Clippy passed with warnings denied
  - private/public NativeManifestV2 canonical encoding contract passed
- `cargo test -p tensorlake artifact_storage::` — 29 passed, 1 local-server test ignored
- required exact-head Envoy push/clone/fsck benchmark — 5.215s / 6.855s / 9.060s, 21.291s total; recorded in artifact_storage#115
- `cargo fmt --all -- --check` and `git diff --check` passed

The currently red full-feature GitHub job vendors the artifact_storage default branch. Its reported Linux Clippy and layout failures are fixed in #115 and the exact source passes locally. Rerun that job after #114 and #115 merge.

Closes #851.

Companions: tensorlakeai/artifact_storage#111, tensorlakeai/artifact_storage#114, tensorlakeai/artifact_storage#115


diptanu:
CI triage for head `7107dcb`:

- `Lint and test`, `Sandbox integration tests`, all CodeQL analyzers, and source-distribution build pass.
- `Rust full-feature workspace tests` fails only because the workflow deliberately vendors `artifact_storage/main`; that source does not have `FsClient::new_native` / native `NodeAttr` until companion server PR tensorlakeai/artifact_storage#111 lands. The identical official target passes locally against #111 (324 passed, 7 local-server-dependent ignored, for each CLI binary).
- Windows smoke and all wheel jobs fail inside `PyO3/maturin-action` before compilation because GitHub returns HTTP 503 with its Unicorn page. A failed-job retry produced the same GitHub 503.
- TypeScript integration completed its native build, then failed only when its workflow queried the successful sandbox job through the GitHub API and received the same HTTP 503.

No code failure is hidden by these red jobs. The release order remains server #111 first, then rerun #848 CI against `artifact_storage/main`.

diptanu:
Repository-mount client workflow is complete and pushed at `8b6e807`.

This adds `repo[:ref-or-full-commit]//subtree`, subtree-safe ingest/conflict projection, targeted `sync`, `rebase`, Git-specific `status`, `log`, `smartlog`, stateless `--ro` presence, exact recovery-chain UX, docs, and version 0.5.79.

Validation:
- official private-crate `just test-cli-full`: 329 passed, 7 ignored, 0 failed
- official release `just build-cli-full` passed
- `cargo fmt --all -- --check` and `git diff --check` passed

The remaining red full-feature CI check is dependency ordering, not a client compile defect: CI vendors `gsvc-mount` from the artifact_storage default branch, while these APIs are in companion PR tensorlakeai/artifact_storage#111. The same build passes locally when vendored from #111. Merge #111 before this PR, then rerun the check.

diptanu:
Brought over the useful `tl fs status` behavior from #847 into this PR’s native-filesystem architecture.

- Public CLI routing and docs are in `650c8f1`: human status now runs before auth/platform hydration; `--json` keeps the live-server contract.
- The private-engine implementation is in tensorlakeai/artifact_storage#115: cached local facts, bounded daemon inspection, and unknown state instead of a false-clean result when dirty-state inspection stalls.
- This branch also includes current `main` via `15bce1e`, resolving the prior merge conflict.

Validation: the artifact-storage authoritative client target passed (144 client tests, 7 ignored), the full Tensorlake Rust workspace tests/doctests passed, Clippy passed with warnings denied, and the required Envoy push/clone/fsck benchmark passed and is recorded in the companion PR.

Issue #130, opened 2025-05-07
extraction_example script throw error


cooleel:
run this one https://github.com/tensorlakeai/tensorlake/blob/d2b8244690d29698d57cea093731c4f819645dac/examples/readme_structured_extraction.py

and failed in structure extraction here is the error message in UI

```
Structured extraction error

Task failed with error Error extracting JSON: Unable to extract data. Error code: 400 - {'error': {'message': "Invalid schema for response_format 'BankStatement': ['null', 'null'] is not valid under any of the given schemas.", 'type': 'invalid_request_error', 'param': 'response_format', 'code': None}}

```

drguthals:
Should be fixed with https://github.com/tensorlakeai/tensorlake/pull/313

PR #860, opened 2026-07-20
Add hidden macOS FS cache convergence helper

diptanu:
## Summary

- Add the hidden, argument-free `tl fs converge` helper entrypoint used by artifact_storage #130.
- Dispatch it before authentication and project resolution; the helper accepts only the versioned bounded JSON protocol on stdin and delegates to the private filesystem client.
- Keep it out of the public command surface. It exists solely as a short-lived process boundary so the long-lived FSKit daemon never performs kernel I/O through its own mount.

## Validation

- `cargo fmt --all -- --check`
- artifact_storage #130 `make test-fs-client`, including helper wire/bounds/fail-closed tests and all public CLI/SDK tests
- live macOS FSKit stress described in artifact_storage #130

## Merge order

Depends on tensorlakeai/artifact_storage#130. Merge artifact_storage #130 first, then this PR, with no CLI release between the two merges.

### DIFF
--- a/crates/cli/src/commands/fs.rs
+++ b/crates/cli/src/commands/fs.rs
@@ -82,6 +82,17 @@ pub mod daemon {
             .await
             .map_err(Into::into)
     }
+
+    /// Run one macOS kernel-cache convergence probe as a separate process from the mount daemon.
+    ///
+    /// The private client owns the versioned JSON stdin/stdout protocol and the filesystem work.
+    /// Keeping this adapter argument-free prevents the hidden command from becoming a second
+    /// user-facing mount surface or inheriting authentication/configuration concerns.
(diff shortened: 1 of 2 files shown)


diptanu:
CI sequencing note: the failing **Rust full-feature workspace tests** job vendors private crates from `artifact_storage/main`, which does not contain `run_kernel_convergence_helper` until dependency PR tensorlakeai/artifact_storage#130 merges. This exact full-feature build passed locally with #130 injected via `make test-fs-client` (334 private-client tests plus all public CLI/SDK suites). Merge #130 first, then rerun this job; no Tensorlake code change is required. The independent lint/test, sandbox integration, macOS private-source contract, and CodeQL jobs are green.

PR #865, opened 2026-07-23
Hydrate the mount scope before the human fs-status early return

diptanu:
Fixes the remaining half of the original slow-mount report (tensorlakeai/artifact_storage#147 item 9): human `tl fs status` run outside a project directory showed `last autosave: unknown (missing project ID; run tl init)` (historically the blanket "server unavailable") because the command returned before `hydrate_scope_from_mount` ran.

The hydration is local-only (cached attach-time mount facts, never the network) and is called best-effort, so the offline-diagnostic contract of human status is preserved: with an unreachable server or invalid credentials the local report still prints.

Verified against a live mount: from `/`, status now reports the real last-autosave and permanent-snapshot count; with a bogus `TENSORLAKE_API_KEY` the full local report still renders.

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -2589,16 +2589,18 @@ async fn run_fs_command(ctx: &mut CliContext, subcmd: FsCommands) -> error::Resu
     // Human status is a local diagnostic: the private engine reads cached immutable session
     // facts plus the local journal/control socket, so it must remain available when auth or the
     // platform is unavailable. `--json` deliberately keeps its live server-record contract and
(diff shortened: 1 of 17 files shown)


diptanu:
Superseded by two focused changes:

- #868 replaced the unreleased Git-backed filesystem SDK implementation with the direct blob-store publication APIs and has merged.
- #869 extracts the remaining `tl fs status` mount-scope hydration fix onto current `main`, with an end-to-end offline-status regression test.

Closing this conflicting combined PR in favor of those replacements.

PR #868, opened 2026-07-24
Publish Cloud Volume files directly to blob storage

diptanu:
## Summary

- exchange API keys and PATs for short-lived, repo-scoped Artifact Storage credentials instead of sending platform credentials to the Git data plane
- publish filesystem writes through checksum-bound presigned object-store PUTs with 64 MiB bounded parts, parallel dedup/upload, CAS head publication, and operation-id retry fencing
- add memory-bounded local-path writes, metadata-only copy/move/snapshot/fork APIs, and persistent credential/HTTP connection reuse across TypeScript and Python filesystem clients
- return bytes, immutable content identity, and total size in one file request; execute byte ranges server-side
- fail closed on malformed, duplicated, omitted, or mismatched direct-upload targets and preserve typed filesystem errors

Companion server PR: tensorlakeai/artifact_storage#166
Tracking design: tensorlakeai/artifact_storage#164

## Review follow-ups included

- project and repo credentials share the expiry-checked cache
- Python reuses one ArtifactStorageClient instead of rebuilding its connection pool per operation
- file-backed uploads use 1 MiB transport frames while remaining bounded across eight concurrent uploads
- TypeScript copies caller-owned Uint8Array data before the async native boundary
- public docs now describe current-head and snapshot reads accurately

## Validation

- just test-node-filesystem-full
- cargo test -p tensorlake artifact_storage
- cargo clippy -p tensorlake-rust-cloud-sdk-py --no-deps -- -D warnings
- cargo fmt --all -- --check
- npm run typecheck (TypeScript SDK)
- python3 -m compileall -q src/tensorlake/filesystem

The companion StorageSDK adapter passes install, formatting/lint, typecheck, build, full tests, and publint; it remains uncommitted pending its repository-required explicit commit approval.

### DIFF
--- a/.github/workflows/typescript_sdk.yaml
+++ b/.github/workflows/typescript_sdk.yaml
@@ -6,6 +6,8 @@ on:
   pull_request:
     paths:
       - "typescript/**"
+      - "crates/rust-cloud-sdk-node/**"
+      - "Cargo.lock"
       - ".github/workflows/typescript_sdk.yaml"
 
   push:
@@ -13,6 +15,8 @@ on:
       - "main"
     paths:
       - "typescript/**"
+      - "crates/rust-cloud-sdk-node/**"
+      - "Cargo.lock"
       - ".github/workflows/typescript_sdk.yaml"
 
 permissions:
@@ -105,6 +109,7 @@ jobs:
 
   integration-tests:
     name: Integration tests
+    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository
     runs-on: ubuntu-latest
     timeout-minutes: 30
     environment: prod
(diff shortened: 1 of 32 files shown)


diptanu:
Final review update:

- Fixed the integration suite so the native-filesystem test only runs when API key, organization ID, and project ID are all configured; an API key alone no longer makes unrelated CI fail during client construction.
- `just test-node-filesystem-full` passes (Rust node binding Clippy/tests, TypeScript typecheck, 25 filesystem tests, native release build).
- The unconfigured integration selection now skips cleanly.
- Companion Artifact Storage PR tensorlakeai/artifact_storage#166 now contains the Linux private-client fix. The exact CI-style vendored source passed Linux `cargo clippy --all-targets -- -D warnings` and its Linux test suite in a Tensorlake sandbox.

The new CI run is in progress. The full-feature Rust job vendors Artifact Storage `main`, so that job can remain merge-order dependent until artifact_storage#166 lands; the companion source itself is green on Linux.

diptanu:
CI completion update:

All Tensorlake-owned checks are green: analysis (Actions/TS/Python/Rust), lint/test, SDK integration tests, sandbox integration, Python unit/integration tests, macOS private-source contract, Linux arm64/amd64 wheels, macOS wheel, Windows wheel, and Windows native-binding smoke test.

The only red check is `Rust full-feature workspace tests`, and its failed log contains only the five Linux dead-code diagnostics from the current `artifact_storage/main` checkout. Those exact five diagnostics are fixed in companion PR tensorlakeai/artifact_storage#166; the exact vendored-source Clippy command and test suite pass against that branch in the Tensorlake Linux sandbox. Merge #166 first, then rerun this job.

PR #875, opened 2026-07-29
cloud-sdk: route --cas builds through the Image Service build plane

diptanu:
## Summary

`--cas` builds (CLI, Python, and Node, which all funnel through `build_sandbox_image` / `import_sandbox_image`) now run through the Image Service build plane instead of the builder-local simulation:

1. **The SDK resolves, the service pins** (compute-engine-internal ADR 0023): the existing Dockerfile plan parser is unchanged; its external references (final-stage FROM, earlier-stage FROMs, `COPY --from`, `RUN --mount=from`) are resolved client-side against the Image Service catalog by exact-string name lookup, with `cas-v1:<64hex>` accepted as a direct pin and no ARG substitution. Resolved parents are submitted with the build request as `{reference, image_id}` pairs (capped at 3, the sandbox volume budget beside `/target`); misses stay registry references the guest pulls.
2. **Client flow is create → upload → seal → poll**: plain (uncompressed) context tar honoring `.dockerignore`, uploaded through the sealed capability with a bare HTTP client (presigned-PUT safe), sealed by sha256, then polled to a terminal status; the published image record is returned. The Dockerfile text is injected at `.tensorlake/Dockerfile` when the file lives outside the context directory. The reconciler drives the builder sandbox; the client never talks to it.
3. **Registry credentials forwarded**: the local `docker login` state (the same source the legacy path ships into its builder) is sent write-only as `registry_credentials_json` on build creation; the Image Service stages it outside its catalog and the builder guest fetches it through a short-lived URL for registry pulls.
4. **Simulation deleted**: simulated CAS source injection, the fake local blob store, the unregistered receipt result, and cas-destination spec stripping are gone. A legacy build whose reference resolves to a CAS template fails early with guidance to use `--cas`.

Requires `TENSORLAKE_IMAGE_SERVICE_URL` until the platform gateway routes to the service.

## Testing

`cargo test --workspace` green (5 new unit tests for direct pins, Dockerfile placement, and context tar creation; 59 sandbox_images tests still pass); `cargo clippy --workspace --all-targets` introduces no new warnings; `cargo fmt` clean. Live verification against a running Image Service stack is the next step and will be posted here.

### DIFF
--- a/crates/cli/src/commands/sbx/image/create.rs
+++ b/crates/cli/src/commands/sbx/image/create.rs
@@ -23,8 +23,17 @@ pub async fn run(
     is_public: bool,
     docker_compat: bool,
     cas: bool,
+    build_args: &[String],
     output_json: bool,
 ) -> Result<()> {
+    let build_args = build_args
+        .iter()
+        .map(|raw| {
+            raw.split_once('=')
+                .map(|(key, value)| (key.to_string(), value.to_string()))
+                .ok_or_else(|| CliError::usage(format!("--build-arg {raw:?} must be KEY=VALUE")))
+        })
+        .collect::<Result<Vec<_>>>()?;
     let options = SandboxImageBuildOptions {
         common: CommonBuildOptions {
             api_url: ctx.api_url.clone(),
@@ -49,6 +58,7 @@ pub async fn run(
         dockerfile_path: PathBuf::from(dockerfile_path),
         dockerfile_text: None,
         context_dir: None,
+        build_args,
     };
 
     let mut renderer = ImageBuildEventRenderer::new();
--- a/crates/cli/src/commands/sbx/image/create.rs
+++ b/crates/cli/src/commands/sbx/image/create.rs
@@ -57,10 +57,8 @@ pub async fn run(
         .map_err(|error| CliError::Other(error.into()))?;
 
     if output_json || cas {
-        // Simulation: --cas must expose the unregistered, builder-local CAS
-        // receipt even when --json was not requested.
-        // Final implementation: this prints the normal registered Platform
-        // image response, while --json remains the durable-path opt-in.
+        // --cas prints the published Image Service image record; --json
+        // remains the opt-in for the legacy platform path.
         println!("{}", serde_json::to_string_pretty(&registered)?);
     }
 
--- a/crates/cli/src/commands/sbx/image/import.rs
+++ b/crates/cli/src/commands/sbx/image/import.rs
@@ -56,10 +56,8 @@ pub async fn run(
         .map_err(|error| CliError::Other(error.into()))?;
 
     if output_json || cas {
-        // Simulation: --cas must expose the unregistered, builder-local CAS
-        // receipt even when --json was not requested.
-        // Final implementation: this prints the normal registered Platform
-        // image response, while --json remains the durable-path opt-in.
+        // --cas prints the published Image Service image record; --json
+        // remains the opt-in for the legacy platform path.
         println!("{}", serde_json::to_string_pretty(&registered)?);
     }
 
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -1469,9 +1469,10 @@ enum ImageCommands {
         #[arg(long = "docker_compat")]
         docker_compat: bool,
 
-        /// Build a CAS (content-addressed streaming) image (non-default). CAS
-        /// images cold-boot by faulting content on demand instead of
-        /// localizing a monolithic snapshot.
+        /// Build a CAS (content-addressed streaming) image through the Image
+        /// Service build plane (non-default). CAS images cold-boot by
+        /// faulting content on demand instead of localizing a monolithic
+        /// snapshot. Requires TENSORLAKE_IMAGE_SERVICE_URL.
         #[arg(long, hide = true)]
         cas: bool,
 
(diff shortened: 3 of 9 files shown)


diptanu:
Live gate verification passed end to end against a full local stack (fdbserver, indexify-server, gVisor dataplane, release cas-fs-daemon vs `s3://tensorlake-cas-build-verify/gate5`, image-service with the FDB catalog), driving everything through the real CLI with `TENSORLAKE_IMAGE_SERVICE_URL`:

1. **Import**: `tl sbx image import alpine:3.20 --cas -n demo/alpine-sdk` — local docker.io credentials forwarded write-only, staged at `credentials/gate5/{build}.json`, fetched by the guest through `--registry-credentials-url`, used for the authenticated registry pull, and the staged object was deleted at terminal status. Published `cas-v1:15a12887…` with the runtime contract verified.
2. **Dockerfile build**: `tl sbx image create --cas --build-arg GREETING=from-build-arg` with `FROM demo/alpine-sdk` — the SDK resolved the reference against the catalog and submitted the pin; the reconciler mounted the parent as a read-only volume; **BuildKit ran in-guest with the `oci` executor** (first live run of that path). Published `cas-v1:fefe2878…`.
3. **Boot**: the built image booted as a sandbox; `/hello.txt` reads `from-build-arg` (build-arg proof), `/parent-release.txt` and `/etc/alpine-release` read `3.20.10` (parent-lineage proof).

One environment fix discovered while verifying: images derived from musl bases (alpine) need musl-linked runtime binaries in the builder's embedded-runtime dir; the glibc pair failed at guest init with `exec … indexify-supervisor: No such file or directory`. The gate builder image now embeds musl-release supervisor/daemon (republished as a 16.5 MB delta thanks to pool dedup).

diptanu:
Rebased over main (now includes #878). Resolution notes:

- #878's `local_image_sign_targets` / `sign_prepared_local_image_downloads` are kept — they serve the legacy dataplane-builder path (`COPY --from` durable downloads), which this PR does not touch.
- The old in-`sandbox_images.rs` `options.cas` special-case (including its `simulated_cas_build_result` helper) is gone on this branch by design: `--cas` dispatches to `image_service_builds::run_image_service_build` before that flow runs, and Image Service builds mount pinned parents as volumes rather than downloading signed snapshots, so #878's signing is not needed on that path.

179 lib tests pass, clippy clean.

diptanu:
Updated for the ingress route (`0f7041a`), which changes how `--cas` reaches the Image Service.

**Before:** required `TENSORLAKE_IMAGE_SERVICE_URL` and talked to the Image Service directly. That is only correct for local development — the Image Service has **no authentication of its own** (it reads the project id from the request body and validates only its format), so it is safe to reach only through the ingress, which authenticates, strips every client-supplied `X-Tensorlake-*` header, and injects verified identity.

**Now:** the base URL defaults to `{api_url}/images/v4`, so a normal client needs no configuration and builds go to `https://api.tensorlake.ai/images/v4/builds`. `TENSORLAKE_IMAGE_SERVICE_URL` survives as a development override.

Request paths are now **prefix-relative** — `/builds`, `/names/{ref}`, `/builds/{id}/context/seal` — because the ingress supplies the `/v1` segment when it maps `/images/v4/` onto the service's routes. A path carrying `/v1` itself would arrive as `/v1/v1/...` and 404. The ingress test asserts exactly that doubled shape, so the cross-repository contract is checked from both sides rather than living only in a comment.

Two smaller things worth noting for review:

- URL resolution is a **pure function** over an explicit override, with a thin wrapper reading the env var. My first version read the env inside the function, and its tests raced each other across parallel test threads — one test's `set_var` leaked into another's assertion. Worth avoiding that pattern generally.
- The stale "Requires TENSORLAKE_IMAGE_SERVICE_URL" text is gone from both CLI help strings and the `CommonBuildOptions::cas` doc comment.

183 lib tests pass, clippy clean, fmt clean. Paired with tensorlake-ingress #234, which must merge first — until it does, there is no `/images/v4` route to send these requests to.

diptanu:
That failure is not from this PR. Rebased onto current main anyway (`18d6f1c`) so the run is against a clean base.

**What failed:** the `Rust full-feature workspace tests` job vendors the private `artifact_storage` repo at **its `main`** (`tests.yaml` line 51) and runs that repo's suite alongside the SDK's. Both failures live there:

- `kernel_mount::tests::logical_statfs_reports_the_local_authority_filesystem` — `left: 59753690, right: 59753691`, a statfs free-block count off by exactly one. Something wrote a block between two reads.
- `logical_publish::tests::head_generation_seen_during_dirty_cut_skips_server_conflict_backoff` — *"a pre-cut generation event paid the 60-second server-conflict backoff: Elapsed(())"*, a tokio timeout under CI load.

Neither test name exists anywhere in this repository. This PR touches `image_service_builds.rs`, one doc comment in `sandbox_images.rs`, and two CLI help strings — nothing that can reach a FUSE statfs count or a publish backoff timer.

**Structural note worth someone owning:** every PR here is gated on the current tip of a separate repo's test suite, so an unrelated flake in `artifact_storage/main` reds every tensorlake PR simultaneously, and it cannot be fixed from this side. The workflow already has a mechanism for pinning a specific `artifact_storage` SHA per head branch (used by two stacked mount PRs); making that the default, or pinning to a known-good SHA rather than a moving `main`, would decouple them. Both failures also look genuinely flaky rather than broken — worth a retry to confirm before anyone digs into `artifact_storage`.

Rebase verification: 183 lib tests pass, clippy clean, fmt clean.

PR #887, opened 2026-08-04
Mount daemon memory: fuser fork (configurable buffer) + jemalloc

diptanu:
Companion to artifact_storage: "Cut mount-daemon RSS 5.4x". Two tensorlake-side changes that the `tl` mount daemon needs:

1. **Pin the fuser 0.18 fork** (`tensorlakeai/fuser@configurable-buffer-size`, upstreamed as cberner/fuser#338) via `[patch.crates-io]` + bump the gsvc-fs-client placeholder to fuser 0.18. The fork adds `Config::read_buffer_size` so the daemon can bound its per-worker FUSE receive buffers (stock fuser hardcodes them to 16 MiB × workers).

2. **Use jemalloc as the `tl` global allocator on Linux.** Heap profiling showed most of the daemon's ~1.2 GB real-build peak was glibc holding freed memory across per-thread arenas (live heap only ~216 MB). jemalloc returns freed pages promptly. This is the **single biggest contributor** to the measured 1170 MB → 215 MB reduction. Linux-only (FUSE daemon + CI/dataplane target); macOS keeps the system allocator.

Must land with the artifact_storage memory PR (they're built together via `just build-cli-full`).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4006,6 +4006,7 @@ dependencies = [
  "tempfile",
  "tensorlake",
  "thiserror 2.0.18",
+ "tikv-jemallocator",
  "tokio",
  "tokio-tungstenite",
  "tokio-util",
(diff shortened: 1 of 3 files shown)


diptanu:
Verified that this PR's jemalloc is safe for the CLI release pipeline — **no workflow changes needed**.

Earlier concern was that `tikv-jemalloc-sys` fails to link under `cargo-zigbuild` (`_rjem_*` undefined symbols). Reproduced an exact release-leg environment (Linux x86 host, `cargo zigbuild --release --target x86_64-unknown-linux-musl -p tensorlake-cli --features mount,git-clone`, this branch's code):

- Build links clean (the earlier failures were macOS→Linux **cross** builds; the release legs are host==target — the aarch64 job already runs on `ubuntu-24.04-arm`).
- The static musl binary runs, jemalloc is the live allocator, and the compiled-in `_rjem_malloc_conf` is active with no environment set: `opt.dirty_decay_ms: 0`, `opt.background_thread: true`.

Sequence: merge #889 (fuser pins, unbreaks the release now) → rebase this PR to the jemalloc + decay-config commits → merge. The release binaries then ship the measured −371 MB daemon-peak win (1152 → 781 MB on the 150k-file burst harness) with zero throughput cost.

PR #909, opened 2026-08-15
Fold upload targets into the lease request and honor Retry-After on capacity 503s

diptanu:
Companion to tensorlakeai/artifact_storage#232 (stacked on artifact_storage#230) — the client half of the computesdk storage-benchmark fixes (2026-08-14 run: 95% success rate from un-retried admission 503s, 778ms median uploads under 16-way write contention).

## Combined lease + targets (one round trip)

`publish_filesystem_files` now declares its prepared blob identities in the `fs/upload-leases` request body and consumes the `targets` array the server folds into the lease response — deleting the standalone target-negotiation round trip for publications of up to 256 distinct blobs (the server's fixed per-request ceiling).

**Compatibility, both directions:**
- **New client → old server**: the old handler took no body extractor, so the declaration is ignored; the lease response carries no `targets`, and the SDK issues the standalone batched target requests exactly as before. The pre-existing `direct_native_push_uploads_payload_only_to_the_presigned_target` mock (an old-server shape) now pins this fallback.
- **Old client → new server**: an empty/`{}` lease body keeps the exact two-request flow (verified byte-identical in artifact_storage's `native_filesystem_combined_lease_and_targets_single_round_trip` e2e against MinIO).
- New mock test `direct_native_push_consumes_targets_folded_into_the_lease` asserts a combined-flow publication issues **no** standalone targets request and declares the right blob identities.

Flow note: the head read now overlaps SHA-256 preparation and the lease follows preparation (it needs the blob identities), so large inputs lose nothing — the lease+targets round trip lands where the standalone targets round trip used to.

## Retry-After on capacity rejections

`send_idempotent` retried 5xx/429 three times with 25/50ms sleeps — a ~75ms total budget that burned itself out inside the benchmark's single 2.5s 503 storm, producing its entire 5% failure rate. It now honors `Retry-After` (whole seconds, capped at 2s, 5 attempts) on 429/503; transport errors and hint-less 5xx keep a fast exponential fallback. New mock test drives two `Retry-After: 1` 503s and asserts the waits were honored and the request succeeds on the third attempt.

## Validation

`cargo test -p tensorlake --lib` (199 passed, including both combined-flow shapes and the retry test), `cargo clippy -p tensorlake --all-targets -- -D warnings`, `cargo fmt`.

With artifact_storage#230+#232 deployed, this closes out the benchmark story: server-side contention went from replan-storms + 503s to group-committed ~433 publications/s locally, and the client stops paying one negotiation round trip per file and stops failing on burst rejections.

### DIFF
--- a/crates/cloud-sdk/src/artifact_storage/mod.rs
+++ b/crates/cloud-sdk/src/artifact_storage/mod.rs
@@ -1448,10 +1448,12 @@ impl ArtifactStorageClient {
             )
         });
 
-        // Head and lease are independent control requests. Issuing them together keeps the
-        // fixed-cost portion of a small publication to three sequential data-plane round trips:
-        // (head+lease), targets, then publish after the parallel direct PUTs. SHA-256 preparation
-        // runs concurrently on the blocking pool, so multi-GiB inputs add no serial setup phase.
+        // The head read runs concurrently with SHA-256 preparation on the blocking pool; the
+        // lease request follows preparation so it can DECLARE the prepared blob identities and
+        // fold the first target batch into the lease response — two sequential control round
+        // trips for a small publication (lease+targets, then publish after the parallel direct
+        // PUTs). An older server ignores the declaration and returns no `targets`, selecting
+        // the standalone batched target requests below, which are behavior-identical.
         let head_request = send_idempotent(
             self.git_request(
                 Method::GET,
@@ -1463,28 +1465,10 @@ impl ArtifactStorageClient {
             )?
             .0,
         );
-        let lease_request = send_idempotent(
-            self.git_request(
-                Method::POST,
-                project_id,
-                filesystem,
-                Some("fs/upload-leases"),
-                git_username,
-                git_token,
-            )?
-            .0
-            .json(&serde_json::json!({})),
-        );
-        let control = async {
-            let (head_response, lease_response) = tokio::try_join!(head_request, lease_request)?;
-            let head = decode_json::<NativeHeadResponse>(head_response, String::new())
-                .await?
-                .into_inner();
-            let lease =
-                decode_json::<NativeDirectUploadLeaseResponse>(lease_response, String::new())
-                    .await?
-                    .into_inner();
-            Ok::<_, SdkError>((head, lease))
+        let head = async {
+            decode_json::<NativeHeadResponse>(head_request.await?, String::new())
+                .await
+                .map(Traced::into_inner)
         };
         let prepared = async {
             prepare.await.map_err(|error| {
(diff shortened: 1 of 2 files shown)


diptanu:
Second commit (`b6e4697`) folds in the high-effort review round — six findings, all addressed:

1. **Flow ordering (the important one):** the first cut gated the lease on `try_join!(head, prepared)`, which never reached the promised 2-RTT path and would have cost pre-fold servers a *fourth* round trip. The lease chain now waits only on SHA-256 preparation and runs concurrently with the head read: 2 RTTs on a folding server, exactly the old 3 on a pre-fold one.
2. **Partial/empty folds degrade, never fail:** standalone negotiation covers whatever the folded response did not return, computed from returned blob ids rather than assumed from the declared count.
3. **`Retry-After: 0` can't strip backoff:** hints are clamped to `[exponential fallback, 2s]`.
4. **The direct object-store PUT** (`upload_native_direct_source`) gets the same 5-attempt hint-honoring budget — S3 SlowDown storms have the same shape as the admission 503s.
5. **Test isolation:** the retry test installs the rustls provider itself (it previously passed only when a sibling test's `ClientBuilder` won the race; reproduced red with `--exact`).
6. **Budget scope (deliberate trade, flagging for sign-off):** worst-case in-call blocking grows from ~75ms to ~8s, but only when the server explicitly sends `Retry-After` on every attempt — hint-less failures stay on the fast 50–800ms fallback. A capacity-hinted wait is the behavior the 2026-08-14 benchmark failures needed; callers with hard deadlines see it only when the server is actively shedding.

199/199 lib tests green including isolated runs; strict clippy and fmt clean.

diptanu:
Second review round (full PR diff, high effort) — 9 findings; `68b6b4e` addresses the load-bearing ones:

**Fixed:**
1. **Declaration rejection is no longer fatal** (the round's real catch): the server enforces its per-request target ceiling by rejecting the whole lease request, and the standalone fallback lives *behind* a successful lease — so a server ceiling lowered below the SDK's mirrored 256 would have been a silent new-client-only publication outage. A rejected declaration now retries once with a bare lease and takes the standalone flow; new mock test pins the 400 → bare-lease → standalone-targets sequence.
2. `retry_delay` uses `max().min()` — `Duration::clamp` panics when min > max, one `ATTEMPTS` bump away (50ms << 6 > 2s).
3. Retryable 5xx/429 bodies are drained before sleeping so retries reuse the pooled connection instead of paying fresh handshakes mid-storm.
4. The retry doc now states the actual contract: hints honored up to 2s per wait; storms priced above the whole budget surface as errors (a deliberate bound for deadline-bound callers).

**Deliberately unchanged, with reasoning:**
- `send_signed_read` keeps its fast 3-attempt budget — reads have a graceful fallback path, and parking a read in hinted multi-second waits is worse than falling back.
- The fallback path's lease can't hide under SHA-256 preparation by construction (it must declare the prepared identities); the transitional new-client→old-server timing cost is handled by deploy order (server fold in artifact_storage#232 ships before the SDK release).
- A doomed publication now costs a folding server a lease + presigns before client pre-flight checks — bounded, lease-expiry-cleaned, accepted.
- Consolidating the module's retry loops onto one helper (`with_transient_retries` exists in ingest.rs) is a follow-up.

200/200 lib tests, strict clippy, fmt. Note: the review's finder coverage was thinner than usual on cross-file/conventions angles (several finder agents never reported); the load-bearing findings above were individually verified against both the SDK and server code.

Issue #931, opened 2026-08-27
tl git push: validate repo name and stop silently auto-creating repos


diptanu:
Live incident shape (2026-08-27, prod): a user intending to push to `linux-local1` ended up with their entire worktree ingested into a freshly auto-created repo named `main`, and their real repo cloned as 0 bytes.

`tl git push <REPO> [BRANCH]` takes the repo as the first positional. Git muscle-memory (`tl git push main`) therefore reads the branch name as the repo. Two client gaps turned that mixup into silent data misdirection:

1. **No repo-name validation at the SDK boundary.** One request went to the wire with an empty repo segment — `POST /project/{proj}/repos//ingest/sessions`, which path-collapses to `/repos/ingest/sessions` and got a 403 from the server fence. The ingest client should refuse empty/invalid repo names before building a URL.
2. **Push auto-creates the target repo.** With auto-create, a mistyped repo name cannot fail — it lands the push in a phantom repo (repo `main` was created and committed to within seconds). Options: require an explicit `--create` flag; or when the target repo does not exist, error with "repo X does not exist (did you mean `tl git push <repo> X`?); pass --create to create it" — especially when the repo argument equals a branch-looking name.

Server-side needs nothing: the fence 403'd the malformed path and the empty repo served an honest empty clone manifest.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #934, opened 2026-08-28
tl git push sends real git history by default (resumable; same SHAs as git push)

diptanu:
Phase 2 of artifact_storage#279 (client half; depends on artifact_storage#281 server-side).

`tl git push <repo> [branch]` now pushes the **local branch's real git history** — same commits, trees, blobs, and SHAs a `git push` would publish — through the resumable upload surface:

- **Git-native negotiation**: server-advertised tips that exist locally become `^have` exclusions; `git pack-objects --revs --delta-base-offset` builds the pack, so git owns object selection and deltification exactly as in a push. Up-to-date short-circuits client-side.
- **Resumable transfer**: the pack streams from the `pack-objects` pipe in bounded parts (server-advertised cap), each part retried independently with backoff; 4xx rejections surface immediately. Reading part N+1 overlaps uploading part N; the whole pack is never in memory.
- **Receive-pack semantics end to end**: `--expect-oid` is force-with-lease (CAS old oid), a missing server ref is a create, and the result printed per ref is receive-pack's own report-status — `rejected refs/... : <reason>` included, with a non-zero exit if anything didn't apply.
- **Failure ergonomics**: a retryable finalize failure keeps the staged parts server-side, so re-running the same command re-drives the same bytes without re-upload.

The old worktree-snapshot behavior moves behind `--snapshot` (unchanged mechanics). `-m` is now snapshot-only — a history push sends existing commits unchanged, so a message would have been silently ignored; it errors with the explanation instead.

Why this matters for 10+GB repos: no 7.5 GiB `git pack-objects`-under-smart-HTTP client memory wall on constrained sandboxes for re-pushes (negotiation prunes), no hour-long single POST to lose at 99%, no token-expiry race mid-transfer, and — the incident that motivated all of this — no more disconnected-lineage repos where a `tl`-seeded repo forces plain git to re-enumerate the world (#931's sibling failure mode).

Cut after artifact_storage#281 deploys; on older servers the session create 404s with a clear error and `--snapshot` remains fully functional.

### DIFF
--- a/crates/cli/src/commands/git.rs
+++ b/crates/cli/src/commands/git.rs
@@ -857,17 +857,224 @@ fn parse_remote_message(raw: &str) -> String {
 /// `tl git push` — resumable chunked push of the current Git worktree as one commit.
 /// Retrying after any failure is safe and cheap: the client re-negotiates and uploads only what
 /// the server still lacks.
+/// The history push (artifact_storage #279): send the local branch's REAL commits — same
+/// SHAs a `git push` would publish — through the resumable upload surface. Negotiation is
+/// git-native: the server's advertised tips that exist locally become `^have` exclusions, the
+/// local git builds the pack, and the server's ordinary receive-pack finalizes it.
+#[allow(clippy::too_many_arguments)]
+async fn push_history(
+    client: &ArtifactStorageClient,
+    project_id: &str,
+    repo: &str,
+    branch: &str,
+    root: &std::path::Path,
+    credential: &tensorlake::artifact_storage::models::GitCredential,
+    expect_oid: Option<String>,
+    output_json: bool,
+) -> Result<()> {
+    use tensorlake::artifact_storage::history_push::{GitPushProgress, GitPushRefUpdate};
+
+    let ref_name = format!("refs/heads/{branch}");
+    let tip = run_git(
+        root,
+        &["rev-parse", "--verify", &format!("{ref_name}^{{commit}}")],
+    )
+    .map_err(|_| {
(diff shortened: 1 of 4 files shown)


diptanu:
Self-review round applied (amended into the commit):

1. **Upload errors were masked as pack failures** — on a mid-upload error the reader drops the pipe, `pack-objects` dies of EPIPE, and the old status-first check reported a bogus "git pack-objects failed" (and could block in `wait()` on a child stuck writing). Now: upload error → kill child, wait, surface the real error; success → drain implies child exit, then a nonzero status is a genuine pack failure.
2. **Fast-forward guard** — without `--expect-oid`, a server tip you have locally must be an ancestor of what you push (`merge-base --is-ancestor`), and a server tip you *don't* have locally is a stale-view error advising `git fetch` — matching git's own push rules. `--expect-oid` remains the explicit force-with-lease escape hatch.

Also verified against the server-side review round on artifact_storage#281: retryable finalize failures now genuinely re-drive kept staging on re-run, which is what this client's "safe to re-run" message promises.

diptanu:
Second review pass: fixed two message-literal defects (whitespace runs baked into the stale-view and non-fast-forward errors by an earlier escaping slip — user-visible garbage), and the stale-view error now covers the disconnected-lineage case honestly: when the server branch has unrelated history (e.g. seeded by a snapshot push), fetching cannot produce ancestry, and pushing means replacing it — `--expect-oid` states that intent. Verified no other string literal in the file renders with embedded whitespace runs.

PR #968, opened 2026-09-09
Replay requests the Sandbox Proxy never forwarded for long enough to outlast a rollout

diptanu:
## Why

Both native bindings already retry in two tiers: transient failures (502/503/504, dropped event streams) for idempotent operations, and replays of requests that *provably never reached the server* — connect failures — for everything, including `run_process`. Both budgets were sized for a different problem: **five retries totalling ~4.65 s**, and **two connect replays**.

An Indexify server rollout takes the control plane away for 7–30 s. As of compute-engine-internal #2083 the Sandbox Proxy reports that window as a **503** with a JSON body whose `code` names the pre-forward failure — `SANDBOX_UPSTREAM_ERROR` (lookup didn't answer), `SANDBOX_NOT_READY` (no route yet), `SANDBOX_UNREACHABLE` (dataplane connect failed, or the dataplane said it doesn't host the sandbox and the route couldn't be refreshed) — instead of the 404 that told clients the sandbox was gone. The SDK retried that 503 for 4.65 s and gave up inside the outage. In Node, `run_process` didn't retry it at all, on the (generally correct) grounds that a 503 might mean the request ran.

Those codes settle that question. The proxy failed *before any dataplane executed the request*; it cannot have run. That makes them safe to replay for any operation, and worth replaying for as long as a rollout lasts.

## Change

**`tensorlake::error`** — two methods on `SdkError`, no change to the public enum:
- `upstream_error_code()` parses `code` out of the JSON body that `ServerError` already carries as its message.
- `never_reached_server()` — true for a connect failure, or **any 5xx** carrying one of `SANDBOX_PROXY_UNDELIVERED_CODES` (the three above plus `AUTH_SERVICE_UNAVAILABLE`, which the proxy also emits before forwarding). Any 5xx, not just 502/503, because proxies older than #2083 relay the lookup's own 500/504 verbatim under `SANDBOX_UPSTREAM_ERROR`. A 4xx under those codes is a definitive passthrough (403, 409) and is not trusted; neither is a timeout or an uncoded 5xx.

**`tensorlake::retry`** (new, shared by both bindings — they each had a copy "ported from the Python binding" and had already started to diverge):
- `RetryPolicy::idempotent(n)` / `RetryPolicy::non_idempotent()`.
- `RetryState::decide(&error, elapsed)` is pure, so the policy is tested without a clock; `retry(client, policy, op)` is the async loop over it.
- Never-delivered failures are replayed **regardless of idempotency** until `UNDELIVERED_REPLAY_BUDGET` (30 s) is spent, with waits capped at 5 s (so the replays keep probing) and clipped so the last one can't overshoot the budget.
- Transient failures of idempotent operations keep the **exact schedule the bindings had** — 0.1 s × 2ⁿ × 0.75, capped at 15 s — pinned by a test. Timeouts remain in neither class.

**Bindings** — local `is_retryable` / `is_safe_to_replay` / `calculate_sleep_time` / `MAX_CONNECT_REPLAYS` / loops replaced with calls into the shared module. `run_process` (both) keeps its non-idempotent replay and picks up the longer budget. The Python blocking loop keeps its stderr "Retrying …" line and now **runs with the GIL released**: it is reached from `&self` methods that hold the GIL, and a 30 s budget with the GIL held would freeze every other Python thread for the length of a rollout.

`create_sandbox` (Python sync + async, Node) is also wrapped under the non-idempotent policy — previously unretried. Note what that does and doesn't buy: sandbox creation is a lifecycle request that never goes through a sandbox lookup, so none of the coded 503s apply to it. The wrap replays **connect failures only**. A proxy-side follow-up could emit a coded 503 when the lifecycle upstream can't be connected to; today that surfaces as an uncoded `PROXY_ERROR` 502, which is correctly not replayed.

Behavioural deltas worth a reviewer's eye:
- A plain **connect failure** now also gets the 30 s budget, not two replays. From the client a proxy being rolled is indistinguishable from a control-plane gap, and both resolve on the same timescale.
- Every wrapped call — including idempotent polls like `get_sandbox` — can now block up to ~30 s during a control-plane gap where it previously failed in ~4.65 s. Callers that run their own deadline around such polls will overshoot it by up to that much. This is the intended trade (surviving the rollout beats failing fast inside it), but it is a visible change.
- `start_process_json` in Python still uses the *idempotent* policy (five retries on any 503/timeout-free transient), as do the other mutating routes that were already under `run_with_retry(5, …)` before this PR (create_snapshot, create_pool, write_file, kill/send_signal, delete_*). An uncoded 502 `PROXY_ERROR` *after* forwarding is retried for those. That predates this PR and deserves the non-idempotent policy; left alone here to keep the change to what the incident showed.

## Tests

`retry::tests` (10) and `error::undelivered_tests` (2): both backoff schedules; proxy-coded 5xx replayed for non-idempotent ops and a plain 503 / a coded 4xx not; budget exhaustion and wait clipping; idempotent transient cap; code parsing; the accepted (status, code) matrix. Two paused-clock tokio tests drive the async loop across a simulated 6-attempt gap and to budget exhaustion.

- `cargo test -p tensorlake --lib`: 227 passed
- `just with-function-agent-core cargo clippy --no-deps -p tensorlake -p tensorlake-rust-cloud-sdk-py -p tensorlake-rust-cloud-sdk-node --all-targets -- -D warnings`: clean (staged from a `main`-based CEI worktree; the primary CEI checkout's placeholder manifest has drifted, which the script rejects)
- `cargo fmt` applied

## Pairs with

- compute-engine-internal #2083 — the proxy's coded 503/502 contract this relies on. Against a proxy without it the lookup gap is still a `SANDBOX_NOT_FOUND` 404 and nothing here replays it. Its latest commit also adds `Retry-After: 1` to every proxy 503; this SDK doesn't read the header yet (`ServerError` has no headers), so the backoff schedule paces it instead.

### DIFF
--- a/crates/cloud-sdk/src/error.rs
+++ b/crates/cloud-sdk/src/error.rs
@@ -77,18 +77,28 @@ pub enum SdkError {
     EventSourceError(String),
 }
 
-/// Error codes the Sandbox Proxy returns when it could not complete a request
-/// *before* forwarding it to the sandbox: the Indexify lookup did not answer
-/// (`SANDBOX_UPSTREAM_ERROR`), the sandbox has no route yet
-/// (`SANDBOX_NOT_READY`), or the dataplane could not be reached
-/// (`SANDBOX_UNREACHABLE`). In every case the request never reached the
-/// sandbox, so replaying it cannot duplicate a side effect — and every case is
-/// what an Indexify server rollout looks like from the client. The proxy's
-/// error body is `{"error": "...", "code": "..."}`.
+/// Error codes the Sandbox Proxy returns when it gave up on a request *before*
+/// any dataplane executed it, so replaying the request cannot duplicate a side
+/// effect:
+///
+/// - `SANDBOX_UPSTREAM_ERROR` — the Indexify lookup did not answer.
+/// - `SANDBOX_NOT_READY` — the sandbox has no route yet.
+/// - `SANDBOX_UNREACHABLE` — the dataplane could not be connected to, or a
+///   dataplane answered that it does not host the sandbox (its own
+///   `SANDBOX_NOT_FOUND` / `SANDBOX_NOT_RUNNING`, sent instead of executing
+///   the request) and the route could not be refreshed in time.
+/// - `AUTH_SERVICE_UNAVAILABLE` — the Platform auth check failed; nothing was
+///   forwarded.
+///
+/// These are what an Indexify server rollout looks like from the client. The
+/// proxy's error body is `{"error": "...", "code": "..."}`. Proxies older than
+/// compute-engine-internal #2083 report the lookup gap as a plain
+/// `SANDBOX_NOT_FOUND` 404, which is not replayed.
 pub const SANDBOX_PROXY_UNDELIVERED_CODES: &[&str] = &[
     "SANDBOX_UPSTREAM_ERROR",
     "SANDBOX_NOT_READY",
     "SANDBOX_UNREACHABLE",
+    "AUTH_SERVICE_UNAVAILABLE",
 ];
 
 /// Why a request failed below the HTTP status layer.
@@ -164,9 +174,13 @@ impl SdkError {
     /// executed it, making a replay safe even for a non-idempotent operation.
     ///
     /// True for a connect failure (see [`SdkError::transport_failure`]) and for
-    /// a 502/503 from the Sandbox Proxy carrying one of
-    /// [`SANDBOX_PROXY_UNDELIVERED_CODES`]. A timeout, or a 5xx without such a
-    /// code, gives no such guarantee.
+    /// a 5xx from the Sandbox Proxy carrying one of
+    /// [`SANDBOX_PROXY_UNDELIVERED_CODES`]. Any 5xx qualifies, not only the
+    /// 502/503 current proxies emit: older proxies relay the Indexify lookup's
+    /// own 500/504 verbatim under `SANDBOX_UPSTREAM_ERROR`, and those requests
+    /// were not forwarded either. A 4xx with one of these codes is a definitive
+    /// answer the proxy passed through (a 403, a 409) and is not replayed. A
+    /// timeout, or a 5xx without such a code, gives no guarantee at all.
     pub fn never_reached_server(&self) -> bool {
         if matches!(self.transport_failure(), Some(TransportFailure::Connect)) {
             return true;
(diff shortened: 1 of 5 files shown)


diptanu:
Code review of bc6e1765 — findings verified against the proxy and binding code, fixed in c3412b23 where they held.

**Fixed**
- **`never_reached_server()` only trusted 502/503.** Pre-#2083 proxies relay the Indexify lookup's own 500/504 verbatim under `SANDBOX_UPSTREAM_ERROR`, and those requests weren't forwarded either. Now any 5xx with an allowlisted code counts; 4xx under the same code (definitive 403/409 passthroughs) still doesn't. Also added `AUTH_SERVICE_UNAVAILABLE` (503, emitted before forwarding).
- **Transient schedule silently shortened.** One `backoff()` with a 5 s cap replaced the bindings' 15 s cap, cutting a `max_retries=10` caller from ~52.8 s to ~23.4 s. Split into `transient_backoff` (original 15 s cap, pinned by a test) and `undelivered_backoff` (5 s cap, so replays keep probing a recovering control plane).
- **GIL held across up to 30 s of sleeps.** `run_bounded_blocking` is reached from `&self` pyclass methods that hold the GIL; with the new budget that freezes every other Python thread for the length of a rollout. The loop now runs under `py.detach`, with `Send` bounds propagated to the blocking wrappers.
- **Docs.** Allowlist comment now explains why each code is pre-forward and names the #2083 dependency; orphaned prose from the deleted local predicates removed in both bindings; dead `[replay_on_connect_failure]` link fixed; `run_process` comments mention proxy pre-forward failures, not just connect failures. PR body rewritten to stop overstating `create_sandbox` coverage (it's a lifecycle route — connect-failure replay only).

**Checked, does not hold**
- *"`SANDBOX_UNREACHABLE` / `SANDBOX_NOT_READY` can follow a forwarded body."* Traced the proxy's stale-route recovery: it is entered from `fail_to_connect` (nothing written) or from the dataplane answering with its own `SANDBOX_NOT_FOUND` / `SANDBOX_NOT_RUNNING` header — an answer it sends *instead of* executing the request. Every exit code from that path is emitted while no dataplane has run the request. The only residual is Pingora's own keep-alive reuse retry, which is the same hazard the existing connect-failure classification already accepts.

**Real, deliberately left alone (called out in the PR body)**
- Mutating routes already under `run_with_retry(5, …)` (start_process, create_snapshot, create_pool, write_file, kill/send_signal, delete_*) retry an uncoded post-forward 502 `PROXY_ERROR`. Pre-existing; moving them to the non-idempotent policy is a separate behaviour change.
- Every wrapped call can now block ~30 s in a control-plane gap; deadline-bound pollers overshoot. Intended trade, now stated in the PR body.
- Retry lives only in the binding wrappers, not in `SandboxesClient`/CLI — architectural, not for this PR.

